PR #2654, opened 2022-02-01
fix(TagsInput): add status validation and other fixes

plagoa:
- update `onChange` signature
- update type definition to extend from `HvInputProps` instead of `HvBaseInputProps`
- Remove max-width limit on tags
- Add outline on focus
- Add error message and status validation
- Disable default behavior when enter is pressed to add a tag to prevent submiting a form where the component might be included 
- Added new story to illustrate usage of the component in a form

### DIFF
--- a/packages/core/src/TagsInput/TagsInput.js
+++ b/packages/core/src/TagsInput/TagsInput.js
@@ -34,6 +34,7 @@ const HvTagsInput = (props) => {
 
     readOnly = false,
     disabled = false,
+    required = false,
 
     label: textAreaLabel,
     "aria-label": ariaLabel,
@@ -217,6 +218,7 @@ const HvTagsInput = (props) => {
       disabled={disabled}
       readOnly={readOnly}
       status={validationState}
+      required={required}
       className={clsx(classes.root, className, {
         [classes.disabled]: disabled,
       })}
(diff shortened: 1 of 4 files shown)


plagoa:
@eliofreitas please take a look. I added a new story to showcase using the `TagsInput` component inside a form.

packages/core/src/TagsInput/TagsInput.js
```diff
@@ -153,7 +162,7 @@ const HvTagsInput = (props) => {
               setValue(newTagsArr);
               setTagCursorPos(tagCursorPos > 0 ? tagCursorPos - 1 : 0);
               inputRef.current?.focus();
-              onChange?.(newTagsArr);
+              onChange?.(undefined, newTagsArr);
```
eliofreitas, line 165: In here we could sent the event coming from the keydown

eliofreitas (changes requested):

eliofreitas (approved):

francisco-guilherme (approved):

diogofscmariano (changes requested):
I would adjust the same. The input should have the required option and actually display an error. Otherwise just remove it. Regarding the tagsInput, we should support the required option, having the * in the label and at least implement that in the sample, since this is a form field.

plagoa:
@diogofscmariano added the required property and updated the story to include that.

diogofscmariano (approved):

PR #2656, opened 2022-02-02
feat(snackbar): adding stacked snackbar provider.

eliofreitas:


packages/core/src/Snackbar/StackedSnackbarProvider/StackedSnackbarProvider.d.ts
```diff
@@ -0,0 +1,31 @@
...
+  snackbarContentProps?: HvSnackbarContentProps;
+}
+
+interface HvProviderContext extends ProviderContext {
+  enqueueSnackbar: (message: SnackbarMessage, options?: HvEnqueueOptions) => SnackbarKey;
+}
+
+export function useHvStackedSnackbar(): HvProviderContext;
+
+export default function HvProvider(props: HvStackedSnackbarProviderProps): JSX.Element | null;
```
nantunes, line 31: Should be named something other than "HvProvider".

francisco-guilherme (approved):

packages/core/src/Snackbar/StackedSnackbarProvider/StackedSnackbarProvider.js
```diff
@@ -0,0 +1,143 @@
...
+    enqueueNotistackSnackbar(
+      JSON.stringify({ message, variant, id, snackbarContentProps }),
+      otherOptions
+    );
+  };
+  return {
+    ...others,
+    enqueueSnackbar,
+  };
+};
```
nantunes, line 22: This will return different references on each hook call, that might lead to unwanted re-renders in its users. I suggest:
```suggestion
const useHvStackedSnackbar = () => {
  const { enqueueSnackbar: enqueueNotistackSnackbar, closeSnackbar } = useSnackbar();

  const enqueueSnackbar = useCallback((message, options) => {
    const { id, variant = "success", snackbarContentProps, ...otherOptions } = options;
    enqueueNotistackSnackbar(
      JSON.stringify({ message, variant, id, snackbarContentProps }),
      otherOptions
    );
  }, [enqueueNotistackSnackbar]);

  return useMemo(() => ({
    enqueueSnackbar,
    closeSnackbar
  }), [enqueueSnackbar, closeSnackbar]);
};
```

packages/core/src/Snackbar/StackedSnackbarProvider/StackedSnackbarProvider.js
```diff
@@ -0,0 +1,143 @@
...
+
+import HvSnackbarContentWrapper from "../SnackbarContentWrapper";
+
+// We override notistack hook to be able to customize the snackbar that should be called.
+const useHvStackedSnackbar = () => {
+  const { enqueueSnackbar: enqueueNotistackSnackbar, ...others } = useSnackbar();
+  const enqueueSnackbar = (message, options) => {
+    const { id, variant = "success", snackbarContentProps, ...otherOptions } = options;
+    enqueueNotistackSnackbar(
+      JSON.stringify({ message, variant, id, snackbarContentProps }),
```
nantunes, line 14: Why not create here the `HvNotistackSnackMessage` instead of stringifying? That will block the ability to pass a node as label.

Notistack's SnackbarMessage can be string or React.ReactNode.

nantunes (changes requested):

nantunes (approved):

diogofscmariano (approved):

PR #2664, opened 2022-02-08
fix(TagsInput): fix reset input on value change

plagoa:
- removed the `onBlur` from the `HvTagsInputProps`
- moved the reset input funcionality to a specific `useEffect` to ensure that the input is only reset if the right conditions are met
- added a new story to illustrate the behavior


eliofreitas (approved):

packages/core/src/TagsInput/TagsInput.js
```diff
@@ -83,6 +83,7 @@ const HvTagsInput = (props) => {
   const [validationMessage, setValidationMessage] = useControlled(statusMessage, "");
 
   const [stateValid, setStateValid] = useState(true);
+  const [resetInput, setResetInput] = useState(false);
```
nantunes, line 86: I suggest you turn `resetInput` into a ref. We want to perform the "reset" when the value changes, the reset flag is auxiliary. Being a stateful var, the useEffect will also unnecessarily trigger when it returns to false.

packages/core/src/TagsInput/TagsInput.js
```diff
@@ -159,6 +151,13 @@ const HvTagsInput = (props) => {
     }
   }, [multiline, tagCursorPos]);
 
+  useEffect(() => {
+    if (resetInput) {
+      setTagInput("");
+      setTagCursorPos(value.length);
+    }
```
nantunes, line 158: Shouldn't `resetInput` return to `false` here?

packages/core/src/TagsInput/TagsInput.js
```diff
@@ -200,6 +198,7 @@ const HvTagsInput = (props) => {
                 ...value.slice(tagCursorPos + 1),
               ];
               setValue(newTagsArr);
+              setResetInput(false);
```
nantunes, line 201: I understand: the reset shouldn't occur when deleting. But why would it be true? Maybe because it wasn't reset back to false in the useEffect?

packages/core/src/TagsInput/TagsInput.js
```diff
@@ -239,6 +239,7 @@ const HvTagsInput = (props) => {
       const newTagsArr = [...value.slice(0, i), ...value.slice(i + 1)];
       setValidationState(validationStates.standBy);
       setValue(newTagsArr);
+      setResetInput(false);
       setTagCursorPos(newTagsArr.length);
       inputRef.current?.focus();
       performValidation(newTagsArr);
```
nantunes, line 245: Not related with this change: I notice the delete code is almost triplicated... Couldn't it be abstracted in some way and be called at the three delete methods?

nantunes (changes requested):

nantunes:
Just another note: the input grows when the error message is visible.
image

That also happens in a regular HvInput, I think the sample needs to have a defined width.

plagoa:
@eliofreitas @nantunes updated the PR addressing your questions.

nantunes (approved):

Issue #2667, opened 2022-02-09
Tooltip is displaying everywhere in Bar Chart after mouse hover


rajashriDalavi:
### Latest version

- [X] I have tested the latest version

### No Hitachi Vantara confidential content

- [X] I'm not including any Hitachi Vantara confidential content

### Current behavior 😯

- Tooltip is displaying everywhere in Bar Chart after mouse hover
-  Tooltip is displaying everywhere on the chart after mouse hover.
![image](https://user-images.githubusercontent.com/13730428/153243806-71665140-21d0-42f5-91fa-d5f12a3a3042.png)

- It should display after hover on bar of the graphs.

### Expected behavior 🤔

- It should display after hover on bar of the graphs.

### Steps to reproduce 🕹

Steps:

1. Tooltip is displaying everywhere in Bar Chart after mouse hover
2. Tooltip is displaying everywhere on the chart after mouse hover.
![image](https://user-images.githubusercontent.com/13730428/153243806-71665140-21d0-42f5-91fa-d5f12a3a3042.png)

### Your Environment 🌎

Device: [e.g. Laptop, Mobile, ....]
Browser [e.g. stock browser, safari, chrome]
Other details [e.g. screen resolution 1920x1080, ...]

diogofscmariano:
Hello @rajashriDalavi. Thanks for submitting this issue. This looks to be an issue with our integration between Plotly and the tooltip we add. We will take a look at it today, although no ETA because we will just start investigating it. 
Please keep in mind uikit wraps PlotlyJS, adding a couple of styling definitions. It is on the team's responsibility to configure the components to fulfill their needs and adjust PlotlyJS configurations.

diogofscmariano:
Hello @rajashriDalavi. A fix for this issue is now available in the latest release. Can you integrate and let us know how it goes?

rajashriDalavi:
Hi @diogofscmariano  Thanks for update. We will plan this in next sprint. Will update status.

plagoa:
@rajashriDalavi closing this issue as a fix has been release. Let us know if this is still relevant.

PR #2670, opened 2022-02-10
fix(BarChart): fix tooltip displaying when hovering outside of the chart bars

plagoa:
This is the proposed fix for:
https://hv-eng.atlassian.net/browse/HVUIKIT-6312

The problem was that the tooltip would display even outside the chart bars. I added a default value for the `hovermode` property on plotly to set it to `closest`. This value can still be overridden by the user to keep grouping functionality like I did in the two updated stories.


diogofscmariano:
fixes: #2667

diogofscmariano (approved):

PR #2678, opened 2022-02-15
fix(TextArea): fixed TextArea scrolling when it shouldn't

plagoa:


nantunes (approved):

eliofreitas (changes requested):
the new example is not static, this may cause problems with aplitools, so could you add `eyes: { include: false }` to the story parameters?, this will exclude the example from aplitools

diogofscmariano (approved):

PR #2679, opened 2022-02-17
feat(Loading): adding a progress bar mode

plagoa:
Notes:
- This only features the progress bar in a determinate mode. I'm a bit unsure as to what the behavior in undeterminate mode should be (I'd assume an infinite looping bar but the specs don't really mention anything about it);
- This is borderline what I'm confortable with as to making a component have some conditional rendering inside, maybe the progress part should be moved to a specific `ProgressBar` component inside the `Loading` component;


francisco-guilherme (approved):
Great Job 👍

hyorran (approved):

packages/core/src/Loading/Loading.d.ts
```diff
@@ -29,4 +29,4 @@ export interface HvLoadingProps
   color?: string;
 }
 
-export default function HvLoading(props: HvLoadingProps): JSX.Element | null;
+export default function HvLoading(props: HvLoadingProps): JSX.Element | null;
```
eliofreitas, line 32: please include an enter at the end of the file to avoid this warning

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,134 @@
...
+ * ProgressBar provides feedback about a process that is taking place in the application.
+ */
+const HvProgressBar = (props) => {
+  const {
+    className,
+    classes,
+    value = 0,
+    error,
+    undeterminate = false,
+    "aria-label": ariaLabel,
```
eliofreitas, line 22: Explicitly defining the aria-label prop in here is not necessary because it is an standard html prop, so it is implicitly present in the `...others`

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,134 @@
...
+              classes.progressBar,
+              clampedValue === 100 && classes.progressDone,
+              error && classes.progressError
+            )}
+          />
+        </div>
+        <span
+          aria-live="polite"
+          style={{ display: "inline-block", width: "1px", height: "1px", overflow: "hidden" }}
+          {...(error && { "aria-invalid": true })}
```
eliofreitas, line 63: I think the aria-invalid should be present where the aria-valuenow property  is placed, because that element is the one that the reader will see

packages/lab/src/ProgressBar/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+        value?: number;
+        /**
+         * Indicates if there was an error while loading.
+         */
+        error?: boolean;
+        /**
+         * 
+         */
+        undeterminate?:boolean
+
```
eliofreitas, line 29: this file format is wrong we shouldn't include so much spaces

eliofreitas (changes requested):
Great work, I left some small comments nothing to worry

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,125 @@
...
+  const clampedValue = clamp(value, 0, 100);
+
+  return (
+    <div
+      className={clsx(className, classes.root, classes.progress)}
+      role="progressbar"
+      aria-valuenow={clampedValue}
+      aria-valuemin="0"
+      aria-valuemax="100"
+      aria-busy={clampedValue !== 100}
```
eliofreitas, line 27: eliminate this is to be done by the user implementing the bar

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,125 @@
...
+      {...(error && { "aria-invalid": true })}
```
eliofreitas, line 28: aria-invalid is only for form elements

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,125 @@
...
+              error && classes.progressError
+            )}
+          />
+        </div>
+        <span
+          aria-live="polite"
+          style={{ display: "inline-block", width: "1px", height: "1px", overflow: "hidden" }}
+        >
+          {clampedValue}
+        </span>
```
eliofreitas, line 57: This logic could be fused with the label, and when it is in an undetermined state it hides the label but remains in the DOM

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,125 @@
...
+
+import { withStyles } from "@material-ui/core";
+
+import styles from "./styles";
+
+/**
+ * ProgressBar provides feedback about a process that is taking place in the application.
+ */
+const HvProgressBar = (props) => {
+  const { className, classes, value = 0, error, undeterminate = false, ...others } = props;
```
eliofreitas, line 16: error -> status:` invalid, valid, in progress ; "success", "in progress", "error"`. 

add a labelProp to spread generic values on the label of the element
should be typed as generic object in js
should be typed as div object in ts

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,125 @@
...
+   */
+  error: PropTypes.bool,
+  /**
+   * Indicates if it shows or not the value label
+   */
+  undeterminate: PropTypes.bool,
+  /**
+   * @ignore
+   */
+  "aria-label": PropTypes.string,
```
eliofreitas, line 122: this is no longer needed

eliofreitas (changes requested):

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,135 @@
...
+      className={clsx(className, classes.root, classes.progress)}
+      role="progressbar"
+      status="inProgress"
+      aria-valuenow={clampedValue} // it only makes sense to put this in ariaProps if the min and max values of the progress bar are intended to be changed by the user
+      aria-valuemin="0"
+      aria-valuemax="100"
+      {...others}
+    >
+      <div className={classes.progressContainer}>
+        <div
```
eliofreitas, line 39: ariaProps -> labelProps
see if we can eliminate the div and use only the `HvTypography`

eliofreitas (changes requested):

PR #2680, opened 2022-02-21
build: add vulnerabilities check in CodeDx

diogofscmariano:


diogofscmariano:
build run in https://github.com/lumada-design/hv-uikit-react/runs/5275805935?check_suite_focus=true

nantunes (approved):

PR #2687, opened 2022-03-03
feat(TagsInput): allow commiting tags with other characters and on blur

plagoa:
- new property `commitOnTag` allows passing an array of key codes that can be used to commit a tag
- new property `commitOnBlur` allows commiting a tag when clicking outside of the input box

### DIFF
--- a/packages/core/src/TagsInput/stories/TagsInput.stories.js
+++ b/packages/core/src/TagsInput/stories/TagsInput.stories.js
@@ -460,3 +460,34 @@ InAForm.parameters = {
     description: { story: "Tags Input in a form." },
   },
 };
+
+export const CustomCommitCharacter = () => {
+  const useStyles = makeStyles(() => ({
+    root: {
+      width: 550,
+    },
+  }));
+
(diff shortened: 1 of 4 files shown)


packages/core/src/TagsInput/stories/TagsInput.stories.js
```diff
@@ -460,3 +460,34 @@ InAForm.parameters = {
...
+    <HvTagsInput
+      id="tags-list-11"
+      label="Custom commit character"
+      description="Will only add a tag when a space or comma is entered or when the user clicks outside the input box and there's text that's not been commited"
+      aria-label="Custom commit character"
+      placeholder="Enter value"
+      classes={{
+        root: classes.root,
+      }}
+      commitTagOn={["Space", "Comma"]}
```
eliofreitas, line 483: I think this is enough for now however, I don't know what would be easier for the user, this or directly passing the code, or somehow passing the character like ","

eliofreitas (approved):

francisco-guilherme (approved):

PR #2692, opened 2022-03-11
build: changed scripts to also run in windows HVUIKIT-6320

HQFOX:


francisco-guilherme (changes requested):
It should be installed as dev dependency.

eliofreitas (approved):

francisco-guilherme (approved):

PR #2711, opened 2022-03-29
chore: allow plotly.js@2

### LINKED ISSUE #2710: Plotly breaks UI-Kit in SSG and ES6 modules imports
### Latest version

- [X] I have tested the latest version

### No Hitachi Vantara confidential content

- [X] I'm not including any Hitachi Vantara confidential content

### Current behavior 😯

[`plotly.js@1` has a bug](https://github.com/plotly/plotly.js/issues/3518) that prevents:
- ES6 module loading (breaking modern dev/bundling tools such as `vite` and `snowpack`)
- Server-Side-Rendering because of the way it depends on `document` (Edit: plotly@2 doesn't really solve this)

As `plotly.js@1` is exported (indirectly) in the top-level `index.js`, it will break as soon as I use any component.

### Expected behavior 🤔

Being able to use UI-Kit in modern bundling tools such as [`vite`](https://vitejs.dev/) ~and SSG frameworks such as [`remix`](https://remix.run/)~

### Steps to reproduce 🕹

Steps:

1. Create a Vite app (`npm init vite`, choose a react template)
2. Install UI-Kit `npm i @hitachivantara/uikit-react-core @material-ui/core plotly.js-basic-dist@1`
3. Import `HvProvier` in `App.tsx`
4. Start dev server: `npm run dev`
5. Notice how the app breaks right away

### Context 🔦

This error is fixed in `plotly.js@2`

### Your Environment 🌎

Latest MacOS, Chrome, UI-Kit

zettca:
Allows an app to use `plotly.js` v1 or v2, which fixes #2710.

~Testing if internal bump to `plotly.js@2` breaks~. It seems like the [breaking changes](https://github.com/plotly/plotly.js/releases/tag/v2.0.0) ~might~do not affect UI-Kit's charts.

Does it make sense to make this bump in a patch/minor?

Looking forward, given:
- that plotly still relies heavily on `window.document`
- that plotly is a large library
- how often the charts are used
- that in Design System they're in a separate "visualisation" component library
it might be worth it to look into splitting the Visualisation components into a separate library

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28771,9 +28771,9 @@
       }
     },
     "plotly.js-basic-dist": {
-      "version": "1.58.5",
-      "resolved": "https://registry.npmjs.org/plotly.js-basic-dist/-/plotly.js-basic-dist-1.58.5.tgz",
-      "integrity": "sha512-Srbt1ACLknY2r655udWwc8cxEan0FvhQUDkX/VW9NATNs5g38mc+OTLmZIRmUrJu2Sr/jDXqZ4EqFnzN9dOXvQ==",
+      "version": "2.11.1",
+      "resolved": "https://registry.npmjs.org/plotly.js-basic-dist/-/plotly.js-basic-dist-2.11.1.tgz",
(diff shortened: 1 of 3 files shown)


francisco-guilherme:
Hi @zettca , it makes a lot of sense to do this update 👍 
We just need to reinsure that it doesn't affect the UI Kit.
We are also planning on moving the visualisations to their own package on the next major.

eliofreitas (approved):

PR #2713, opened 2022-03-29
style: added new blur effect on backdrop

hyorran:
style(global-actions): added new blur effect on backdrop


plagoa (approved):

francisco-guilherme (changes requested):
@hyorran please just add the scope of the commit message: style:(global-actions) ...

PR #2729, opened 2022-04-07
feat(AppSwitcherPanel): promote component to core

plagoa:


packages/core/src/AppSwitcherPanel/Action/Action.js
```diff
@@ -0,0 +1,204 @@
...
+  const [validIconUrl, setValidIconUrl] = useState(true);
+
+  const renderApplicationIcon = () => {
+    if (iconElement) {
+      return iconElement;
+    }
+
+    if (iconUrl && validIconUrl) {
+      return (
+        // eslint-disable-next-line jsx-a11y/alt-text
```
eliofreitas, line 42: Maybe we can avoid this disable by having the `description=""` as default or in alternative maybe using the role="presentation" as the image may not be critical for accessibility it is just decoration

packages/core/src/AppSwitcherPanel/Action/Action.js
```diff
@@ -0,0 +1,204 @@
...
+  const {
+    id,
+    className,
+    classes,
+    application,
+    onClickCallback = () => {},
+    isSelectedCallback = () => false,
+  } = props;
+
+  const { name, description, disabled, iconElement, iconUrl, url, target } = application;
```
eliofreitas, line 26: maybe we could enhance the api with a `genericProp` to spread extra values to the components below

eliofreitas (changes requested):

francisco-guilherme (approved):

francisco-guilherme (approved):

PR #2731, opened 2022-04-11
Hv UI kit 6326 pagination number of visible items not aligned vertically

hyorran:


packages/core/src/VerticalNavigation/styles.js
```diff
@@ -8,6 +8,7 @@ const styles = (theme) => ({
 
     background: theme.hv.palette.atmosphere.atmo1,
     boxShadow: theme.hv.shadows[1],
+    clipPath: "inset(0px -12px 0px 0px)",
```
francisco-guilherme, line 11: This change is out of the scope of this bug fix.

francisco-guilherme (changes requested):

packages/core/src/VerticalNavigation/styles.js
hyorran, line 11:
```suggestion
```

eliofreitas (changes requested):
commit message is corrupted

PR #2732, opened 2022-04-11
fix(TagsInput): fix incorrect input width when no tags exist

plagoa:


eliofreitas (approved):

packages/core/src/TagsInput/TagsInput.js
```diff
@@ -483,7 +483,7 @@ const HvTagsInput = (props) => {
           })}
         {!(disabled || readOnly) && (
           <HvListItem
-            className={clsx(!multiline && classes.singleLine)}
+            className={clsx(!multiline && classes.singleLine, value.length === 0 ? "empty" : "")}
```
eliofreitas, line 486: Maybe in here we could use a conditional class instead of a conditional string something like `classes.empty`, and in the styles refer to it as `$empty` which would refer to the property/class that was defined

HQFOX (approved):

francisco-guilherme (approved):

PR #2737, opened 2022-04-14
build: fixed commit  hooks not working

HQFOX:
The commit hooks weren't working on my machine. This is the way I found for it to work but, i need confirmation that it works as expected for everyone else.


francisco-guilherme:
Closed in favor of #2746

PR #2738, opened 2022-04-14
fix(dotPagination): added missing typescript definition and description.

HQFOX:
…ription HVUIKIT-6297

### DIFF
--- a/packages/lab/src/DotPagination/DotPagination.d.ts
+++ b/packages/lab/src/DotPagination/DotPagination.d.ts
@@ -1,11 +1,26 @@
-import * as React from "react";
 import { StandardProps } from "@material-ui/core";
+import { HvRadioGroupProps } from "@hitachivantara/uikit-react-core";
 
-export type HvDotPaginationClassKey = "root";
+export type HvDotPaginationClassKey = "root" | "horizontal" | "radioRoot";
 
-export type HvDotPaginationProps = StandardProps<
-  React.HTMLAttributes<HTMLDivElement>,
-  HvDotPaginationClassKey
->;
+export interface HvDotPaginationProps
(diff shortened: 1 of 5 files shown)


eliofreitas (changes requested):
a couple of things to change:

- The commit should be fix, we want this in the change-log
- Remember to update the typescript definition, because the component is now extending from the hvRadiogroup

francisco-guilherme (approved):

packages/lab/src/DotPagination/DotPagination.js
```diff
@@ -38,22 +38,25 @@ const HvDotPagination = (props) => {
...
+      orientation="horizontal"
+      {...others}
+    >
+      {range(pages).map((i) => (
+        <HvRadio
+          classes={{ root: classes.radioRoot }}
+          key={i}
+          value={i}
+          checked={page === i}
+          onChange={() => onPageChange(i)}
```
eliofreitas, line 53: This onPageChange could also bubble up the Change event to the user 

`onChange={() => onPageChange(i)}` -> `onChange={(event) => onPageChange(event, i)}`

packages/lab/src/DotPagination/DotPagination.d.ts
```diff
@@ -1,11 +1,26 @@
...
+   */
+  pages?: number;
+  /**
+   * The currently selected page.
+   */
+  page?: number;
+  /**
+   * The callback fired when the page value changes.
+   */
+  onPageChange?: (value: number) => void;
```
eliofreitas, line 19: if the event is included this signature should also be updated

eliofreitas (changes requested):

eliofreitas (approved):

Issue #2740, opened 2022-04-18
[NotificationPanel] - When a notification has no action associated onClick event we should be able to remove the css cursor


andrefrua:
### Latest version

- [X] I have tested the latest version

### No Hitachi Vantara confidential content

- [X] I'm not including any Hitachi Vantara confidential content

### Current behavior 😯

We got a requirement from PD that when a notification has no origin URL the notification shouldn't be clickable.
This is easily achievable by removing our actions from the onClick event, however, currently there is no way to remove the css  cursor indicating that the notification is clickable.

### Expected behavior 🤔

We need a way to inform the panel that a notification is not clickable and showing it in the UI by not adding the `pointer` cursor from the notification element.

### Steps to reproduce 🕹

Just open the notification panel and verify that all the notification are displayed with the `pointer` cursor.

### Context 🔦

We got a requirement from PD that when a notification has no origin URL the notification shouldn't be clickable.

### Your Environment 🌎

All environments

PR #2741, opened 2022-04-18
fix(NotificationPanel): toggles pointer cursor according to the new isclickable property (#2740)

andrefrua:


francisco-guilherme:
Closed in favor of #2742

PR #2748, opened 2022-04-20
fix(ProgressBar): corrected typescript definition

HQFOX:
I also created a file "ProgressBarSimulator.js" so that the component used in testing and in the storybook to simulate the progress bar moving isn't repeated in two files.

Removed "magic strings" used in implementation for the status, now the corresponding values are stored in an object: statusCategories.


hyorran:
Just fix the description please.

hyorran (approved):

packages/lab/src/ProgressBar/ProgressBar.js
```diff
@@ -9,6 +9,8 @@ import { withStyles } from "@material-ui/core";
 
 import styles from "./styles";
 
+const statusCategories = { inProgress: "inProgress", completed: "completed", error: "error" };
```
eliofreitas, line 12: You can include an `object.freeze({ inProgress: "inProgress", completed: "completed", error: "error" };)` to ensure this object remains consistent

eliofreitas (changes requested):

eliofreitas (changes requested):

eliofreitas (approved):

PR #2754, opened 2022-04-22
feat(TimePicker): promote component to core

plagoa:


francisco-guilherme (changes requested):
You forgot to add the component both to index.js and index.d.ts.

packages/core/src/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,725 @@
...
+  const minutesInputState =
+    time?.minutes != null && time.minutes !== "" && time.minutes >= 0 && time.minutes <= 59;
+  const secondsInputState =
+    time?.seconds != null && time.seconds !== "" && time.seconds >= 0 && time.seconds <= 59;
+
+  return hourInputState && minutesInputState && secondsInputState;
+};
+
+/**
+ * A TimePicker component used to choose the time, following specifications provided by Design System. Still in development.
```
francisco-guilherme, line 62: Please remove the "Still in development".
Maybe the "following specifications provided by Design System" is redundant.
It seems good enough just "A TimePicker component used to choose the time."

plagoa: Well spotted, I'll remove it.

francisco-guilherme (approved):

packages/core/src/TimePicker/enums.js
```diff
@@ -0,0 +1,34 @@
+const TimePickerUnits = {
```
eliofreitas, line 1: all the objects in this enum file could use an Object.freeze() to prevent any undesired modification

plagoa:
@eliofreitas addressed your issues.

eliofreitas (approved):

PR #2763, opened 2022-05-04
feat(Stack): create Stack component

plagoa:


francisco-guilherme (approved):

packages/core/src/Stack/Stack.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * The divider component to be used between the stack elements.
+   * - If `true` the Material-UI Divider component will be used.
+   * - If a React node is passed then the custom divider will be used.
+   */
+  divider?: boolean | React.ReactNode;
+  /**
+   * The properties to pass on to the Material-UI component.
+   */
+  dividerProps: object;
```
eliofreitas, line 31: check comment above

packages/core/src/Stack/Stack.js
```diff
@@ -0,0 +1,155 @@
...
+  const widths = useMemo(() => [...theme.breakpoints.keys], [theme]);
+
+  /**
+   * @returns {string} - Returns a direction for the stack: column or row. If the
+   *                     `direction` property is a string and a valid direction then we
+   *                     use it. If it's an object with multiple directions by breakpoint
+   *                     we use the appropriate one or search for the nearest breakpoint
+   *                     smaller than the current one to use.
+   */
+  const getDirection = useCallback(() => {
```
eliofreitas, line 40: This function doesn't receive any parameter, yet in line 52 some parameters are passed. However receiving the parameters may be good, if it receives parameters and avoids accessing the values from context the function could be pure and avoid the `useCallback`

eliofreitas (changes requested):

plagoa:
@eliofreitas updated accordingly.

eliofreitas (approved):

PR #2767, opened 2022-05-06
fix: change default exports version

zettca:
Currently, UI-Kit's packages default `exports` are to `/dist/modern`. This `modern` version uses ESM `import`/`export` syntax, which is still not widely supported. Furthermore, the [`exports`](https://nodejs.org/docs/latest-v16.x/api/packages.html#exports) entrypoint points to a different version from the `main` (legacy) entrypoint.

~I think that the `main` and `exports` entrypoints should point to the same version. Otherwise libraries might read from one or the other. And they should point to the top-level `/dist/`/`"node"` export, for compatibility reasons, as it is defined in `"main"`~
(see conversation below)

I recently tried upgrading to [`jest@28` which is using `exports`](https://jestjs.io/docs/upgrading-to-jest28#packagejson-exports), and the tests broke because of the the UI-Kit's modern module. But I've had this issue previously in other libraries/frameworks, like when trying [Remix](https://remix.run/)

```
    SyntaxError: Unexpected token 'export'

      1 | import React from "react";
      2 | import { MemoryRouter } from "react-router-dom";
    > 3 | import { HvProvider } from "@hitachivantara/uikit-react-core";
        | ^
```


nantunes (dismissed):
That is the intended setup: `main` targets a commonjs modules release, `module` ships ES5, and `exports` is ES2017.
By default webpack 4 projects should be reading `module`, webpack 5 should use `exports`.

At the time this was one of the reference articles used: https://web.dev/i18n/en/publish-modern-javascript/, but others suggested the same approach for having a package shipping several module formats.

zettca:
> That is the intended setup: `main` targets a commonjs modules release, `module` ships ES5, and `exports` is ES2017. By default webpack 4 projects should be reading `module`, webpack 5 should use `exports`.
> 
> At the time this was one of the reference articles used: https://web.dev/i18n/en/publish-modern-javascript/, but others suggested the same approach for having a package shipping several module formats.

Webpack isn't the problem in this case, jest is. The latest jest@28 [still doesn't support ESM OOTB](https://jestjs.io/docs/ecmascript-modules). [Here's a related issue](https://github.com/microsoft/accessibility-insights-web/pull/5421#issuecomment-1109168149), where of all of the workarounds, the one chosen was a jest resolver [to remove "exports"](https://github.com/microsoft/accessibility-insights-web/pull/5421/commits/9ad4e618019298d82732d49d00aafb846fb6bac7#diff-10ce3d2c4d812be4f9bbaceb999b48c91538e4ccfc23dc585c59deb316d0d689R24) from the library.

I looked into how popular UI libraries are dealing with this, and I couldn't find a single one using the `"exports"` key:
- https://unpkg.com/browse/@material-ui/core@4.12.4/package.json
- https://unpkg.com/browse/@mui/material@5.6.4/package.json
- https://unpkg.com/browse/@mantine/core@4.2.2/package.json
- https://unpkg.com/browse/@carbon/react@1.2.0/package.json
- https://unpkg.com/browse/antd@2.13.14/package.json
- https://unpkg.com/browse/react-bootstrap@2.3.1/package.json
- https://unpkg.com/browse/semantic-ui-react@2.1.2/package.json

It looks like MUI have tried this in the past, but [reverted it](https://github.com/mui/material-ui/pull/26656) because it broke NextJS [and also reportedly jest](https://github.com/mui/material-ui/pull/26656#issuecomment-857549476).
Again, these `"exports"` are also breaking when trying to use UI-Kit with `Remix` (maybe NextJS too). 

Would you be against not having `"exports"` declared all-together? It also resolves the `jest` and Remix issues.

zettca:
Edit: Out of curiosity, I've also tried `NextJS`. Same issue thing happens.
```
> Build error occurred
/Users/brhenriques/Desktop/uikit-next/node_modules/@hitachivantara/uikit-react-core/dist/modern/index.js:2
export { default as HvAccordion } from "./Accordion";
^^^^^^

SyntaxError: Unexpected token 'export'
```

nantunes:
@zettca I'll have to take a better look at the issue, but don't have the time right now. Is this a blocking issue for you?

zettca:
> @zettca I'll have to take a better look at the issue, but don't have the time right now. Is this a blocking issue for you?

All right 👍 

It is not blocking ATM.

zettca:
For clarity: this PR isn't a work-in-progress, it's waiting for reviews.

francisco-guilherme (approved):

PR #2781, opened 2022-05-23
fix(QueryBuilder): allow configuring styles via classes object

plagoa:


eliofreitas (changes requested):
The common usage of the classes object is to customize the root components this might need another approach

eliofreitas (approved):

PR #2794, opened 2022-06-02
fix(query-builder): adding styles to the query builder to avoid mater…

eliofreitas:
…ial warning


zettca (approved):
Totally missed this file 👍

francisco-guilherme (approved):

PR #2805, opened 2022-06-14
fix(filter-group): counter component was not counting correctly

plagoa:


eliofreitas (changes requested):
waiting for requested test

PR #2823, opened 2022-06-30
fix(dropdown): allow number `id`

zettca:


eliofreitas (changes requested):
The change seems to be expanding the API to also support numbers which is fine, however having two commits makes github accumulates the changes as just erasing values from the API, i think if you squash the commit it should solve the problems

zettca:
> The change seems to be expanding the API to also support numbers which is fine, however having two commits makes github accumulates the changes as just erasing values from the API, i think if you squash the commit it should solve the problems

Apologies, it was late and I didn't write the PR's description. Updated!

eliofreitas (changes requested):
waiting for adjusment to API to avoid deleting the id

eliofreitas (approved):

PR #2843, opened 2022-07-08
fix(navigation): adding item as links

eliofreitas:


HQFOX (approved):

francisco-guilherme (changes requested):
Please add the `href` and `target` to:
https://github.com/lumada-design/hv-uikit-react/blob/d52e7d55f1766e76ef93600d91117022368b576e/packages/core/src/Header/Navigation/Navigation.js#L47

francisco-guilherme (approved):

PR #2868, opened 2022-07-26
fix(NotificationPanel): remove locale logic app

zettca:
This locale loading logic should live in the Apps' root configurations, not inside a UI Kit component.
This logic will be present in the apps anyways, and the dynamic import can mess up bundling and tree-shaking

@andrefrua @lumada-design/imperial please review.

PS: something's wrong with tagging @lumada-design/imperial as reviewer 🤔


andrefrua (approved):
I don't see any negative impact on our side with this change 👍

HQFOX (approved):

PR #2870, opened 2022-07-26
fix(FileUploader): add missing id in type definition

HQFOX:
Added required id, and added id in type definition in order to prevent the `Warning: Each child in a list should have a unique "key" prop` 

The fact that is now required changes the api but it only shows a warning.

### DIFF
--- a/packages/core/src/FileUploader/FileUploader.d.ts
+++ b/packages/core/src/FileUploader/FileUploader.d.ts
@@ -19,6 +19,10 @@ export type HvFileUploaderClassKey = "root";
 export interface FileUploaderProps
   extends StandardProps<React.HTMLAttributes<HTMLDivElement>, HvFileUploaderClassKey> {
   /**
+   * Id to be applied to the root node.
+   */
+  id: string;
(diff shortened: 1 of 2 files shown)


packages/core/src/FileUploader/FileUploader.d.ts
```diff
@@ -18,6 +18,10 @@ export type HvFileUploaderClassKey = "root";
 
 export interface FileUploaderProps
   extends StandardProps<React.HTMLAttributes<HTMLDivElement>, HvFileUploaderClassKey> {
+  /**
+   * Id to be applied to the root node.
+   */
+  id: string;
```
zettca, line 24: This won't lead to a warning, it will lead to a compilation error

zettca: I opened an issue regarding the `id`s and `key`s: https://github.com/lumada-design/hv-uikit-react/issues/2873

francisco-guilherme:
Closed and tracked in #2873

PR #2883, opened 2022-08-09
fix(file-uploader): fix undefined id's on file list

HQFOX:


francisco-guilherme (approved):

plagoa (approved):

packages/core/src/FileUploader/FileList/FileList.js
```diff
@@ -1,20 +1,22 @@
...
 
 const FileList = ({ id, classes, list = [], removeFileButtonLabel, onFileRemoved }) => {
+  const elementId = useUniqueId(id, "hvfilelist");
+
   const hasFiles = list.length > 0;
   if (!hasFiles) return null;
 
   return (
-    <ul id={id} className={classes.list}>
+    <ul id={setId(id, "list")} className={classes.list}>
```
zettca, line 15: Here I would be passing an `id` to `FileList`, but that passed `id` will never be visible

```tsx
<FileList id="idsSuckBtw" />
```

And then if I do something like `document.getElementById("idsSuckBtw")` I would get nothing, because it would get some internal suffix I don't know about.

Also, we're changing the current existing `id`.
This could be considered a breaking change, as some users might be using the `id` for tests. But regarding this I'd say its their own fault [for not using better selectors](https://testing-library.com/docs/queries/about#priority) 😛

packages/core/src/FileUploader/FileList/FileList.js
```diff
@@ -1,20 +1,22 @@
...
       {list.map((data) => (
         <File
           key={data.id}
-          id={setUid(id, data.id)}
+          id={setId(elementId, "values")}
```
zettca, line 19: Wasn't this meant to be passed to the `key` instead? Would need the `index` though.

If this was intended then:
- All the `File`s will be getting the same `id`, which is wrong
- We're adding extra stuff to the DOM that is unnecessary and unasked for
  - I don't like polluting the DOM with useless `id`s. Also, someone might use them for testing, which is bad

^ The snapshot tests actually caught these two _issues_, but nobody reads snapshots 🤷

PR #2887, opened 2022-08-10
feat(time-picker): add read only mode

plagoa:


francisco-guilherme (approved):

packages/core/src/TimePicker/TimePicker.d.ts
```diff
@@ -44,6 +44,19 @@ export interface HvTimePickerProps
...
+  required: boolean;
+  /**
+   * Indicates that the form element is disabled.
+   */
+  disabled: boolean;
+  /**
+   * Indicates that the form element is in read only mode.
+   */
+  readOnly: boolean;
+
```
zettca, line 59: these should be optional

plagoa: They should, indeed. Thanks!

PR #2892, opened 2022-08-12
feat(query-builder): add read only mode

plagoa:


packages/lab/src/QueryBuilder/Rule/Operator/Operator.js
```diff
@@ -33,6 +33,7 @@ const Operator = ({ id, combinator, attribute, operator }) => {
       placeholder={labels.rule.operator.placeholder}
       values={values}
       disabled={values.length === 0}
+      readOnly={readOnly}
```
zettca, line 36: Not really related to the QueryBuilder, but I think it would be nice to be able to select the values from `readOnly` dropdowns. Currently we can't because of the `userSelect: "none"`.

We could enable `userSelect` in `headerReadOnly`

image

Just an idea 😃

zettca (approved):
Looking great! Thanks for working on this so fast 👍 

Just some minor comments:

packages/lab/src/QueryBuilder/stories/QueryBuilder.stories.js
```diff
@@ -153,3 +153,110 @@ InitialQuery.parameters = {
...
+      />
+      <pre>{JSON.stringify(mongoQuery, null, 2)}</pre>
+    </>
+  );
+};
+
+ReadOnly.parameters = {
+  docs: {
+    description: {
+      story: "With query parsed to Mongo",
```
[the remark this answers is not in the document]
plagoa: I just wanted to have a sample to have a bit of everything but maybe I can thin it down a bit.

packages/lab/src/QueryBuilder/Rule/Operator/Operator.js
plagoa: Ah, that "kind of" came up when doing this work, but I'll check with the DS team to see what they think about it.

packages/lab/src/QueryBuilder/QueryBuilder.d.ts
```diff
@@ -48,6 +48,10 @@ export interface HvQueryBuilderProps
    * An object containing all the labels.
    */
   labels?: QueryBuilderLabels;
+  /**
+   * A flag indicating if the Query Builder is in read only mode.
+   */
+  readOnly: boolean;
```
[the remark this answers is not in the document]
plagoa: By now, I should be getting an electric shock everytime I do this 🤦‍♂️

packages/lab/src/QueryBuilder/Rule/Operator/Operator.js
plagoa: Updated the PR with this change since it can be considered part of this work.

zettca: awesome!

zettca (approved):

HQFOX (approved):

PR #2897, opened 2022-08-18
feat(ActionsGeneric): add iconButton variant

zettca:
Adds the possibility to have `ActionsGeneric` with buttons of variant `icon`.

Passing the `icon` prop isn't possible, because it's being used for something else. Also, this also implies that the label isn't visible (`aria-label` instead)

@lumada-design/imperial Let me know what you think.

### DIFF
--- a/packages/core/src/ActionsGeneric/ActionsGeneric.js
+++ b/packages/core/src/ActionsGeneric/ActionsGeneric.js
@@ -15,6 +15,7 @@ const ActionsGeneric = ({
   actions = [],
   actionsCallback,
   maxVisibleActions = Infinity,
+  iconButtons,
   ...others
 }) => {
   if (!Array.isArray(actions)) return isValidElement(actions) ? actions : null;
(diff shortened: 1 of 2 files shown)


packages/core/src/ActionsGeneric/ActionsGeneric.js
```diff
@@ -28,15 +29,17 @@ const ActionsGeneric = ({
...
         className={classes.button}
         disabled={actDisabled ?? disabled}
         onClick={(event) => actionsCallback?.(event, id, action)}
-        startIcon={renderedIcon}
+        startIcon={iconButtons ? undefined : renderedIcon}
+        aria-label={iconButtons ? label : undefined}
         {...other}
       >
-        {label}
+        {iconButtons ? renderedIcon : label}
```
eliofreitas, line 42: To improve this, we could use the label as a condition to understand whether the user wants an icon button or a button with text and icon, by switching between` label == null ? renderedIcon : label`, with this we can avoid increasing a new prop to the API check: https://github.com/lumada-design/hv-uikit-react/commit/e2497590742d7cac812e2d4314c5f49357299b1a

eliofreitas (changes requested):

nantunes:
...

zettca:
I've tried the suggested alternative, and it didn't quite work as intended.

Lead to https://github.com/lumada-design/hv-uikit-react/issues/2929

I'll close this for now, and reopen when I have the time to work on it and come up with another solution

PR #2910, opened 2022-08-25
fix(BulkActions): select all color & string

zettca:
WIP DS review/release


zettca:
Closing this for now, as this is only likely to be released in dS v5

PR #2918, opened 2022-09-02
feat(chart): add custom tooltips to charts. MLAAS-1523

rubenbarroshv:


stephaniegaspar (approved):

packages/viz/src/Chart/Chart.d.ts
```diff
@@ -4,10 +4,22 @@ import { HvPlotProps } from "./Plot";
...
    */
   tooltipType?: "single" | "multiple";
+  /**
+   *
+   */
+  customTooltip?: React.ReactNode;
+  /**
+   * Custom object containing any value to send to the customTooltip
+   */
+  tooltipData?: Record<string, unknown>;
```
eliofreitas, line 22: Greetings @rubenbarroshv thank you for your contribution, to be able to merge we have 3 points for you to improve

1- Missing documentation in line 15 - 17,
2- we could shorten the name `customTooltip` to only tooltip because there is no other prop using that name, just specify in the documentation that it is an override
3- maybe we could eliminate the `tooltipData `Prop and just pass the data embedded in the component, unless the prop needs to be somehow mapped with an internal value of the chart

eliofreitas (changes requested):

packages/viz/src/Chart/Chart.d.ts
rubenbarroshv: I was trying to use the tooltipData before, but you are right, I don't need it anymore. If I will need it again I will propose another update 😄

eliofreitas (approved):

francisco-guilherme (approved):

PR #2926, opened 2022-09-05
chore: migration guide, docs review, cherry picks

francisco-guilherme:


MIGRATION_v4x.md
```diff
@@ -0,0 +1,158 @@
...
+- [Components](#components)
+- [What's next?](#whats-next)
+
+# From v3 to v4
+
+Having reached a point of stability and maturity we have approached v4 mostly as a **technical release**, with a focus on cleanup and organization in order to **unlock the next stage of growth** (streamline the adoption of new versions of the Design System and improve the customization experience).
+
+## MUI 5 Migration
+
+We have been following [MUI](https://mui.com/) for a long time and using many of the solutions provided since day one. Our approach has always been to use what we find value and unlock the remaining features for our users. Following the same spirit we decided that it was important to upgrade to the latest MUI v5.
```
eliofreitas, line 18: Suggestion: We have been following [MUI](https://mui.com/) for a long time and using many of the solutions provided since day one. Our approach has always been to use what we find _**valuable and enable access to the remaining features for our users**_. Following the same spirit we decided that it was important to upgrade to the latest MUI v5.
_**Continuing with this spirit we decided**_ that it was important to upgrade to the lastest MUI V5

MIGRATION_v4x.md
```diff
@@ -0,0 +1,158 @@
...
+
+- **Styling Solution** - This is our biggest coupling point with MUI today since we are leveraging entirely their [JSS styling system](https://mui.com/system/styles/basics/). For many reasons MUI 5 moved from [JSS to emotion](https://mui.com/blog/mui-core-v5/#migration-from-jss-to-emotion), but since we could keep using JSS, we thought it would be a perfect compromise for now, since it gives us the necessary time to do our own evaluation.
+
+### Update MUI packages
+
+Install the Material UI v5 packages.
```
eliofreitas, line 24: Somewhere around this section we should highlight the changes to the typescript definition of the theme, each project may need to create an override of the material theme to include our .hv value otherwise the app may work but typescript is not going to recognize the theme

packages/core/src/Provider/Provider.js
```diff
@@ -182,32 +183,18 @@ HvProvider.propTypes = {
...
-   * Defaults to `false`. Will be removed in the next major release.
+   * By default the baseline styles are applied globally to the application.
+   * If you need to scope the CSS to avoid styling conflicts, you can set this prop to `"scoped"`.
+   * If you are providing the baseline styles, you can set this prop to false.
    *
-   * @see https://lumada-design.github.io/uikit/master/?path=/docs/foundation-css-baseline--main
+   * @see https://lumada-design.github.io/uikit/master/?path=/docs/theme-css-baseline--page
    */
-  disableCssBaseline: PropTypes.bool,
+  cssBaseline: PropTypes.oneOf(["global", "scoped", false]),
```
eliofreitas, line 197: Maybe instead of false we could use "none" so it is more inline with the enum typing instead of enum + bool

PR #2927, opened 2022-09-06
fix(chart): fix custom tooltips ts definition. MLAAS-1523

rubenbarroshv:


packages/viz/src/Chart/Chart.d.ts
```diff
@@ -3,6 +3,17 @@ import { HvPlotProps } from "./Plot";
...
+type TooltipElementObject = {
+  color: string;
+  name: string;
+  value: number | string;
+};
+
+type TooltipDataObject = {
+  title: string;
+  elements: TooltipElementObject[];
+};
```
eliofreitas, line 15: Export both types, so people can use it if needed

rubenbarroshv: fixed on both cases 😄

eliofreitas (approved):

francisco-guilherme (approved):

PR #2939, opened 2022-09-22
Chore: Update TableHooks.stories.js

AJDowdsHV:


francisco-guilherme (approved):

nantunes (commented):
Looks like `display: contents` browser support is still incomplete. https://caniuse.com/css-display-contents

In particular on Safari HTML Tables are not accessible with display: contents applied (WebKit bugs 239479 and 239478) and before version 16 it has many severe implementation bugs that renders content inaccessible (some older Apple devices are no longer eligible for updates and stuck with Safari 15.x).

I would not promote for now this usage in the UI Kit documentation. You're free to use it in your app, if the partial support coverage is high enough for your use case.

AJDowdsHV:
> Looks like `display: contents` browser support is still incomplete. https://caniuse.com/css-display-contents
> 
> In particular on Safari HTML Tables are not accessible with display: contents applied (WebKit bugs 239479 and 239478) and before version 16 it has many severe implementation bugs that renders content inaccessible (some older Apple devices are no longer eligible for updates and stuck with Safari 15.x).
> 
> I would not promote for now this usage in the UI Kit documentation. You're free to use it in your app, if the partial support coverage is high enough for your use case.

Implementing it as it's currently shown in the docs just caused an error for me so I thought this would be preferable to that but no problem if not, feel free to close if you don't want to take it.

PR #2950, opened 2022-10-25
feat(WarningText): add ARIA role

zettca:
While browsing through the accessibility tree, trying to find a good selector for the `HvWarningText`, I noticed it didn't have any semantic role. Looking at the ARIA roles, it seems like it's missing the [`status` ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role):

> The status role defines a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) containing advisory information for the user that is not important enough to be an [alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role).

Furthermore, the `status` role already has the:
> implicit [aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live) value of polite and an implicit [aria-atomic](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-atomic) value of true


nantunes (approved):
LGTM. Just unsure about what is the codedx commit and how it relates...

zettca:
> LGTM. Just unsure about what is the codedx commit and how it relates...

It's because of the force-pushes to `master` ☠️

francisco-guilherme (approved):

Issue #2960, opened 2022-11-07
Characters "_" or "-" aren't displayed on listRoute.label prop of HvBreadCrumb


stephaniegaspar:
### Latest version

- [X] I have tested the latest version

### No Hitachi Vantara confidential content

- [X] I'm not including any Hitachi Vantara confidential content

### Current behavior 😯

On the following sample case:
`<HvBreadCrumb listRoute=[{label: "_something-", path: "/"}] />`,  the label of the first element defined on `listRoute` will appear "something", not "_something-".

### Expected behavior 🤔

Should appear "_something-".

### Steps to reproduce 🕹

Steps:

1. Go to [lumada-ml-model-management-ui-app](https://dev-mlaasdev.hitachi-lumada.io/hitachi-solutions/lumada-ml-model-management/lumada-ml-model-management-ui-app/projects/5d960d27-815e-4523-a57a-63ff3e61f5e5/models/b094dc72-f5dc-446c-aaeb-d0d9b613c350)
2. Notice that HvBreadCrumb is rendered with no text, and it should be rendered like _______ > _

![image](https://user-images.githubusercontent.com/108751625/200371187-8f5c775f-2937-4140-8e68-d6ab0736482c.png)

### Your Environment 🌎

Device: Laptop
OS: Windows 10
Browser: Chrome 
107.0.5304.88

francisco-guilherme:
Hi @stephaniegaspar, tks for reaching out.

I don't recall the reason behind this behaviour, but your requirement seems to make sense.
We'll refine this internally and move to our backlog once prioritised.

Is this something critical on your side? If so, do you have the time to contribute with the fix?

stephaniegaspar:
Hi @francisco-guilherme,
In our side, the issue is still on backlog but I have time to fix it.
If you already created the issue on your backlog, you can assign it to me.

francisco-guilherme:
If you have the time, please go ahead with the fix. It would be very helpful.

stephaniegaspar:
Okay 👍

stephaniegaspar:
https://github.com/lumada-design/hv-uikit-react/pull/2986

PR #2986, opened 2022-11-23
feat(breadcrumb): display other type of characters on breadcrumb label (#2960)

stephaniegaspar:


packages/core/src/BreadCrumb/BreadCrumb.js
```diff
@@ -3,10 +3,9 @@ import PropTypes from "prop-types";
 import clsx from "clsx";
 import { withStyles } from "@mui/styles";
 import { DropRightXS } from "@hitachivantara/uikit-react-icons";
-import startCase from "lodash/startCase";
 import isNil from "lodash/isNil";
 import { HvLink, HvTypography, HvOverflowTooltip } from "..";
-import { pathWithSubMenu, removeExtension } from "./utils";
+import { pathWithSubMenu, removeExtension, startCase } from "./utils";
```
zettca, line 8: I'm unsure if the original intent was to also remove `_.-` or just to capitalise, but `startCase` is also removing any other (it seems like) non-ASCII character, making it harder or impossible to correctly add i18n to the component. Something like `Bênção?` turns into `Bencao` 🤣

Instead of changing the behaviour for these 2 `_.` cases, I think we should "solve" it for all cases. `startCase` should be removed all together, in favour of `textTransform: "capitalize"` which does the same capitalisation.

This way the user has control over which characters are filtered or not, and can even customise/disable the capitalisation by overriding the styles

stephaniegaspar: On my last push,
Removed the use of `startCase` function of lodash library and `removeExtension` function. 
Added css style: `textTransform: capitalize`.

packages/core/src/BreadCrumb/BreadCrumb.js
```diff
@@ -17,8 +16,13 @@ const Page = ({ Component, onClick, elem, classes }) => (
     data={elem}
     classes={{ a: classes.a }}
   >
-    <HvTypography noWrap component="div" variant="highlightText" className={classes.link}>
-      <HvOverflowTooltip data={startCase(elem.label)} />
+    <HvTypography
+      noWrap
+      component="div"
+      variant="highlightText"
+      className={clsx(classes.link, classes.label)}
```
zettca, line 23: I would reuse the existing classes (not growing the API), but this is fine too 👍

zettca (approved):
#LessIsMore. LGTM 👌🚀

francisco-guilherme (approved):

PR #3011, opened 2022-12-13
feat(snackbar): add warning styling

rubenbarroshv:


andrew-cockayne:
As a general comment, though minor, I think "warning" should come before "error" in all cases, as typically, error is described last. Order consistency has value, so we should have `default`, `success`, `warning` and then `error`.

stephaniegaspar (approved):

PR #3021, opened 2022-12-16
feat(VerticalNavigation): add expand / collapse vertical navigation

andrefrua:
- This is the expandable / collapsable Vertical Navigation component as it is in the Control Plane


packages/lab/src/VerticalNavigation/VerticalNavigation.js
```diff
@@ -0,0 +1,144 @@
...
+  HvVerticalNavigationTree,
+  HvVerticalNavigationActions,
+  HvButton,
+  setId,
+} from "@hitachivantara/uikit-react-core";
+import { Backwards, Forwards } from "@hitachivantara/uikit-react-icons";
+
+import styles from "./styles";
+
+const HvVerticalNavigation = ({
```
zettca, line 17: I might be missing something, but isn't this an example usage of the `HvVerticalNavigation` component? This doesn't add any feature that `HvVerticalNavigation` doesn't already support. Why isn't this a sample in the existing core component?
Besides, the existing "expand/collapse" examples marked as outdated. If this is the "recommended" approach, then I'd be a nice addition 👌

andrefrua: Currently we are waiting for some feedback from DS regarding the Collapse / Expand feature and also what should happen to the TreeView when in Collapsed mode. 
Even though the Collapse / Expand feature is marked as outdated in the UI-KIT, the feature is present in LDOS as required by the pattern we got at the time.

For now, and until we have all the information, this PR will remain in Draft, but I agree with you that this should simply be a sample inside the VerticalNavigation option of the Core section.

andrefrua: There are some issues when the panel is collapsed with multiple levels of item present in the menu, we reached out to PD team and are waiting for some feedback on this matter.
However and since there are other teams in need of this component, and their use case doesn’t require multiple levels on the menu we decided along with the UI-KIT team to merge the PR as it is, and review the component after we get feedback from PD.

andrefrua:
There are some issues when the panel is collapsed with multiple levels of item present in the menu, we reached out to PD team and are waiting for some feedback on this matter.
However and since there are other teams in need of this component, and their use case doesn’t require multiple levels on the menu we decided along with the UI-KIT team to merge the PR as it is, and review the component after we get feedback from PD.

francisco-guilherme (approved):

PR #3032, opened 2022-12-29
feat(panel): add panel component

HQFOX:
In the previous component it was possible to add certain styles directly like so: `<HvPanel width="100%" height="200px" boxShadow={theme.hv.shadows[1]}>`

In this version the styles are added with a style object like so: 
` <HvPanel style={{ width: "100%", height: "200px", boxShadow: theme.shadows.sm,  }}>`
I don't know if it is worth it to add those style props.

### DIFF
--- a/packages/core/src/components/Panel/Panel.tsx
+++ b/packages/core/src/components/Panel/Panel.tsx
@@ -1,0 +1,38 @@
+import clsx from "clsx";
+import { HvBaseProps } from "../../types";
+import { StyledDiv } from "./Panel.styles";
+
+/**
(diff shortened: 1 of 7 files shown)


packages/core/src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,38 @@
+import clsx from "clsx";
+import { HvBaseProps } from "index";
```
MEsteves22, line 2: We need to import from `"../../types"` for the props to appear in the storybook.

MEsteves22 (approved):

plagoa (approved):

PR #3049, opened 2023-01-09
feat(Focus): add Focus component

plagoa:


HQFOX (approved):

packages/core/src/components/Header/Navigation/MenuItem/MenuItem.tsx
```diff
@@ -2,7 +2,7 @@ import { useContext } from "react";
 
 import { MenuBar, NavigationItemProp, Typography } from "components";
 import { HvBaseProps } from "types";
-import { isKeypress, keyboardCodes } from "utils";
+import { isKeypress, KeyboardCodes } from "utils";
```
francisco-guilherme, line 5: @plagoa `KeyboardCodes ` should be kept as camelCase, right?

francisco-guilherme (changes requested):
@plagoa `KeyboardCodes` references should be kept as camelCase, right?

francisco-guilherme (approved):

PR #3057, opened 2023-01-11
fix: path to the HvBaseProps in the components

plagoa:


packages/core/src/components/Avatar/Avatar.tsx
```diff
@@ -3,7 +3,7 @@ import { theme } from "@hitachivantara/uikit-styles";
 import { AvatarProps as MuiAvatarProps } from "@mui/material/Avatar";
 import clsx from "clsx";
 import { CSSProperties } from "react";
-import { HvBaseProps } from "types";
+import { HvBaseProps } from "../../types";
 import { useImageLoaded } from "../../hooks";
```
francisco-guilherme, line 7: here the import should be from alias hooks

packages/core/src/components/Avatar/Avatar.tsx
```diff
@@ -3,7 +3,7 @@ import { theme } from "@hitachivantara/uikit-styles";
...
 import { decreaseSize } from "../../utils";
```
francisco-guilherme, line 8: here the import should be from alias utils

francisco-guilherme (changes requested):

packages/core/src/components/Avatar/Avatar.tsx
plagoa: PR updated.

packages/core/src/components/Avatar/Avatar.tsx
plagoa: PR updated.

plagoa:
@francisco-guilherme updated PR with changes requested.

francisco-guilherme (approved):

PR #3058, opened 2023-01-11
fix: remove static width and height from the button when in icon mode

plagoa:
I don't think we need this here. The button will take whatever size the icon inside has and we might need icons bigger than 32px (check the `Avatar` stories).

### DIFF
--- a/packages/core/src/components/Button/Button.styles.ts
+++ b/packages/core/src/components/Button/Button.styles.ts
@@ -249,8 +249,6 @@ export const StyledButton = styled("button")(
     ...(iconOnly && {
       margin: 0,
       padding: 0,
-      width: theme.sizes.sm,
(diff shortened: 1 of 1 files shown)


francisco-guilherme (approved):

HQFOX:
the static width and height were there because without it the size of the icon only button is 32 x 36 instead of the specified 32 x 32, but it's a small amount i don't know if it makes any practical difference

plagoa:
> the static width and height were there because without it the size of the icon only button is 32 x 36 instead of the specified 32 x 32, but it's a small amount i don't know if it makes any practical difference

Hmm, maybe there's another we can work around this to achieve the same result without removing these props. I'll check that later.

francisco-guilherme (approved):

PR #3061, opened 2023-01-12
feat(ListContainer): add ListContainer component

plagoa:


HQFOX (approved):

packages/core/src/components/ListContainer/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+        interactive && classes?.interactive,
+        interactive && "interactive",
+        selected && classes?.selected,
+        selected && "selected",
+        disabled && classes?.disabled,
+        disabled && "disabled",
+        startAdornment != null && classes?.withStartAdornment,
+        startAdornment != null && "startAdornment",
+        endAdornment != null && classes?.withEndAdornment,
+        endAdornment != null && "endAdornment"
```
MEsteves22, line 191: It think it should be endAdornment != null && "withEndAdornment".

packages/core/src/components/ListContainer/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+        condensed && classes?.condensed,
+        condensed && "condensed",
+        interactive && classes?.interactive,
+        interactive && "interactive",
+        selected && classes?.selected,
+        selected && "selected",
+        disabled && classes?.disabled,
+        disabled && "disabled",
+        startAdornment != null && classes?.withStartAdornment,
+        startAdornment != null && "startAdornment",
```
MEsteves22, line 189: I think it should be startAdornment != null && "withStartAdornment".

packages/core/src/components/ListContainer/ListItem/ListItem.styles.tsx
```diff
@@ -0,0 +1,95 @@
...
+    ...($startAdornment && {
+      "& > div": {
+        float: "left",
+      },
+
+      "& svg": {
+        boxShadow: "none !important",
+        outline: "none !important",
+      },
+      ".disabled>& svg *.color0": {
```
MEsteves22, line 79: The fill color is not changing when the item is disabled. I think the "disabled" class is not being attached to the start adornment so this selector is not working. The same is happening for the end adornment.

packages/core/src/components/ListContainer/ListItem/ListItem.styles.tsx
```diff
@@ -0,0 +1,95 @@
...
+      },
+      "&.disabled": {
+        cursor: "not-allowed",
+      },
+    }),
+    ...($selected && {
+      backgroundColor: theme.list.hoverColor,
+    }),
+    ...($disabled && {
+      color: theme.colors.atmo5,
```
MEsteves22, line 68: In DS5, the background color also changes when the item is disabled.

plagoa:
@MEsteves22 thanks for the feedback. I made the necessary changes. It should be good now.

plagoa:
@MEsteves22 PR updated!

packages/core/src/components/ListContainer/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+        clsx("startAdornment", disabled && "disabled"),
+        classes?.startAdornment
+      ),
+    [classes?.startAdornment, disabled, handleOnClick, selected, startAdornment]
+  );
+  const clonedEndAdornment = useMemo(
+    () =>
+      applyClassNameToElement(
+        endAdornment,
+        "endAdornment",
```
MEsteves22, line 155: We should also change this to `clsx("endAdornment", disabled && "disabled")` in order for the end adornment to change its color when disabled.

packages/core/src/components/ListContainer/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+    </StyledListItem>
+  );
+
+  return interactiveProp ? (
+    <HvFocus
+      rootRef={topContainerRef}
+      selected={selected}
+      disabledClass={disabled || undefined}
+      strategy={role === "option" ? "listbox" : "menu"}
+      classes={{ focus: `background-color: ${theme.colors.atmo3}; z-index: 2` }}
```
MEsteves22, line 214: I think we also need to pass `classes.focus`. Otherwise, the users won't be able to override or extend the focus style.

plagoa: ah, you're right.

PR #3075, opened 2023-01-19
feat(baseRadio): add base radio component #HVUIKIT-6543

HQFOX:


plagoa (approved):

packages/styles/src/themes/ds5Theme.ts
```diff
@@ -78,6 +78,12 @@ const ds5Theme = makeTheme((theme) => ({
     baseCheckBox: {
       hoverColor: theme.colors.acce2s,
     },
+    baseRadio: {
```
MEsteves22, line 81: DS3 and DS5 have the same values here. I think the hover color in [DS5](https://designsystem.hitachivantara.com/6c705d900/v/30842/p/298efd-selection/b/411129/i/119173171) is `acce2s` and has a border radius of `2px`.

packages/core/src/components/BaseRadio/BaseRadio.styles.tsx
```diff
@@ -0,0 +1,38 @@
...
+  MuiRadio,
+  transientOptions
+)(({ $focusVisible }: { $focusVisible: boolean }) => ({
+  padding: theme.baseRadio.padding,
+  cursor: "pointer",
+  "&:hover": {
+    backgroundColor: theme.baseRadio.hoverColor,
+  },
+  borderRadius: 0,
+  "&:disabled": {
```
MEsteves22, line 20: For some reason, this styling is not being applied. I think the root element doesn't have the `disabled` attribute for this selector to be applied. But I think passing this inside `MuiRadioClasses.disabled` should work.

HQFOX: You are right, fixed it in this next push

PR #3084, opened 2023-01-23
feat(BaseInput): add BaseInput component

plagoa:


HQFOX (approved):

packages/core/src/components/BaseInput/BaseInput.styles.tsx
```diff
@@ -0,0 +1,257 @@
...
+    width: "100%",
+    borderRadius: "2px",
+    border: `1px solid ${theme.baseInput.borderColor}`,
+    backgroundColor: theme.colors.atmo1,
+    "&.inputRootInvalid": {
+      border: `1px solid ${theme.baseInput.invalidColor}`,
+    },
+    "&.inputRootReadOnly": {
+      border: `1px solid ${theme.colors.atmo5}`,
+    },
```
MEsteves22, line 175: There's a border around the input when it's readonly but it's only supposed to appear in DS5 and not DS3. Furthermore, in [DS5](https://designsystem.hitachivantara.com/6c705d900/v/30842/p/6700f7-input/b/50959e/i/183325454) it seems like the input has the same `atmo2` background color as the disabled state when it's readonly.

plagoa: Yep, I was working on it. PR updated.

MEsteves22 (approved):

PR #3085, opened 2023-01-23
feat(baseSwitch): add base switch component #HVUIKIT-6544

HQFOX:


plagoa (approved):

packages/core/src/components/BaseSwitch/BaseSwitch.styles.tsx
```diff
@@ -0,0 +1,99 @@
+import styled from "@emotion/styled";
+import { theme } from "@hitachivantara/uikit-styles";
+import { Switch as MuiSwitch, switchClasses } from "@mui/material";
+import { outlineStyles } from "utils";
+import { transientOptions } from "utils/transientOptions";
+
+export const StyledSwitch = styled(
```
MEsteves22, line 7: I think there's only some styling missing from [DS5](https://designsystem.hitachivantara.com/6c705d900/v/30842/p/298efd-selection/b/37978b/i/119173172), namely the track color seems to be `atmo1` when it's not selected, there's a blue hover background, and the disabled state is different. There's also a box shadow on the thumb that was added by Mui that we should remove for DS3 and DS5.

MEsteves22 (approved):

PR #3086, opened 2023-01-25
feat(Input): add Input component

plagoa:


MEsteves22 (approved):

packages/core/src/components/Input/Input.stories.tsx
```diff
@@ -0,0 +1,649 @@
...
+    );
+  },
+};
+
+export const Email: StoryObj<HvInputProps> = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "Required Input that only accepts numbers and limits the maximum length to 5.",
```
MEsteves22, line 379: Wrong description: "Required Input that only accepts valid emails.".

plagoa: Fixed!

MEsteves22 (approved):

PR #3098, opened 2023-02-02
feat(DropDownMenu): add DropDownMenu component

plagoa:


packages/core/src/components/BaseDropdown/BaseDropdown.styles.tsx
```diff
@@ -183,8 +183,12 @@ export const StyledExtension = styled(
     $backgroundColor: string;
```
MEsteves22, line 183: We need to remove `$backgroundColor` from the types definition and from the `StyledExtension` components used inside the `BaseDropdown.tsx` file.

packages/core/src/components/DropDownMenu/DropDownMenu.tsx
```diff
@@ -0,0 +1,188 @@
...
+  /** The variant to be used in the header. */
+  category?: HvButtonVariant;
+  /** A Jss Object used to override or extend the styles applied to the empty state component. */
+  classes?: HvDropDownMenuClasses;
+};
+
+/**
+ * A drop-down menu is a graphical control element, similar to a list box, that allows the user to choose a value from a list.
+ */
+export const HvDropDownMenu = ({
```
MEsteves22, line 67: We need to wrap the component with `withId` since the [master branch](https://github.com/lumada-design/hv-uikit-react/blob/master/packages/core/src/DropDownMenu/DropDownMenu.js#L224) does it too. 
I also had to use this wrapper for the `FileUploader` component. Otherwise the accessibility was breaking since the component did not have an id when it was not provided.

MEsteves22 (approved):

PR #3108, opened 2023-02-06
fix: review Typography component

plagoa:


francisco-guilherme (approved):

MEsteves22:
Everything seems good 🙏 I think we just need to remove unwanted text styling from the `baseDropdown` and `baseInput` components.

plagoa:
@MEsteves22 PR updated with those two components.

MEsteves22 (approved):

PR #3127, opened 2023-02-15
feat(dropdown): add dropdown component

HQFOX:


packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+    return labels?.select || !multiSelect ? (
+      <StyledTypography
+        as="div"
+        $selectionDisabled={disabled}
+        variant={isOpen || hasSelection ? "normalText" : "placeholderText"}
+        className={clsx(
+          classes?.placeholder,
+          disabled && classes?.selectionDisabled
+        )}
+      >
```
MEsteves22, line 456: We also need to add: `dropdownClasses.placeholder` and `disabled && dropdownClasses.selectionDisabled`.

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+        {selectionLabel.selected}
+      </StyledTypography>
+    ) : (
+      <StyledTypography
+        as="div"
+        $selectionDisabled={disabled}
+        className={clsx(
```
MEsteves22, line 463: We also need to add: `dropdownClasses.placeholder` and `disabled && dropdownClasses.selectionDisabled`.

app/src/layout/ThemeSwitcher/ThemeSwitcher.tsx
```diff
@@ -1,29 +1,30 @@
...
+  width: 150,
+});
 
 export const ThemeSwitcher = () => {
   const { selectedTheme, selectedMode, colorModes, themes, changeTheme } =
     useContext(HvThemeContext);
 
   return (
     <HvBox sx={{ display: "flex", gap: 20 }}>
-      <HvDropdown
```
MEsteves22, line 10: The `HvDropdown` component is also used in the stories associated with the theming. Thus, we should also apply these fixes to the `.storybook/docs/Theme/Theming/ThemeSwitcher` and  `.storybook/docs/Theme/Theming/ThemeStructures` files.

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+      status={validationState}
+      disabled={disabled}
+      readOnly={readOnly}
+      required={required}
+      className={clsx(dropdownClasses.root, className, classes?.root)}
+      $selectionDisabled={disabled}
+      {...others}
+    >
+      {(hasLabel || hasDescription) && (
+        <StyledLabelContainer className={classes?.labelContainer}>
```
MEsteves22, line 509: We should also add `dropdownClasses.labelContainer`.

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+          {hasLabel && (
+            <StyledLabel
+              id={setId(elementId, "label")}
+              label={label}
+              className={clsx(classes?.label, classes?.label)}
```
MEsteves22, line 514: The classes are duplicated. I think one of them should be from `dropdownClasses`.

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+            />
+          )}
+
+          {hasDescription && (
+            <HvInfoMessage
+              id={setId(elementId, "description")}
+              className={clsx(classes?.description, classes?.description)}
```
MEsteves22, line 521: The classes are duplicated. I think one of them should be from `dropdownClasses`.

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+            >
+              {description}
+            </HvInfoMessage>
+          )}
+        </StyledLabelContainer>
+      )}
+      <StyledDropdown
+        id={setId(id, "dropdown")}
+        classes={{
+          root: clsx(classes?.dropdown, classes?.dropdown),
```
MEsteves22, line 531: The classes are duplicated. I think one of them should be from `dropdownClasses`.

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,77 +1,615 @@
...
+          arrow: clsx(classes?.arrow, classes?.arrow),
```
MEsteves22, line 532: The classes are duplicated. I think one of them should be from `dropdownClasses`.

packages/core/src/components/Dropdown/dropdownClasses.tsx
```diff
@@ -0,0 +1,54 @@
+import { getClasses } from "utils";
```
MEsteves22, line 1: Maybe it would be better to separate the classes for the dropdown and dropdown list in different files to maintain consistency 🙏 That's what we did with the AppSwitcher for example.

MEsteves22 (commented):
I noticed the following when I was looking at the storybook:
- The dropdown has a border in the "Read Only" story in DS3 when it shouldn't have one
- In the "Single Selection No Selection" story, the placeholder font size increases when clicking on the dropdown. Also, when I select the option with the long text, the text overflows the dropdown
- The placeholder color should be lighter

MEsteves22 (approved):

PR #3167, opened 2023-03-01
fix: load UI Kit styles first to enable styling overrides

MEsteves22:
- `CacheProvider` from emotion is used to prepend the UI Kit styles;
- `react-google-charts` removed from the core package and switched to the project root.

### DIFF
--- a/packages/core/src/providers/Provider.tsx
+++ b/packages/core/src/providers/Provider.tsx
@@ -1,8 +1,9 @@
-import { css, Global } from "@emotion/react";
+import { css, Global, CacheProvider } from "@emotion/react";
 import { CssBaseline, getThemesVars } from "@hitachivantara/uikit-styles";
 import { parseThemes } from "utils";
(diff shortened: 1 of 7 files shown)


plagoa (approved):

packages/core/src/providers/Provider.tsx
```diff
@@ -63,7 +64,16 @@ export const HvProvider = ({
...
 
   return (
-    <>
+    <CacheProvider
+      /**
+       * Moves UI Kit styles to the top of the <head> so they're loaded first.
+       * This enables users to override the UI Kit styles if necessary.
+       */
+      value={createCache({
+        key: "css",
```
nantunes, line 73: Maybe something more unique to UI Kit? To avoid conflicts it the app uses its CacheProvider.

MEsteves22: Yes, you are right 🙏 I changed it to `hv-uikit-css`.

PR #3170, opened 2023-03-03
chore: update CSS vars names. #HVUIKIT-6613

plagoa:
- the only relevant file for this is the `utils.ts` file, all the rest are the updated snapshots.


packages/styles/src/utils.ts
```diff
@@ -15,12 +15,13 @@ const toCSSVars = (obj: object, prefix = "-") => {
     }
   }
 
+  console.log("toCSSVars", vars);
```
MEsteves22, line 18: `console.log` to remove 😛

packages/styles/src/utils.ts
```diff
@@ -32,6 +33,7 @@ export const mapCSSVars = <T extends object>(
     }
   }
 
+  console.log("mapCSSVars", vars);
```
MEsteves22, line 36: `console.log` to remove 😛

MEsteves22 (approved):

plagoa:
@MEsteves22 removed the `console.log` !

PR #3175, opened 2023-03-06
feat(calendar): add calendar component

HQFOX:


packages/core/src/components/Calendar/calendarClasses.tsx
```diff
@@ -0,0 +1,26 @@
...
+  "rangeCalendarContainer",
+  "singleCalendar",
+  "focusSelection",
+  "calendarMonthlyCell",
+  "calendarMonthlyCellSelected",
+];
+
+const calendarClasses = getClasses<HvCalendarClasses>(
+  classKeys,
+  "HvCalendarClasses"
```
MEsteves22, line 23: Just to maintain some consistency, I think we should use `HvCalendar` since we usually use the component's name here

plagoa (approved):

PR #3178, opened 2023-03-06
chore: update custom theme

plagoa:


HQFOX (approved):

packages/core/src/components/Card/Card.tsx
```diff
@@ -51,10 +51,11 @@ export const HvCard = ({
...
         styles.root,
-        bgcolor &&
-          css({
-            backgroundColor: theme.colors[bgcolor],
-          }),
+        // bgcolor &&
+        css({
+          backgroundColor:
+            theme.card.backgroundColor || (bgcolor && theme.colors[bgcolor]),
+        }),
```
MEsteves22, line 58: I think by doing this the background set in the theme will always be used instead of the one passed to the component through the prop `bgcolor`

plagoa: Ah, you're right. Will fix it!

MEsteves22: Yes, it also makes sense for me

MEsteves22 (approved):

MEsteves22 (approved):

PR #3211, opened 2023-03-14
refactor: theming

MEsteves22:
- `HvProvider` refactored
- `createTheme` utility added 
- `useTheme` hook added 
- Get started and provider documentation reviewed 

Note: The theming documentation will be reviewed on another PR.


packages/core/src/components/BaseDropdown/BaseDropdown.tsx
```diff
@@ -143,7 +142,7 @@ export const HvBaseDropdown = ({
   onContainerCreation,
   ...others
 }: HvBaseDropdownProps) => {
-  const { rootId } = useContext(HvThemeContext);
+  const { rootId } = useTheme();
```
nantunes, line 145: This one seems strange coming from a `useTheme` hook (it was already, coming from a `HvThemeContext`). I understand why, and that it is purely theme/styling related, but for consumers it is more something like the app container.

A `useRoot` returning the `id` and possibly a `ref` for the dom element would be interesting.

nantunes: Also, being in separate contexts would also allow changing the theme without triggering re-renders in components using the root element's id/reference, as it most likely never change.

packages/core/src/hooks/useTheme.ts
nantunes, line 1: If the hook is exposing all the properties in the context, `return useContext(HvThemeContext);` is enough... The `useMemo` is overkill.

nantunes (approved):

packages/core/src/hooks/useTheme.ts
MEsteves22: I changed the hook to not use the `useMemo` hook as you mentioned.

plagoa (approved):

PR #3245, opened 2023-03-27
feat(Theme): add Typography section to theme generator and update generator layout

plagoa:


francisco-guilherme (approved):

MEsteves22:
I'm having some trouble updating the following values for the typography: 
- Font size: unable to change the input value
- Font weight: the dropdown value is not updating (it doesn't happen with all values)
- Line weight: the dropdown value is not updating

There's also a problem with the scrollbar inside the theme generator drawer. If I open all sections except the typography one, I'm unable to scroll all the way down.

francisco-guilherme (approved):

PR #3300, opened 2023-04-27
feat(datePicker): add date picker component

HQFOX:


packages/styles/src/themes/ds3.ts
```diff
@@ -500,6 +500,12 @@ const ds3 = makeTheme((theme: HvTheme) => ({
   multiButton: {
     disabledBackgroundColor: theme.colors.atmo2,
   },
+  datePicker: {
+    panelBorder: `1px solid ${theme.colors.atmo1}`,
+    panelBackgroundColor: theme.colors.atmo1,
+    dropdownHeaderOpenBorder: `1px solid ${theme.colors.atmo1}`,
+    dropdownPlaceholderColor: theme.colors.secondary_80,
```
MEsteves22, line 507: I believe the placeholder color in DS3 is `secondary_60`

docs/overview/migration/MigrationV5.stories.mdx
```diff
@@ -662,3 +662,7 @@ The `theme` property was removed since the component accesses the theme object i
 ### Code editor <a id="code-editor" />
 
 The `theme` property was removed since the component accesses the theme object internally.
+
+### DatePicker <a id="date-picker" />
```
MEsteves22, line 666: Under the "UI Kit packages" section, we should also remove the date picker from the list of components missing for the core package

packages/core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,694 @@
...
+        : theme.datePicker.dropdownPlaceholderColor,
+    });
+
+    return <StyledTypography variant={variant}>{text}</StyledTypography>;
+  };
+
+  const renderInput = (dateString) =>
+    styledTypography(
+      dateString,
+      "label",
```
MEsteves22, line 516: I think we should use `label` for DS5 but `body` for DS3 because the font weight lighter on DS3

docs/overview/migration/MigrationV5.stories.mdx
MEsteves22: No, look for "The following components from `@hitachivantara/uikit-react-core` are missing:" and the date picker will be there

HQFOX: thanks! i missed it

packages/core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,690 @@
...
+        : theme.datePicker.dropdownPlaceholderColor,
+    });
+
+    return <StyledTypography variant={variant}>{text}</StyledTypography>;
+  };
+
+  const renderInput = (dateString) =>
+    styledTypography(
+      dateString,
+      theme.datePicker.placeholderVariant,
```
MEsteves22, line 513: I believe we need to get this from the `activeTheme` in the theme context since `theme.datePicker.placeholderVariant`is a CSS variable

packages/core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,692 @@
...
+        : theme.datePicker.dropdownPlaceholderColor,
+    });
+
+    return <StyledTypography variant={variant}>{text}</StyledTypography>;
+  };
+
+  const renderInput = (dateString) =>
+    styledTypography(
+      dateString,
+      activeTheme?.name == "ds3" ? "body" : "label",
```
MEsteves22, line 515: I think adding the `placeholderVariant` property to the themes like you had before was preferable in this case. This way, if someone uses a custom theme based on DS3 (and the theme name is not ds3), the `label` variant will always be used. However, instead of getting the CSS var `theme.datePicker.placeholderVariant` like before, we should do `activeTheme?.datePicker.placeholderVariant` to get the actual value. What do you think?

HQFOX: Yes I agree, thank you , I'm going change it now

MEsteves22 (approved):

francisco-guilherme (approved):

PR #3302, opened 2023-04-28
chore: add the tutorial

plagoa:
- added tutorial section to the application
- added a link to the tutorial to the header
- uses the UI Kit `Dialog` component as step layout


HQFOX (approved):

packages/core/src/components/Dialog/Dialog.styles.tsx
```diff
@@ -29,7 +29,7 @@ export const StyledPaper = styled(
   "& svg *.color0": {
     fill: theme.colors.secondary,
   },
-  [`&.${MuiPaperClasses.root}`]: {
```
MEsteves22, line 32: By removing `&` the styles are not being applied to the dialog paper. Is this the expected behaviour? The dialog still works because it's using MUI's styles but if these styles are not necessary maybe we should remove them.

PR #3307, opened 2023-05-03
refactor: rename type files to `d.ts`

zettca:
This helps the compiler and devs know that only typings should be declared in these typings files.


zettca:
needs further changes to be parsed. I'll leave this for another time

PR #3327, opened 2023-05-15
chore: review Banner component to allow external styling overrides

plagoa:


packages/core/src/components/Banner/Banner.styles.tsx
```diff
@@ -1,24 +1,22 @@
...
-  Snackbar,
-  transientOptions
-)(({ $isOpen }: { $isOpen: boolean }) => ({
-  [`&.${snackbarClasses.root}`]: {
-    ...($isOpen && {
+export const styles: {
+  root: CSSInterpolation;
+  anchorOriginTopCenter: CSSInterpolation;
+  anchorOriginBottomCenter: CSSInterpolation;
+} = {
```
MEsteves22, line 8: We should use something like `Partial<Record<keyof HvBannerClasses, CSSInterpolation>>` as proposed in the `ColorPicker` PR to avoid repeating the classe names. The same should be done for the remaining files.

packages/core/src/components/Banner/BannerContent/BannerContent.tsx
```diff
@@ -73,52 +75,58 @@ export const HvBannerContent = forwardRef<HTMLDivElement, HvBannerContentProps>(
...
-              {...(effectiveActionsPosition === "bottom-right" && {
-                action: actions,
-                actionCallback: actionsCallback,
-              })}
-              {...actionProps}
+              classes={{
+                root: cx(css(styles.root), classes?.root),
+                message: cx(css(styles.message), classes?.message),
+                action: cx(css(styles.action), classes?.action),
+              }}
```
MEsteves22, line 94: `bannerContentClasses` are missing. I believe it should be:

```
classes={{
       root: cx(bannerContentClasses.root, css(styles.root), classes?.root),
       message: cx(bannerContentClasses.message, css(styles.message), classes?.message),
       action: cx(bannerContentClasses.action, css(styles.action), classes?.action),
 }}
```

packages/core/src/components/Banner/Banner.tsx
```diff
@@ -103,42 +104,43 @@ export const HvBanner = ({
...
-        onClose={onClose}
-        {...bannerContentProps}
-      />
-    </StyledSnackbar>
+    <ClassNames>
+      {({ css, cx }) => (
+        <Snackbar
+          id={id}
+          open={open}
+          className={cx(css(bannerClasses.root), className)}
```
MEsteves22, line 112: Some classes are missing. I also think that we should continue to use the `rootClosed` classe to maintain backward compatibility with the previous API and by doing so we don't need to have the `[`&.open`]` selector.

Maybe something like:

```
className={className}
classes={{
            root: open ? cx(bannerClasses.root, css(styles.root), classes?.root) : cx(bannerClasses.rootClosed, css(styles.rootClosed), classes?.rootClosed),
            anchorOriginTopCenter: cx(
              bannerClasses.anchorOriginTopCenter,
              css(styles.anchorOriginTopCenter),
              classes?.anchorOriginTopCenter
            ),
            anchorOriginBottomCenter: cx(
              bannerClasses.anchorOriginBottomCenter,
              css(styles.anchorOriginBottomCenter),
              classes?.anchorOriginBottomCenter
            ),
}}
```

MEsteves22 (changes requested):

packages/core/src/components/Banner/Banner.styles.tsx
MEsteves22: I was planning on opening a PR to apply this to the other components after the `ColorPicker` is merged.

packages/core/src/components/Banner/Banner.tsx
plagoa: yeah, I wondered about that and thought I could improve, but I'll keep how it was like you suggested.

packages/core/src/components/Banner/Banner.styles.tsx
plagoa: updated.

MEsteves22: You forgot to update the other files: message container, banner content, and action container :P

packages/core/src/components/Banner/Banner.tsx
MEsteves22: Okay, great 🙏 

I believe you forgot to update the `className`. It's currently `className={cx(css(bannerClasses.root), className)}` and should be `className={className}`.

PR #3331, opened 2023-05-16
chore: improve typings for style classes

MEsteves22:


packages/core/src/components/Banner/BannerContent/ActionContainer/ActionContainer.styles.tsx
```diff
@@ -1,12 +1,10 @@
...
 
-export const styles: {
-  actionContainer: CSSInterpolation;
-  actionsInnerContainer: CSSInterpolation;
-  closeAction: CSSInterpolation;
-  iconContainer: CSSInterpolation;
-} = {
+export const styles: Partial<
+  Record<keyof HvActionContainerClasses, CSSInterpolation>
+> = {
```
zettca, line 7: I don't think `Partial` is necessary after all, but I guess it doesn't hurt 👍 

For slightly better typings, we could leverage the "new" [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html) operator.
⚠️ But I think it's better to leave this for later, and do it in all components. This is a good enough improvement

zettca (approved):
Nice 👌

PR #3337, opened 2023-05-17
fix: install missing `clsx` in lab

zettca:
Was leading to a wrong build file structure:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38619,6 +38619,7 @@
       "dependencies": {
         "@emotion/cache": "^11.10.5",
         "@hitachivantara/uikit-react-core": "^5.6.0",
+        "clsx": "^1.2.1",
         "react-color": "^2.19.3",
         "usehooks-ts": "^2.9.1"
       },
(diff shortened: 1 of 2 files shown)


packages/code-editor/tsconfig.json
```diff
@@ -5,9 +5,7 @@
     "baseUrl": "src",
     "rootDir": "src",
     "outDir": "dist",
-    "paths": {
```
MEsteves22, line 8: We need to keep this alias configuration. Otherwise we'll get a bunch of errors when building the package  🙏

packages/lab/tsconfig.json
```diff
@@ -5,9 +5,7 @@
     "baseUrl": "src",
     "rootDir": "src",
     "outDir": "dist",
-    "paths": {
```
MEsteves22, line 8: We need to keep this alias configuration. Otherwise we'll get a bunch of errors when building the package 🙏

MEsteves22 (changes requested):

MEsteves22 (approved):

PR #3338, opened 2023-05-17
feat: color picker promoted to core

MEsteves22:


packages/core/package.json
```diff
@@ -59,14 +59,16 @@
     "react-popper": "^2.3.0",
     "react-resize-detector": "^8.0.4",
     "react-table": "^7.8.0",
-    "react-window": "^1.8.8"
+    "react-window": "^1.8.8",
+    "react-color": "^2.19.3"
```
zettca, line 63: Can we sort these imports? This way we get simpler diffs the next time we add a package (and maybe fewer conflicts)

This should do the trick:
```
npm remove _dummy
```

francisco-guilherme (approved):

PR #3341, opened 2023-05-17
fix(table): align and update to ds5 icon on expandable rows

HQFOX:
this pr is to fix the miss align expand icon on the expand `useHvExpand` table renderer.
It also fixes some things I noticed along the way:
    - adjust the table cell padding to the requirements that was causing the icon miss align to be noticeable in the first place.
    - change how the styles were being applied to the `TableRow` and `TableCell` components to fix some styles that were not being applied in the `listrow` version of the table.
    - add ds5 styles to `listrow` version o the table.
    - updated the expand icons to the ones in ds5.
    - added icon parameters to `useHvExpand` so that it is possible to comply with ds3.


MEsteves22:
When testing the PR, I found the following:
- In the first table story (Main), the rows should have different colors: "grey" - "white" - "grey" - "white" - etc... 
- The hover color in DS3 should be a "light grey". It's "white" at the moment.
- In the `useGroupBy` story, the sub rows should be "light grey" in DS3.
- In DS5, selected rows should have a "light grey" background color.
- In DS5, table rows usually have a "white" background color. They are "light grey" at the moment.

We should also write a brief explanation in the migration guide regarding the changes with the `useHvExpand`.

packages/core/src/components/Table/TableCell/TableCell.styles.tsx
```diff
@@ -0,0 +1,162 @@
...
+    minHeight: 32,
+    "td&": {
+      height: 32,
+    },
+    backgroundColor: "inherit",
+    ...(theme.typography.body as CSSProperties),
+    fontFamily: theme.fontFamily.body,
+
+    [`&.${tableCellClasses.sorted}`]: {
+      backgroundColor: alpha(hexToRgb(theme.colors.atmo1), 0.4),
```
MEsteves22, line 36: Since `theme.colors.atmo1` is a CSS variable, it won't work with the `hexToRgb` method. This is the background color created:

```
rgba(10, NaN, NaN, 0.4, NaN, -0.047, NaN, NaN, NaN, 0.039, NaN, 0.004);
```

We need to use real values that are available in the React context side. 

This is happening with in order place in the table cell and table row.

MEsteves22 (changes requested):

MEsteves22:
- I noticed the following behaviour for the stripped table (The color is not updating when the active color mode and/or active theme change):

https://github.com/lumada-design/hv-uikit-react/assets/43220251/7e529f4e-44c3-4ddd-a556-a1f926e6e3ab

- I believe we lost the hover color for rows in DS3.

- For the `hvExpandColumn` renderer I would do the following because `@emotion/css` is not a `core` dependency if I'm not mistaken:

```
<ClassNames>
          {({ css }) => (
            <>
              {hasContent && (
                <HvButton
                  icon
                  variant="secondaryGhost"
                  aria-label={
                    row.isExpanded
                      ? collapseRowButtonAriaLabel
                      : expandRowButtonAriaLabel
                  }
                  aria-expanded={row.isExpanded}
                  onClick={expandedProps?.onClick}
                  classes={{ root: css({
                                      position: "absolute",
                                      left: 0,
                                      top: "50%",
                                      transform: "translateY(-50%)",
                                 }) }}
                >
                  {row.isExpanded ? ExpandedIcon : CollapsedIcon}
                </HvButton>
              )}

              <HvOverflowTooltip data={hvStringFallback(value)} />
            </>
          )}
</ClassNames>
```

packages/core/src/components/Table/TableRow/TableRow.tsx
```diff
@@ -37,107 +37,28 @@ const StyledTableRow = (c: any) =>
...
-          borderBottom: "none",
-        },
-        [`&.${tableRowClasses.expanded}`]: {
-          backgroundColor: theme.colors.atmo1,
-        },
-      }),
       ...($striped && {
-        "&:nth-of-type(even)": {
-          backgroundColor: $stripedColor,
+        [`&:nth-of-type(even)`]: {
```
zettca, line 49: We're not using variables, so this doesn't have to be computed. ESLint (sometimes) complains about this

```suggestion
        "&:nth-of-type(even)": {
```

packages/core/src/components/Table/TableRow/TableRow.tsx
```diff
@@ -171,43 +91,75 @@ export const HvTableRow = forwardRef<HTMLElement, HvTableRowProps>(
 
     const TableRow = useMemo(() => StyledTableRow(Component), [Component]);
 
+    const even = getVarValue(theme.table.rowStripedBackgroundColorEven);
+
+    const odd = getVarValue(theme.table.rowStripedBackgroundColorOdd);
```
zettca, line 96: I'm worried about the performance implications here...
We're running `getVarValue`, which manipulates the DOM to get the value, for every render of every (potentially many) row...

Should we be memoizing this somehow?

packages/core/src/components/Table/TableRow/TableRow.tsx
```diff
@@ -171,43 +93,89 @@ export const HvTableRow = forwardRef<HTMLElement, HvTableRowProps>(
...
+      odd = getVarValue(theme.table.rowStripedBackgroundColorOdd);
+
+      stripedColorEven = checkValidHexColorValue(even)
+        ? hexToRgbA(even, 0.6)
+        : even;
+
+      stripedColorOdd = checkValidHexColorValue(odd)
+        ? hexToRgbA(odd, 0.6)
+        : odd;
+    }, [activeTheme?.colors?.modes[selectedMode]]);
```
zettca, line 120: Mutating variables like this can lead to all sorts of problems. We should use `useState`.
Also, `getVarValue` still gets run every render like this.

```suggestion
    const [even, setEven] = useState(getVarValue(theme.table.rowStripedBackgroundColorEven));
    const [odd, setOdd] = useState(getVarValue(theme.table.rowStripedBackgroundColorOdd));

    const stripedColorEven = checkValidHexColorValue(even)
      ? hexToRgbA(even, 0.6)
      : even;

    const stripedColorOdd = checkValidHexColorValue(odd)
      ? hexToRgbA(odd, 0.6)
      : odd;

    useEffect(() => {
      setEven(getVarValue(theme.table.rowStripedBackgroundColorEven));
      setOdd(getVarValue(theme.table.rowStripedBackgroundColorOdd));
    }, [activeTheme?.colors?.modes[selectedMode]]);
```

docs/overview/migration/MigrationV5.stories.mdx
```diff
@@ -739,3 +740,10 @@ const images = ["img1.png", "img2.png", "img3.png"];
 +  {images.map((src) => <HvCarouselSlide src={src} />)}
 +</HvCarousel>
 ```
+
+### Table Renderer - hvExpandColumn <a id="table-renderer-expand-column" />
+
+The default icon for this component has changed to the one specified in DS5 specifications, in order to be possible to revert to the previous icon or customize to any other, the following properties where added:
```
MEsteves22, line 746: "where added" => "were added"

docs/overview/migration/MigrationV5.stories.mdx
```diff
@@ -739,3 +740,10 @@ const images = ["img1.png", "img2.png", "img3.png"];
...
+
+- `isExpandedIcon`
```
MEsteves22, line 748: We need to update these now since the names changed.

packages/core/src/components/Table/TableCell/TableCell.tsx
```diff
@@ -297,78 +127,111 @@ export const HvTableCell = forwardRef<HTMLElement, HvTableCellProps>(
 
     const TableCell = useMemo(() => StyledTableCell(Component), [Component]);
 
+    const sortedColorValue = getVarValue(theme.table.rowSortedColor);
```
MEsteves22, line 130: I believe we are having the same problem here as we had with the stripped table in the `TableRow` component right?

HQFOX: The fix for that, was not working as intended, the first render was not correct, I sort of reverted it in the next commit and suggested  @zettca that maybe is better to open a separate issue to find a better way to deal with that problem

MEsteves22: Okay, makes sense!

francisco-guilherme (approved):

PR #3342, opened 2023-05-17
fix: header hover

MEsteves22:
- The hover was not behaving correctly: when hovering a parent item, the child items were also highlighted. This was detected on our local app. On storybook everything seemed to be fine.


plagoa (approved):

andrefrua:
We also detected this on App Shell. Thanks for the fix 👍

Issue #3314, opened 2023-05-05
HvDialog - blinking background and icon color


brunonovaishv:
### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

- When the `HvDialog` component renders the background properties change (opacity and transition) and this causes a visual effect of the background blinking.
- When using the `HvFileUploader `component inside of `HvDialog`, and after successfully uploading a file, the checkmark icon has the color `--uikit-colors-secondary` (grey color)

### Expected behavior 🤔

- The background of `HvDialog` component shouldn't have that visual effect while render occurs.
- When using the `HvFileUploader `component inside of `HvDialog`, and after successfully uploading a file, the checkmark icon should have the color `--uikit-colors-positive` (green color)

### Steps to reproduce 🕹

Steps:

1. Use a `HvFileUploader` inside of `HvDialog`
2. Upload a file
3. Wait for the file to be successfully uploaded to see the checkmark icon

### Your Environment 🌎

Device: Laptop
OS: Ventura 13.3.1
Browser: chrome
Version: 113.0

plagoa:
@brunonovaishv this has been fixed on [@hitachivantara/uikit-react-core@5.7.4](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.7.4)

PR #3350, opened 2023-05-19
fix(HvDialog): fix blinking background. #3314

stephaniegaspar:


nantunes (changes requested):

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -140,6 +142,22 @@ export const HvDialog = ({
...
+          open={open}
+          onClick={(event) => wrappedClose(event)}
+          $backColor={
+            activeTheme?.colors?.modes[selectedMode].atmo4 || theme.colors.atmo4
+          }
+          {...backdropProps}
+        />
+      ),
+    }),
+    []
```
zettca, line 158: Missing the dependency array! Without it we might get wrong behaviour when `open`, `backdropProps`, `activeTheme`, `onClose` change.

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -140,6 +142,22 @@ export const HvDialog = ({
     ? withTooltip(closeButtonDisplay, buttonTitle, "top")
     : closeButtonDisplay;
 
+  const slots = useMemo<ModalSlots>(
```
zettca, line 145: `ModalSlots` is from `@mui/base` which isn't a direct dependency or peer-dependency.
`MuiDialogProps["slots"]` is more future-proof and is already imported

```suggestion
  const slots = useMemo<MuiDialogProps["slots"]>(
```

packages/core/src/components/Dialog/Dialog.tsx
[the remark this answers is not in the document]
stephaniegaspar: Without dependencies, the `StyledBackdrop` will remain fixed and never be rerendered due to change of `HvDialog` states and also its parent components' states.

stephaniegaspar: Is there another way to maintain the state of the component and make `StyledBackdrop` fixed?

stephaniegaspar: Resolved on my last update 🚀

packages/core/src/components/Dialog/Dialog.tsx
stephaniegaspar: Resolved on my last update 🚀

packages/core/src/components/Dialog/Dialog.tsx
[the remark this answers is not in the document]
nantunes: The dependencies aren't optional. You'll have to chain stabilise each dependency, using useMemo, useCallback or moving it to outside of the function if they're independent of state. https://react.dev/learn/lifecycle-of-reactive-effects#what-to-do-when-you-dont-want-to-re-synchronize

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -140,6 +144,19 @@ export const HvDialog = ({
     ? withTooltip(closeButtonDisplay, buttonTitle, "top")
     : closeButtonDisplay;
 
+  const slots = useMemo<MuiDialogProps["slots"]>(
+    () => ({
+      backdrop: (backdropProps) => (
+        <HvDialogBackdrop
+          open={open}
+          onClick={(event) => wrappedClose(event)}
```
nantunes, line 152:
```suggestion
          onClick={wrappedClose}
```

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -140,6 +144,19 @@ export const HvDialog = ({
...
+          {...backdropProps}
+        />
+      ),
+    }),
+    [wrappedClose]
```
nantunes, line 157: Still missing the `open` dependency.

nantunes (changes requested):

packages/core/src/components/Dialog/Dialog.tsx
stephaniegaspar: 👍

packages/core/src/components/Dialog/Dialog.tsx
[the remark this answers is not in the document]
stephaniegaspar: Agreed. Thank you!

nantunes (approved):

packages/core/src/components/Dialog/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from "./Title";
+export * from "./Backdrop";
```
zettca, line 2: This means that `HvDialogBackdrop` is now a new component (exported at the root level) that's part of the API - We don't want that!

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -15,6 +15,7 @@ import { withTooltip } from "@core/hocs";
 import dialogClasses, { HvDialogClasses } from "./dialogClasses";
 import { useTheme } from "@core/hooks";
 import { ClassNames } from "@emotion/react";
+import { HvDialogBackdrop } from "./Backdrop";
```
zettca, line 18: I think this component doesn't "deserve" it own file structure and to be exported - we only need it internally for `Dialog.tsx`.

We should have it declared there.
But we definitely don't want this exported to the "public API"

zettca (changes requested):

packages/core/src/components/Dialog/Dialog.tsx
stephaniegaspar: I understand. I'll change it again 🤔

packages/core/src/components/Dialog/index.ts
stephaniegaspar: One more try to go...

packages/core/src/components/Dialog/Dialog.tsx
stephaniegaspar: Updated ⬇️

zettca (approved):
👍

PR #3351, opened 2023-05-19
fix(VerticalNavigation): truncate text on slider navigation

plagoa:


packages/core/src/components/VerticalNavigation/NavigationSlider/NavigationSlider.styles.tsx
```diff
@@ -27,3 +33,12 @@ export const StyledListItem = styled(HvListItem)({
     },
   },
 });
+
+export const StyledOverflowTooltip = styled(
+  (props: HvOverflowTooltipProps) => <HvOverflowTooltip {...props} />,
+  transientOptions
+)(({ $hasEndAdornment }: { $hasEndAdornment: boolean }) => ({
+  ...(!$hasEndAdornment && {
+    paddingRight: 8,
```
zettca, line 42:
```suggestion
    paddingRight: theme.space.xs,
```

packages/core/src/components/VerticalNavigation/VerticalNavigation.test.tsx
```diff
@@ -208,6 +208,23 @@ const SliderSample = () => {
...
+      observe: vi.fn(),
+      unobserve: vi.fn(),
+      disconnect: vi.fn(),
+    }));
+  });
+
+  afterEach(() => {
+    window.ResizeObserver = ResizeObserver;
+    vi.restoreAllMocks();
+  });
```
zettca, line 226: I needed this in my PR as well.
I think we should move to the global test setup, as this is part of the DOM APIs that JSDOM currently doesn't mock

zettca: I opened a PR for this https://github.com/lumada-design/hv-uikit-react/pull/3354

plagoa: Nice, I'll leave this as is then.

MEsteves22 (approved):

PR #3352, opened 2023-05-19
fix(VerticalNavigation): ensure no unselected items are marked as selected

plagoa:


packages/core/src/components/VerticalNavigation/TreeView/TreeView.tsx
```diff
@@ -231,7 +231,9 @@ export const HvVerticalNavigationTreeView = forwardRef(
     );
 
     const isChildSelected = useCallback(
-      (id) => selected.startsWith(id),
+      // the second part of the condition is to ensure that the id we're
+      // looking at is actually of a child (ie, there's at least one "-")
+      (id) => selected.startsWith(id) && selected.indexOf("-") !== -1,
```
zettca, line 236:
```suggestion
      (id) => selected.startsWith(id) && selected.includes("-"),
```

zettca (approved):

packages/core/src/components/VerticalNavigation/TreeView/TreeView.tsx
plagoa: I'll update 👍

francisco-guilherme (approved):

PR #3358, opened 2023-05-22
chore(VerticalNavigation): created a property to customize popup styling

andrefrua:
Hi guys,

On App Shell we need to be able to customize the VerticalNavigation popup, at this point we need to customize the maximum height of the popup. The approach we were using was to wrap App Shell in a GlobalStyles and the then force the customization of the popup with this css code:

```
      div[role="tooltip"] {
        max-height: calc(100vh - 64px);
        overflow-y: auto;
      }
```

However, by doing so, we detected that this could cause some issue on the embedded application when using tooltips, we need to get rid of that global style but still be able to customize the Vertical Navigation popup.

The content of this PR opens the possibility for us to customize the styling of the popup, however, we are aware that this is not at all a very elegant solution. If you have a better approach to achieve this, please let us know and we will change the code to the appropriate one. In the meantime it would be really useful for us to have this merged to solve our issue.

### DIFF
--- a/packages/core/src/components/VerticalNavigation/Navigation/Navigation.tsx
+++ b/packages/core/src/components/VerticalNavigation/Navigation/Navigation.tsx
@@ -160,8 +160,6 @@ export const HvVerticalNavigationTree = ({
   defaultSelected,
   onChange,
 
-  popupStyles,
-
   ...others
 }: HvVerticalNavigationTreeProps) => {
   const [selected, setSelected] = useControlled(selectedProp, defaultSelected);
@@ -359,7 +357,10 @@ export const HvVerticalNavigationTree = ({
               data={navigationPopup.data}
               onClose={handleNavigationPopupClose}
               onChange={handleChange}
-              popupStyles={popupStyles}
+              className={clsx(
+                verticalNavigationTreeClasses.navigationPopup,
+                classes?.navigationPopup
+              )}
             />
           )}
           {children}
@@ -436,10 +437,6 @@ export interface HvVerticalNavigationTreeProps
    * target - the behavior when opening an url.
    */
   data?: NavigationData[];
-  /**
-   * Object with the custom css styles to be applied to the popup.
-   */
-  popupStyles?: any;
 }
 
 export type NavigationMode = "treeview" | "navigation" | "slider";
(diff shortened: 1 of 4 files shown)


packages/core/src/components/VerticalNavigation/Navigation/Navigation.tsx
```diff
@@ -357,6 +359,7 @@ export const HvVerticalNavigationTree = ({
               data={navigationPopup.data}
               onClose={handleNavigationPopupClose}
               onChange={handleChange}
+              popupStyles={popupStyles}
```
zettca, line 362: Is overriding inline `styles` really necessary for your use case? If the goal is styling, then we should use classes instead of inline styles. Classes allow for much more customisation, and is our recommended approach for styling components.

I'm referring to something like this:
```suggestion
              className={classes?.popup}
```

packages/core/src/components/VerticalNavigation/Navigation/Navigation.tsx
```diff
@@ -433,6 +436,10 @@ export interface HvVerticalNavigationTreeProps
    * target - the behavior when opening an url.
    */
   data?: NavigationData[];
+  /**
+   * Object with the custom css styles to be applied to the popup.
+   */
+  popupStyles?: any;
```
zettca, line 442: There is no need for `any`; we know exactly what we allow here and can provide much better typings. This should be `CSSProperties`.

When in doubt, hover a prop/attribute, and It'll tell you what type it should be.

image

andrefrua:
@zettca Thanks for your help on this. I've made some changes, let me know if you're ok with this 👍

zettca (approved):
Nice 🚀

MEsteves22 (dismissed):

MEsteves22:
@andrefrua The tests are breaking. Can you update the snapshots that are breaking while running the tests?

MEsteves22 (changes requested):
@andrefrua The tests are currently breaking. Can you rebase you branch, please? We fixed this on the master branch.

Also, nothing major, but can you replace `className={classes?.navigationPopup}` by:

```
className={clsx(verticalNavigationTreeClasses.navigationPopup, classes?.navigationPopup)}
```

The `verticalNavigationTreeClasses.navigationPopup` class is missing. We usually add these static classes in case someone needs to target them in a CSS selector for example.

andrefrua:
@MEsteves22 Snapshots updated 👍

andrefrua:
@MEsteves22 Done 👍

PR #3364, opened 2023-05-24
feat(Button): polymorphic button

MEsteves22:


packages/core/src/components/Button/Button.stories.tsx
```diff
@@ -194,3 +194,33 @@ export const Semantic = ({ onClick }) => {
...
+export const SetComponent = ({ onClick }) => {
+  return (
+    <HvBox sx={{ display: "flex", gap: 20, padding: 20 }}>
+      <HvButton onClick={onClick}>Button element</HvButton>
+      <HvButton onClick={onClick} component="a">
+        Link element
+      </HvButton>
+      <HvButton onClick={onClick} component="div">
+        Div element
+      </HvButton>
```
zettca, line 207: I believe that the most common use-cases will be to use the router's `Link`, or a custom button element. Can we add a sample of that?

We can add a dummy one, like so:
```tsx
const Link = ({to, children}) => <a href={to}>{children}</a>
```

packages/core/src/components/Button/Button.tsx
```diff
@@ -116,15 +119,18 @@ export const HvButton = forwardRef<HTMLButtonElement, HvButtonProps>(
...
-      <StyledButton
+      <StyledComponent
         id={id}
         className={clsx(className, classes?.root, buttonClasses.root)}
         ref={ref}
         onClick={onClick}
         disabled={disabled}
         onFocus={onFocusHandler}
         onBlur={onBlurHandler}
+        tabindex="0" // Allow keyboard focus for all elements
```
zettca, line 133: Do we really want this? If we do `component="div"` without an `onClick`, why should be `HvButton` be focusable? We'd be focusing an element that "does nothing".

Come to think of it, we probably shouldn't be promoting using `component="div"`, for accessibility reasons.

packages/core/src/components/Button/Button.stories.tsx
```diff
@@ -194,3 +194,33 @@ export const Semantic = ({ onClick }) => {
...
 };
+
+export const SetComponent = ({ onClick }) => {
+  return (
+    <HvBox sx={{ display: "flex", gap: 20, padding: 20 }}>
+      <HvButton onClick={onClick}>Button element</HvButton>
+      <HvButton onClick={onClick} component="a">
+        Link element
+      </HvButton>
+      <HvButton onClick={onClick} component="div">
```
zettca, line 205: A `div` with `onClick` - should we be promoting this?

packages/core/src/components/Button/Button.stories.tsx
```diff
@@ -194,3 +194,33 @@ export const Semantic = ({ onClick }) => {
 Semantic.args = {
   onClick: clickAction,
 };
+
+export const SetComponent = ({ onClick }) => {
+  return (
+    <HvBox sx={{ display: "flex", gap: 20, padding: 20 }}>
+      <HvButton onClick={onClick}>Button element</HvButton>
+      <HvButton onClick={onClick} component="a">
```
zettca, line 202: If we're using an `a`, we want an `href` here, not `onClick`

```suggestion
      <HvButton component="a" href="https://example.com">
```

packages/core/src/components/Button/Button.tsx
```diff
@@ -46,6 +46,8 @@ export interface HvButtonProps
   classes?: HvButtonClasses;
   /** Whether the Button is selected or not. */
   selected?: boolean;
+  /** The component to be used as the root node. Defaults to `button`. */
+  component?: React.ElementType<React.HTMLAttributes<HTMLElement>>;
```
zettca, line 50: Will this work for custom components (ie `Link`)? 🤔
Also, will we be able to  pass an `href` in `...others`? This `extends React.ButtonHTMLAttributes`.

For now, we could also extend from `Record<string, any>`, but ideally `HvButton` should inherit the `component`'s props.

packages/core/src/components/Button/Button.stories.tsx
MEsteves22: I will add this

packages/core/src/components/Button/Button.tsx
MEsteves22: No, it will not. I'll review this, thanks!

packages/core/src/components/Button/Button.tsx
MEsteves22: I added this because inside the button component we use the `onFocus` and `onBlur` handlers to add a focus styling around the button. Without this the styling was not being applied to the button if another component was used.

Maybe we should change this and only allow the styling to be applied if the `onClick` handler is provider?

MEsteves22: Yes, you are right. I will remove this and the div example. Thanks!

MEsteves22:
@zettca I updated my PR. I removed from Storybook the examples we shouldn't promote and added an example for the custom link like you asked. Regarding this example, if the user wants the same styling and/or use the features we have for the button, they should use `...others` as I have in my example. Let me know what you think.

I also updated `HvButtonProps` and the props should now update depending on the root node used. I used some types we already had for polymorphic components. I forgot we had them initially.

Just one note: I had to manually set the args table because Storybook is not able to correctly infer the properties when we use types (instead of interfaces) and import from aliases.

packages/core/src/components/BulkActions/BulkActions.tsx
```diff
@@ -54,7 +54,7 @@ export interface HvBulkActionsProps extends HvBaseProps {
   /**
    * Function called when the "select all pages" button is clicked toggled.
    */
-  onSelectAllPages?: HvButtonProps["onClick"];
+  onSelectAllPages?: HvButtonProps<"button">["onClick"];
```
zettca, line 57: We shouldn't need to specify `<"button">` every time. We can default it to `"button"`

```suggestion
  onSelectAllPages?: HvButtonProps["onClick"];
```

packages/core/src/components/Button/Button.tsx
```diff
@@ -116,21 +138,24 @@ export const HvButton = forwardRef<HTMLButtonElement, HvButtonProps>(
...
     return (
-      <StyledButton
+      <StyledComponent
         id={id}
-        className={clsx(className, classes?.root, buttonClasses.root)}
         ref={ref}
+        className={clsx(className, classes?.root, buttonClasses.root)}
         onClick={onClick}
         disabled={disabled}
+        aria-disabled={disabled}
```
zettca, line 150: Is this necessary? 🤔 
We already have `disabled`, and [they're semantically the same](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#description). The difference is that `disabled` does "more stuff", which we actually do want

packages/core/src/components/Button/Button.tsx
```diff
@@ -79,24 +95,30 @@ const mapVariant = (
...
-      disabled,
+      disabled = false,
       className,
       startIcon,
       endIcon,
       icon = false,
       size,
       radius = "base",
       overrideIconColors = true,
+      as: Component = "button",
```
zettca, line 119: I do like `as`; it's shorter and clear enough, but we're already using the `component` prop in other components (same as MUI), including the previous `HvButton` (currently an undocumented breaking change)

I think we should keep the `component` prop name. What do you think?

```suggestion
      component: Component = "button",
```

MEsteves22: Okay, I will change our polymorphic types for use component `instead`

packages/core/src/components/Button/Button.tsx
MEsteves22: Okay, I will remove this

packages/core/src/components/Button/Button.tsx
MEsteves22: The `as` property comes from the `PolymorphicComponent` type. This type is used by our `HvBox` component. If I change `as` to `component` this will lead to a breaking change for the `HvBox` component.

MEsteves22: But I don't believe the `HvBox` is currently being used to be honest but this will be a breaking change nonetheless.

MEsteves22: Maybe we can discuss this with the team and see what they think (if we should update `as` to `component`)? I will updated the migration guide afterwards.

zettca: Sure 👍 

I wasn't aware of other components using it.

Well, `HvBox` doesn't have a documentation page, and it's barely used in out components and samples.

I think that we definitely should be leveraging `PolymorphicComponentRef` in other polymorphic components, so that we get the correct typings. So, to me this isn't a `HvBox` vs `HvButton`, but a `HvBox` vs many components 😅.

I think it makes more sense to just drop the `PolymorphicComponentRef` "feature" in `HvBox` and add it to the components (ie. Typography, Table, Avatar, etc.) that are already using a `component` prop and UI-Kit users commonly use. It wouldn't be a breaking change or change in API, just a drop in type safety. Alternatively, we could duplicate the type

Some examples of the issue we have:
```tsx
<HvTypography src="typography is not an img">
<HvAvatar component="a" href="asda" /> // does not accept `href`
```

packages/core/src/components/Button/Button.tsx
MEsteves22: I added this property because not all HTML elements have this `disabled` property. For example, if my root node is a `a` element, I need the `aria-disabled` property to tell the accessibility tree that the element is disabled.

MEsteves22:
@zettca @plagoa @HQFOX  As discussed, I changed the `a` property to `component` for the `PolymorphicComponentRef` type. Because of this, `HvBox` now receives the `component` property instead of `a`. 

@francisco-guilherme This is a breaking change since `HvBox` and `PolymorphicComponentRef` are exported by the core package. However, we decided to make this change in order for all polymorphic components to receive the same `component` property. Otherwise, we would have some components using `component` (table, typography, avatar, etc) and others using `as`. Do you agree with this change? `HvBox` and `PolymorphicComponentRef` are most likely not used by our users at the moment since they are undocumented so we don't think it's going to be disruptive.

Regarding the other changes, I added the `tabIndex` and `aria-disabled` properties when the button is disabled because when the root component is not a button (a link for example) the button was still focusable and not disabled in the accessibility tree when the `disabled` property was used.

zettca (approved):
Looking great 👌 🚀

I doubt users will be using `<HvButton disabled component="a" href=".." />`, but doesn't hurt to add 👍

francisco-guilherme (approved):

Issue #3368, opened 2023-05-24
`HvStepNavigation` width passes the size of `HvWizard`


stephaniegaspar:
### Which version of NEXT UI Kit are you opening this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

On "xl" breakpoint, the `HvStepNavigation` is rendered with fullwidth and passes the size of `HvWizard`.

![image](https://github.com/lumada-design/hv-uikit-react/assets/108751625/511eb080-8fbf-4d04-9b72-957e474a4cdc)

### Steps to reproduce 🕹

Enter sandbox link: 

And/or the steps to reproduce:

1. Go to [wizard page documentation](https://lumada-design.github.io/uikit/master/?path=/docs/lab-wizard--main)
2. Click on "Show Wizard" button to render a demo sample

### Your Environment 🌎

Device: [e.g. Laptop, Mobile, ....]
Browser [e.g. stock browser, safari, chrome]
Other details [e.g. screen resolution 1920x1080, ...]

stephaniegaspar:
Hi @zettca 👋 
The main issue for now the missing logic to give `HvStepNavigation` a max width when the `window` is larger than the pixels provided to "lg" breakpoint key on the default ui-kit theme. Can you please check the PR related to this issue that I opened recently? Thank you!

PR #3369, opened 2023-05-24
fix(HvWizard, HvStepNavigation): step navigation with maxWidth on xl breakpoint. #3368

stephaniegaspar:
* `HvStepNavigation`: Definition of missing maxWidth on `xl` breakpoint
* `HvWizard`: Extend definition of props `maxWidth` and `fullWidth` of `HvDialog`. Value `lg` is now default for prop `maxWidth`.


zettca (approved):
LGTM 👌

PR #3372, opened 2023-05-25
docs: add Server Side sample to the Table docs

plagoa:


MEsteves22 (approved):

packages/core/src/components/Table/stories/TableSamples/LoadingContainer.tsx
```diff
@@ -0,0 +1,63 @@
...
+    width: "100%",
+    height: "100%",
+  }),
+  overlay: css({
+    position: "absolute",
+    transition: "background-Color .2s ease",
+    zIndex: -1,
+  }),
+  blur: css({
+    backgroundColor: hexToRgbA("#EEEEEE"),
```
zettca, line 19: Leftovers? This will always be overriden

PR #3374, opened 2023-05-25
fix(VerticalNavigation): no padding on menu items

plagoa:
`HvListItem` expects `undefined` if no adornment is present and having `<div>{item.icon}</div>` incorrectly messes that up.


zettca (approved):
This wouldn't be necessary if we weren't hyjacking the user-provided `startAdornment` element 😞

PR #3376, opened 2023-05-25
feat(Viz): viz provider and line chart added

MEsteves22:
- `HvVizProvider` created for visualisations themes.
- `HvLineChart` created: line, stacked, area, range slider, and multiple series.

At the moment the data can either be [columns](https://uwdata.github.io/arquero/api/#table), [rows](https://uwdata.github.io/arquero/api/#from), or an arquero table. In the future, we can add new data formats such as [CSV](https://uwdata.github.io/arquero/api/#fromCSV), [JSON](https://uwdata.github.io/arquero/api/#fromJSON), [Apache Arrow](https://uwdata.github.io/arquero/api/#fromArrow), and [so on](https://uwdata.github.io/arquero/api/#table-constructors). 

This is a first iteration for charts. We will most likely need to refactor the code along the way if other charts need to share the same API. For now I decided to keep the code as it is since I believe this is something that will make more sense in the future when working on the other charts.


packages/viz/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+    if (Array.isArray(measures.fields)) {
+      for (let i = 0; i < measures.fields.length; i++) {
+        const field = measures.fields[i];
+        aggregations[field] = getAggregation(
+          Array.isArray(agFunction) ? agFunction[i] ?? "sum" : agFunction,
+          field
+        );
+      }
+    }
```
nantunes, line 200: Had some trouble reading this... It is probably easier to first ensure you always working with arrays.

```suggestion
    const measuresFields = Array.isArray(measures.fields)
      ? measures.fields
      : [measures.fields];
    const agFunction = Array.isArray(measures.aggregation)
      ? measures.aggregation
      : [measures.aggregation];

    const aggregations = measuresFields.reduce(
      (acc, field, i) => ({
        ...acc,
        [field]: getAggregation(agFunction[i] ?? "sum", field),
      }),
      {}
    );
```

packages/viz/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,404 @@
...
+        horizontalRangeSlider?.show === true
+          ? {
+              type: "slider",
+              orient: "horizontal",
+            }
+          : undefined,
+    };
+  }, [horizontalRangeSlider?.show]);
+
+  return (
```
nantunes, line 386: How does the component react to changes? In particular with changes in the data?

packages/viz/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,404 @@
...
+    valueFormatter?: (value?: string | number) => string;
+  };
+  /** Legend options. */
+  legend?: {
+    /**
+     * Whether to show the legend or not.
+     *
+     * If they are multiple series, the legend will appear by default. Otherwise, the legend will not be shown.
+     */
+    showLegend?: boolean;
```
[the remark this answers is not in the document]
MEsteves22: Yes, I will change this

packages/viz/src/components/LineChart/LineChart.tsx
MEsteves22: Okay, I will update this code to improve readability

packages/viz/src/components/LineChart/LineChart.tsx
MEsteves22: I tested this with the "With Interval Updates" story where the data is updating every second and the range slider was always resetting to the initial values. Because of this it was very difficult to use the range slider.

This was fixed in a new commit (I also added the changes you mentioned) and the range slider is no longer resetting the values every time the data updates. This is the current behaviour:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/99b152bf-c1ce-48fb-a636-0ce793ade760

packages/viz/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,432 @@
...
+      tableData = table(data);
+    }
+
+    const measuresFields = Array.isArray(measures.fields)
+      ? measures.fields
+      : [measures.fields];
+
+    const agFunction = Array.isArray(measures.aggregation)
+      ? measures.aggregation
+      : [measures.aggregation];
```
nantunes, line 189: Sorry, my previous suggestion removed the ability to specify a single aggregation that applies to all measures.

```suggestion
      : measuresFields.map(() => measures.aggregation as HvChartAggregation);
```

MEsteves22:
I just noticed that when I started using `setOptions` the tooltip styling is no longer updating when the color mode is changed 😅 I will look into this.

MEsteves22:
The tooltip problem was fixed in my last commit. I had to stop updating the tooltip options with `setOptions`.

MEsteves22:
Regarding the tooltip, I plan to remove the usage of the `@emotion/css` library when this [PR](https://github.com/lumada-design/hv-uikit-react/pull/3379) is merged since we do not want to promote its usage internally. However, I had to use it for now because it's not possible to use `ClassNames` or styled components.

MEsteves22:
@lumada-design/imperial This PR should be squash

packages/viz/src/providers/Provider.tsx
```diff
@@ -0,0 +1,151 @@
...
+ */
+export const HvVizProvider = ({ children }: HvVizProviderProps) => {
+  const { activeTheme, colorModes, selectedMode, selectedTheme } = useTheme();
+
+  const [theme, setTheme] = useState<string | undefined>();
+
+  const registerThemes = (
+    themeName: string,
+    modes: string[],
+    themeStructure?: HvTheme
```
zettca, line 49: This looks like a pure function - it could be extracted from the component

packages/viz/src/providers/Provider.tsx
```diff
@@ -0,0 +1,151 @@
...
+            width: 2,
+          },
+        },
+      });
+    });
+  };
+
+  useEffect(() => {
+    registerThemes(selectedTheme, colorModes, activeTheme);
+  }, [selectedTheme]);
```
zettca, line 142: dependency array sync 😬

packages/viz/src/providers/Provider.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  useEffect(() => {
+    setTheme(`${selectedTheme}-${selectedMode}`);
+  }, [selectedMode, selectedTheme]);
+
+  return (
+    <HvVizContext.Provider value={{ theme }}>{children}</HvVizContext.Provider>
```
zettca, line 149: I would memoize `value`, for future-proofing performance reasons

zettca (approved):
Awesome work 👍 

Some minor notes (that can be handled later)

packages/viz/src/providers/Provider.tsx
MEsteves22: Yes, good idea I'll memorize it

packages/viz/src/providers/Provider.tsx
MEsteves22: I will leave it there for now, but yes this is something we can extract later on 🙏

Issue #3375, opened 2023-05-25
`HvTable` with scrollbar when has "actions" column


stephaniegaspar:
### Which version of NEXT UI Kit are you opening this bug for?

None

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

When `HvTable` has an action colum with `HvDropDownMenu` on each cell, shows an horizontal scrollbar. 
The problem seems to be the `HvDropDownMenu` width that passes the size of the cell.

### Expected behavior 🤔

The table should not have an horizontal scrollbar when every column fits the content.

### Steps to reproduce 🕹

Steps to reproduce:

1. Open the [`HvTable` documentation page](https://lumada-design.github.io/uikit/master/?path=/docs/guides-table--responsive-table#simple-table)
2. Shows the scrollbar with the "actions" column 
![image](https://github.com/lumada-design/hv-uikit-react/assets/108751625/9cbd774e-9ec1-4306-823b-9c31b9b0f79f)

### Context 🔦

Related work for this issue: [MLAAS-2247](https://hv-eng.atlassian.net/browse/MLAAS-2247)

### Your Environment 🌎

Device: [e.g. Laptop, Mobile, ....]
Browser [e.g. stock browser, safari, chrome]
Other details [e.g. screen resolution 1920x1080, ...]

[MLAAS-2247]: https://hv-eng.atlassian.net/browse/MLAAS-2247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

MEsteves22:
Hello 👋

Thank you for opening this issue. This was fixed on version [5.10.0](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.10.0).

plagoa:
This has been fixed in [@hitachivantara/uikit-react-core@5.11.0](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.11.0)

PR #3383, opened 2023-05-29
fix(Table): no scrollbar added when column has actions

MEsteves22:
- Fixes issue: https://github.com/lumada-design/hv-uikit-react/issues/3375


zettca (approved):
👌

stephaniegaspar (approved):

PR #3389, opened 2023-05-31
fix(VerticalNavigation): open slider where item is selected

HQFOX:


packages/core/src/components/VerticalNavigation/Navigation/Navigation.tsx
```diff
@@ -7,7 +7,7 @@ import {
   ComponentProps,
 } from "react";
 import { clsx } from "clsx";
-import uniqueId from "lodash/uniqueId";
+import { uniqueId } from "lodash";
```
zettca, line 10: Importing `lodash` like this will being in the entire bundle - we don't want that

```suggestion
import uniqueId from "lodash/uniqueId";
```

zettca (changes requested):

MEsteves22 (approved):

PR #3392, opened 2023-06-01
chore: add `eslint` overrides to `app/`

zettca:


nantunes (approved):
I know this is just the sample app, so it is not critical... But as general rules I would suggest:
1. Use explicit "no-alert" inline ignores, near the usage of alert, possibly with an explanation;
2. Only allow warn and error (no-console: ["error", { allow: ["warn", "error"] }]), as those are usually used intentionally. Require explicit inline ignore near the usage of console.log (most of the time unintended left-overs).

francisco-guilherme (approved):

PR #3403, opened 2023-06-05
chore: add app publication in github actions

MEsteves22:


.github/workflows/tests-pa11y.yml
```diff
@@ -20,8 +20,10 @@ jobs:
         uses: ./.github/actions/install-dependencies
 
       - name: Build Documentation
-        uses: ./.github/actions/build-documentation
+        uses: ./.github/actions/build-artifacts
         if: github.event_name == 'workflow_dispatch'
+        with:
+          build-app: false
```
zettca, line 26: Nice 👌

.github/actions/build-artifacts/action.yml
```diff
@@ -0,0 +1,40 @@
...
+  using: "composite"
+  steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install Dependencies
+      uses: ./.github/actions/install-dependencies
+
+    - name: Build Documentation
+      run: npx cross-env NODE_OPTIONS=--max_old_space_size=8192 build-storybook --docs -o dist --quiet
```
zettca, line 21: Shouldn't this script live in our `package.json` scripts? We already have `build:doc`
`max_old_space_size` seems to be no longer necessary (was an old workaround)

```suggestion
      run: npm run build:doc
```

.github/actions/deploy-artifacts/action.yml
```diff
@@ -39,23 +49,35 @@ runs:
...
-    - name: Push Documentation
+    - name: Push Documentation and App
       run: |
         cd ${{ github.workspace }}/gh-docs
         git config --global user.name 'github-actions[bot]'
         git config --global user.email 'github-actions[bot]@users.noreply.github.com'
         git add -A ${DOCUMENTATION_PUBLISH_FOLDER}
         git commit -m "${DOCUMENTATION_PUBLISH_MESSAGE}"
+        git add -A ${APP_PUBLISH_FOLDER}
+        git commit -m "${APP_PUBLISH_MESSAGE}"
```
zettca, line 76: Separate commits & messages seem like unnecessary extra work - is there any advantage I'm not seeing? 🤔
A single commit/push will simplify/reduce the git tree of the repo. It probably won't make much of a difference, but it might help reduce push/sync conflicts.

.github/actions/deploy-artifacts/action.yml
```diff
@@ -26,10 +26,20 @@ runs:
         ref: ${{ env.DEPLOYMENT_REF }}
         desc: ${{ env.DOCUMENTATION_PUBLISH_MESSAGE }}
 
-    - name: Checkout Documentation Repository
+    - name: Start App Deployment
+      uses: bobheadxi/deployments@v1
+      id: deployment
```
zettca, line 31: This probably needs to have another name

zettca (commented):
Looking great overall 👍

Just a few notes/comments:

.github/actions/build-artifacts/action.yml
MEsteves22: I believe we still need this `max_old_space_size` property. I tried using `npm run build:doc` and got the following [error](https://github.com/lumada-design/hv-uikit-react/actions/runs/5177698255/jobs/9328459698):

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

.github/actions/deploy-artifacts/action.yml
MEsteves22: No special reason behind doing separate commits. It was just a personal preference. 

In the end I had to change this code because [it's not possible to use deployment more than once in the same scope](https://github.com/lumada-design/hv-uikit-react/actions/runs/5176135648/jobs/9324752139). However, if I use it only once to deploy both the app and documentation, it's only possible to set one environment url when finishing the development status. Thus, I now use the matrix strategy from GitHub to be able to run this `deploy-artifacts` for both the app and documentation and have environment urls for both.

francisco-guilherme (approved):

PR #3413, opened 2023-06-06
feat(VerticalNavigation): move chevrons to the right

plagoa:
New version:

image


nantunes (commented):
I wouldn't call this a fix... It was the DS spec that changed. It is a feature. Potentially a breaking change: it doesn't change functionality, but products that have visual checks in their pipelines will break. But lets keep it simple.

Also, I know it isn't in the scope of this, but those left padding of sub-items continue to be wrong and inconsistent.

plagoa:
@nantunes I'll change this to feature, I actually asked myself the same thing. I'll take a look at the paddings too.

PR #3419, opened 2023-06-09
chore: fix linting

francisco-guilherme:


packages/core/src/components/TimeAgo/TimeAgo.tsx
```diff
@@ -64,10 +64,9 @@ export const HvTimeAgo = ({
     showSeconds,
   });
 
-  // eslint-disable-next-line react/jsx-no-useless-fragment
-  if (justText && timestamp) return <>{timeAgo}</>;
-
-  return (
+  return justText && timestamp ? (
+    <>timestamp</>
```
MEsteves22, line 68: I believe it's `<>{timeAgo}</>` and not `<>timestamp</>`.

packages/core/package.json
```diff
@@ -42,17 +42,18 @@
...
   "dependencies": {
     "@emotion/cache": "^11.10.5",
+    "@emotion/css": "^11.10.6",
+    "@emotion/serialize": "^1.1.2",
     "@hitachivantara/uikit-react-icons": "^5.1.11",
     "@hitachivantara/uikit-react-shared": "^5.0.15",
     "@hitachivantara/uikit-styles": "^5.8.1",
+    "@mui/base": "^5.0.0-beta.3",
     "@popperjs/core": "^2.11.6",
-    "@types/react-table": "^7.7.14",
```
MEsteves22, line 49: Regarding the react table types, I believe we need to install them as a dependency and not a dev dependency because we need these types to pass transitively to our users since we extend them.  
I believe this is something @nantunes mentioned when we were having problems with the react table types a few weeks ago.

packages/core/src/components/ScrollTo/withTooltip.tsx
```diff
@@ -8,7 +8,7 @@ import { withTooltip as withTooltipUtil } from "@core/hocs";
 const hideTooltip = (event: React.MouseEvent<HTMLDivElement>) => {
   const isOverFlow =
     (event.target as HTMLDivElement).children.length > 1
-      ? Array.of(...(event.target as HTMLDivElement).children).some(
+      ? Array.from((event.target as HTMLDivElement).children).some(
```
zettca, line 11: Unrelated, but if we want the `HTMLDivElement`, we should be using `event.currentTarget`
```suggestion
      ? Array.from((event.currentTarget).children).some(
```

packages/core/src/components/QueryBuilder/queryToMongo.tsx
```diff
@@ -16,7 +16,6 @@ function isAnd(rule) {
 
 const groupHasNoSubGroupsOrJustOneOR = (group) => {
   let foundOr = false;
-  // eslint-disable-next-line no-plusplus
```
zettca, line 19: Could likely be refactored into a simpler & easier to read iteration, instead of a loop. Worst case scenario, a `forEach`

packages/core/src/components/Table/stories/TableHooks/TableHooks.stories.tsx
```diff
@@ -151,12 +157,6 @@ const UseHvPagination = () => {
     getHvPaginationProps,
   } = useHvData<NewEntry, string>({ columns, data }, useHvPagination);
 
-  const EmptyRow = () => (
-    <HvTableRow>
-      <HvTableCell colSpan={100} />
-    </HvTableRow>
-  );
```
zettca, line 158: Declaring components inside components is definitely an anti-pattern, but moving this outside the sample code makes the sample harder to read...
We should (maybe not now) work on improving this

packages/core/src/components/TimeAgo/TimeAgo.tsx
zettca: Why not just `timestamp`, like it used to be? `string` is a valid `ReactNode`. Why convert it to a `ReactElement`? It's _kinda_ going against the `justText` prop too...

packages/core/src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -53,7 +53,7 @@ describe("HvDatepicker", () => {
 
       calendarButtons = await findAllByRole("button");
       firstDayOfTheMonth = await findAllByText("1");
-      expect(calendarButtons.length).toBe(42);
+      expect(calendarButtons.length).toBe(43);
```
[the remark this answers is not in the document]
zettca: Probably coming from this: https://github.com/lumada-design/hv-uikit-react/pull/3419#discussion_r1229151046

packages/core/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -213,7 +213,11 @@ export const Form: StoryObj<HvDialogProps> = {
               onSubmit={(event) => {
                 event.preventDefault();
                 const formData = new FormData(event.currentTarget);
-                setPostData(Object.fromEntries(formData.entries()));
+                const data = {};
+                formData.forEach((value, key) => {
+                  data[key] = value;
+                });
```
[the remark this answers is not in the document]
francisco-guilherme: You're right, there were some ts issues, but were solved by adding the missing configs:

https://github.com/lumada-design/hv-uikit-react/blob/7fb347c5c780c0e3daacbca1316e4985d7e5d444/tsconfig.json#L5

packages/core/src/components/FilterGroup/FilterGroup.test.tsx
```diff
@@ -168,13 +168,15 @@ describe("FilterGroup", () => {
 
     expect(dropdownElement).toHaveAttribute("aria-expanded", "true");
 
-    const applyButton = getByRole("button", { name: /apply/i });
+    const applyButton = getAllByRole("button", { name: /apply/i })[1];
```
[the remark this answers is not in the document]
francisco-guilherme: By adding the needed element role has these side effects here https://github.com/lumada-design/hv-uikit-react/pull/3419#discussion_r1229151046

packages/core/src/components/TimeAgo/TimeAgo.tsx
francisco-guilherme: I needed to cast it because of ts. let me know what you think.

packages/core/src/components/BaseDropdown/BaseDropdown.tsx
```diff
@@ -442,7 +442,7 @@ export const HvBaseDropdown = ({
         {...attributes.popper}
       >
         <ClickAwayListener onClickAway={handleOutside}>
-          <div onKeyDown={handleContainerKeyDown}>
+          <div role="button" tabIndex={0} onKeyDown={handleContainerKeyDown}>
```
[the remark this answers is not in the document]
francisco-guilherme: as agreed i disabled the linting rule here

packages/core/src/components/TimeAgo/TimeAgo.tsx
zettca: What's the TS issue? I couldn't reproduce it on my side... 🤔 
Do we need a `JSX.Element`/`EmotionJSX.Element` to always be returned?

If so, then the fragment is not so "useless" after all, at it might make sense to keep it...

packages/core/src/components/Dialog/Dialog.stories.tsx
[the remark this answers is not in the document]
zettca: Strange, I never got those 🤔

zettca (approved):
LGTM

packages/core/package.json
```diff
@@ -64,11 +66,18 @@
     "react-window": "^1.8.8"
   },
   "devDependencies": {
+    "@playwright/test": "^1.34.3",
+    "@storybook/react": "^6.5.16",
+    "@testing-library/react": "^14.0.0",
+    "@testing-library/user-event": "^14.4.3",
```
zettca, line 72: The tests are failing - `HvCarousel`'s images are role `presentation` instead of `img`.
Might be caused by this bump 🤔

packages/core/src/components/QueryBuilder/queryToMongo.tsx
```diff
@@ -16,17 +16,16 @@ function isAnd(rule) {
...
-  for (let i = 0; i !== group.rules.length; ++i) {
-    const rule = group.rules[i];
+
+  group.rules.forEach((rule) => {
```
zettca, line 20: This is not quite equivalent, because of the `return`. We can leave this as it was - this function is scary

MEsteves22: Yup, we need to leave this as it was. Applitools broke because of this since the function is not behaving as it was before.

PR #3422, opened 2023-06-12
fix(Header): fix incorrect height for the items on the Header

plagoa:


packages/core/src/components/Header/Navigation/MenuItem/MenuItem.styles.tsx
```diff
@@ -42,16 +42,16 @@ export const MenuItemLi = styled(
   ...(!$selected &&
     !$isMenu && {
```
MEsteves22, line 43: Since the height is now `100%`, I think we should add something line `borderBottom: ${theme.header.selectedItemBorderBottomThickness} solid ${theme.header.backgroundColor}` here to have this white bottom border in DS5 when hovering an item. 

Screenshot 2023-06-13 at 09 28 07

plagoa: Yeah, I think that's a good idea. I'll update the PR.

packages/core/src/components/Header/Navigation/MenuItem/MenuItem.styles.tsx
```diff
@@ -42,16 +42,20 @@ export const MenuItemLi = styled(
...
     }),
   ...(!$selected &&
     $isMenu && {
       marginTop: theme.header.secondLevelSelectedItemBorderTopThickness,
-      marginBottom: theme.header.secondLevelSelectedItemBorderBottomThickness,
+      paddingBottom: theme.header.secondLevelSelectedItemBorderBottomThickness,
       paddingTop: theme.header.secondLevelSelectedItemBorderBottomThickness,
-      height: `calc(100% - ${theme.header.secondLevelSelectedItemBorderTopThickness} - ${theme.header.secondLevelSelectedItemBorderBottomThickness})`,
+      height: "100%",
+      borderTop: `${theme.header.selectedItemBorderTopThickness} solid ${theme.header.secondLevelBackgroundColor}`,
```
MEsteves22, line 57: Shouldn't we use the second level selected item borders?

plagoa: Yes, thanks. PR updated.

MEsteves22 (approved):

PR #3431, opened 2023-06-15
chore: allow manually editing the code in the theme creator

plagoa:


app/src/generator/Sidebar/Sidebar.tsx
```diff
@@ -36,29 +34,28 @@ const Zindices = lazy(() => import("generator/Zindices"));
...
-  const { selectedTheme, selectedMode, colorModes, themes, changeTheme } =
-    useTheme();
+  const { selectedTheme, selectedMode, colorModes, changeTheme } = useTheme();
 
-  const { updateCustomTheme, open } = useContext(GeneratorContext);
+  const { customTheme, updateCustomTheme, open, themeChanges } =
+    useContext(GeneratorContext);
 
-  const [themeName, setThemeName] = useState("customTheme");
+  const [themeName, setThemeName] = useState(customTheme.name);
```
zettca, line 42: If `setThemeName` is only used to update `themeName` when `customTheme.name` changes, then isn't this just the `customTheme.name` itself?

```suggestion
  const themeName = customTheme.name;
```

It works either way, I'm just commenting [because `useEffect`s are evil](https://react.dev/learn/you-might-not-need-an-effect) 😈 (extra re-renders, dependency array, etc.)

zettca (approved):

packages/core/src/utils/theme.ts
```diff
@@ -127,6 +127,7 @@ export const createTheme = (
   }
 
   // Created theme
+  customizedTheme.base = base;
```
MEsteves22, line 130: We already have this (`customizedTheme.base = base`) in the `createTheme` on line 103. We can remove this one.

MEsteves22:
I noticed the dropdown to select the base theme was removed. I think it would be preferable to keep it because some users may not know which base themes we have.

packages/core/src/utils/theme.ts
plagoa: Right, remains from previous work :)

plagoa:
> I noticed the dropdown to select the base theme was removed. I think it would be preferable to keep it because some users may not know which base themes we have.

Yeah, that makes sense. I was trying to avoid having multiple ways of doing the same thing, but that's a good point.

MEsteves22 (approved):

PR #3433, opened 2023-06-16
fix(ActionBar): height & alignment in ds5

zettca:
- Fix `HvActionBar` DS5 styles. Border is `1px` `atmo3`
- Port styling to `createClasses`
- Improve tests

image


packages/core/src/components/ActionBar/ActionBar.tsx
```diff
@@ -1,23 +1,23 @@
...
+import { ExtractNames } from "@core/utils";
+import { staticClasses, useClasses } from "./ActionBar.styles";
+
+export { staticClasses as actionBarClasses };
+
+export type HvActionBarClasses = ExtractNames<typeof useClasses>;
 
 export interface HvActionBarProps extends HvBaseProps {
   /** A Jss Object used to override or extend the styles applied to the component. */
   classes?: HvActionBarClasses;
```
MEsteves22, line 11: Nothing major because we only have one class, but to maintain backward compatibility we should use `Partial<HvActionBarClasses>`.

MEsteves22 (approved):

packages/core/src/components/ActionBar/ActionBar.tsx
zettca: Nice catch 👀

PR #3434, opened 2023-06-16
chore: activate pa11y and publish report

plagoa:


.github/workflows/tests-pa11y.yml
```diff
@@ -35,4 +35,31 @@ jobs:
...
+    name: Publish pa11y Report
+    needs: run-tests
+    if: always()
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Deploy Artifacts
```
zettca, line 56: I don't think that we need a "whole job" for this, we can `if: always()` a step

```suggestion
      - name: Deploy Artifacts
        if: always()
```

.github/workflows/tests-pa11y.yml
```diff
@@ -35,4 +35,31 @@ jobs:
...
           "npx http-server dist --port 6006 --silent" \
-          "npx wait-on tcp:6006 && npm run test:pa11y"
\ No newline at end of file
+          "npx wait-on tcp:6006 && npm run test:pa11y" || true
+
+      - name: Archive pa11y report
+        uses: actions/upload-artifact@v3
+        with:
+          name: pa11y-report
+          path: ${{ github.workspace }}/pa11y-ci-report
```
zettca, line 44: We're storing quite a few things now. We should probably define a shorter [retention-period](https://github.com/actions/upload-artifact#retention-period) here.
We can look into this later

zettca:
Report is looking great BTW 👌 🚀

zettca (approved):
:shipit:

MEsteves22 (approved):

francisco-guilherme (approved):

PR #3441, opened 2023-06-20
fix(Storybook): add Open Sans font to the preview iframe

plagoa:


.storybook/preview-head.html
```diff
@@ -0,0 +1,62 @@
+<!-- .storybook/preview-head.html -->
+<style>
+  @font-face {
```
zettca, line 3: Couldn't we import our `font-faces.css` file instead? 🤔

zettca (approved):

PR #3446, opened 2023-06-21
chore: app type references added and type errors fixed

MEsteves22:
- As discussed I added the type references for the app to fix errors like this:

Screenshot 2023-06-21 at 15 30 59

- I also fixed type errors

### DIFF
--- a/app/src/components/components/BulkActions/BulkActions.tsx
+++ b/app/src/components/components/BulkActions/BulkActions.tsx
@@ -2,9 +2,9 @@ import styled from "@emotion/styled";
 import {
   HvCheckBox,
   HvActionGeneric,
-  HvListValue,
   HvBulkActions,
   HvPagination,
+  HvBulkActionsProps,
 } from "@hitachivantara/uikit-react-core";
 import { Add, Delete, Preview, Lock } from "@hitachivantara/uikit-react-icons";
 import { theme } from "@hitachivantara/uikit-styles";
@@ -113,10 +113,10 @@ export const BulkActions = () => {
     setData(newData);
   };
 
-  const handleAction = (
-    event: React.SyntheticEvent,
-    id: string,
-    action: HvActionGeneric | HvListValue
+  const handleAction: HvBulkActionsProps["actionsCallback"] = (
+    event,
+    id,
+    action
   ) => {
     const selected = data.filter((el) => el.checked);
 
--- a/app/src/components/components/BulkActions/BulkActions.tsx
+++ b/app/src/components/components/BulkActions/BulkActions.tsx
@@ -103,13 +103,21 @@ export const BulkActions = () => {
     setData(newData);
   };
 
-  const handleChange = (_, i: number, checked: boolean) => {
+  const handleChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+    i: number,
+    checked: boolean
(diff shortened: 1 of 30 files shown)


app/src/components/components/BulkActions/BulkActions.tsx
```diff
@@ -103,13 +103,21 @@ export const BulkActions = () => {
...
     newData[i + pageSize * page].checked = checked;
     setData(newData);
   };
 
-  const handleAction = (_, __, action: HvActionGeneric | HvListValue) => {
+  const handleAction = (
+    event: React.SyntheticEvent,
+    id: string,
+    action: HvActionGeneric | HvListValue
+  ) => {
```
zettca, line 120: We shouldn't be "redefining" the types. It's extra work, prone to errors, less future-proof, etc.

```suggestion
  const handleAction: HvBulkActions["actionsCallback"] = (event, id, action) => {
```

app/src/components/components/ButtonConfigurator/ButtonConfigurator.tsx
```diff
@@ -35,7 +36,7 @@ export const ButtonConfigurator = () => {
           label: name,
         }))}
         onChange={(item) => {
-          setVariant(item.value as HvButtonVariant);
+          setVariant((item as HvListValue).value as HvButtonVariant);
```
zettca, line 39: We definitely need a proper Select element...

app/src/components/components/Calendar/Calendar.tsx
```diff
@@ -16,7 +16,14 @@ export const Calendar = () => {
...
 
-  const onChangeHandler = (event, value) => {
+  const onChangeHandler = (
+    event:
+      | React.ChangeEvent<
+          HTMLTextAreaElement | HTMLInputElement | HTMLButtonElement
+        >
+      | undefined,
+    value: Date | DateRangeProp
+  ) => {
```
zettca, line 26: same as above
```suggestion
  const onChangeHandler: HvCalendarProps["onChange"] = (event, value) => {
```

app/src/components/components/Calendar/Calendar.tsx
```diff
@@ -28,21 +35,23 @@ export const Calendar = () => {
...
+        endDate: value as Date,
       });
     }
   };
 
   const visibleDateChangeHandler = (
-    event: any,
+    event:
+      | React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>
+      | undefined,
```
zettca, line 52: same as above, and for other instances

zettca (approved):
Awesome 💪

Now we just need the same for our whole codebase 😅 

> TypeScript isn't making your life worse. It's just showing you how bad your life already is

app/src/components/components/ButtonConfigurator/ButtonConfigurator.tsx
MEsteves22: Agreed 🙏

app/src/components/components/BulkActions/BulkActions.tsx
MEsteves22: You're right, I will change this 🙏

plagoa (approved):

PR #3447, opened 2023-06-21
fix(button): grow according to content in icon mode

zettca:
- fixes usage with larger `HvAvatar` (or any other content eg. Icons)
- improve `HvAvatar` & `HvButton` samples
- improve `HvButton` tests

image


plagoa:
@zettca can we update the `Buttons` story to have one of the avatars be square? That way we can validate that everything's ok in the variant also.

zettca:
> actually, what's the expected behavior for buttons with long text? adapt to the text or cut it?

I think that typically `HvButton` just grows to the size they need, so I think that this case is not that common. But I half-expected the text to behave a bit better (eg. wrap to 2 lines) 🤷
At the end of the day, it's the users' responsibility to make sure the button is large enough for the text it has.

plagoa (approved):

plagoa:
it seems like with these changes the `Carousel` has some problem. It seems that that theme prop should be set to `16px` now.

image

MEsteves22 (approved):

PR #3451, opened 2023-06-22
refactor(DotPagination): port styled components to createClasses

HQFOX:
Sometimes the styles of the component were not being correctly overridden like the image shows, which would cause a applitools error on the Customized Dot Pagination story, updated to createClasses and the styles are being correctly overridden.

Screenshot 2023-06-22 164726


packages/core/src/components/DotPagination/DotPagination.tsx
```diff
@@ -47,7 +45,7 @@ export interface HvDotPaginationProps
   /**
    * A Jss Object used to override or extend the styles applied.
    */
-  classes?: HvDotPaginationClasses;
+  classes?: Partial<HvDotPaginationClasses>;
```
zettca, line 48: We no longer need `Partial` (nor above). It's included in `ExtractNames`.

Here it's triplicated!:

```tsx
classes?: Partial<Partial<Partial<ReturnType<typeof useClasses>["classes"]>>>;
```

packages/core/src/components/DotPagination/DotPagination.tsx
```diff
@@ -85,22 +83,24 @@ export const HvDotPagination = ({
...
 
+  const { classes, cx } = useClasses(classesProp);
+
   const icons = getSelectorIcons(unselectedIcon, selectedIcon, classes);
 
   return (
-    <StyledRadioGroup
-      className={clsx(className, classes?.root, dotPaginationClasses.root)}
+    <HvRadioGroup
+      className={cx(className, classes.root)}
```
zettca, line 92: `className` should come after, as it's a user-defined and should override our styles
```suggestion
      className={cx(classes.root, className)}
```

zettca (commented):
Looking good 👌 

some small notes:

packages/core/src/components/DotPagination/DotPagination.tsx
```diff
@@ -1,15 +1,13 @@
...
-  StyledRadioGroup,
-} from "./DotPagination.styles";
+import { HvRadio, HvRadioGroup, HvRadioGroupProps } from "@core/components";
+import { ExtractNames } from "@core/utils";
+
+import { staticClasses, useClasses } from "./DotPagination.styles";
+
+export { staticClasses as dotPaginationClasses };
+
+export type HvDotPaginationClasses = Partial<ExtractNames<typeof useClasses>>;
```
MEsteves22, line 10: This `Partial` is no longer needed too since `ExtractNames` already has this.

MEsteves22 (approved):

zettca (approved):
🚀

PR #3453, opened 2023-06-22
fix(DatePicker): have DatePicker use the HvFormStatus type for its status prop

plagoa:
- Removed the `HvDatePickerStatus` type in favor of `HvFormStatus`
  - I opted to leave the `empty` value for the `HvDatePicker` status as it might make sense to have it there
- Updated the `DatePicker` stories to have a "Variants"  story instead of multiple unnecessary stories for each variant

### DIFF
--- a/packages/core/src/components/DatePicker/DatePicker.tsx
+++ b/packages/core/src/components/DatePicker/DatePicker.tsx
@@ -13,6 +13,7 @@ import {
   HvTypography,
   HvBaseDropdown,
   HvLabel,
+  HvFormStatus,
 } from "@core/components";
 import { useControlled, useLabels, useTheme, useUniqueId } from "@core/hooks";
 import { HvBaseProps } from "@core/types";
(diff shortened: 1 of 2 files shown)


packages/core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -34,8 +35,6 @@ const DEFAULT_LABELS = {
   invalidLabel: "Invalid date",
 };
 
-export type HvDatePickerStatus = "standBy" | "valid" | "invalid";
```
zettca, line 37: Not only this type is being exported to the users, it's being used in our samples. Also, `HvDatePickerProps["status"]` isn't a great alternative since you'll get the `undefined`.
Therefore, this is a type that users are probably using it in the apps. We should deprecate it, instead of removing it

We can do something like:
```tsx
/** @deprecated use `HvFormStatus` instead */
export type HvDatePickerStatus = HvFormStatus;
```

plagoa: Yeah, you're right. I'll update, thanks.

MEsteves22 (approved):

MEsteves22:
🚀

I would just give more height to the Variants story because we have to scroll down if we open the calendar.

zettca (approved):

PR #3458, opened 2023-06-23
chore: remove useEffect causing infinite re-renders on the Theme Creator

plagoa:
- removed unnecessary evil useEffect causing infinite re-renders on the Theme Creator


zettca (approved):
👋 😈

MEsteves22:
I noticed that if I update the base theme with the dropdown, it's not updating.

PR #3466, opened 2023-06-27
fix(Button): styles & port to `createClasses`

zettca:
- port `HvButton` to `createClasses` util
- fix button fontFamily
- fix outlines showing when interacting with mouse (use focus-visible)
- simplify/refactor styles


packages/core/src/components/Button/Button.tsx
```diff
@@ -106,10 +109,9 @@ export const HvButton: <C extends React.ElementType = "button">(
   ) => {
     const {
```
plagoa, line 110: @zettca can we remove the default value of `base` on the ` radius` prop? I think this is still ignoring the `theme.button.borderRadius` prop, preventing us from using that theme override.

MEsteves22:
I noticed the following for the Multi Button component:
- The icons are missing from the "Vertical Orientation" story 
- The "Disabled" and "Disabled Item" stories are different from what we have in master (extra borders).

packages/core/src/components/Button/Button.tsx
zettca: yeah, makes sense 👍

zettca:
> I noticed the following for the Multi Button component:
> 
> * The icons are missing from the "Vertical Orientation" story
> * The "Disabled" and "Disabled Item" stories are different from what we have in master (extra borders).

I've removed the content container and haven't look at the Applitools diffs yet 😐

MEsteves22 (approved):

PR #3475, opened 2023-06-28
chore: refactor theme creator

plagoa:
- refactor of the logic on the Theme Creator 
- removed undo / redo buttons. 
  - this functionality is available on the `Code Editor` since it's no longer readonly
  - if we find we need these controls, we should consider implementing it by using some library that provides us this functionality.


zettca (approved):
👌

MEsteves22 (approved):

PR #3479, opened 2023-06-29
fix: export HvDrawer

Sam-Cooksley:
- Added Drawer export in components `index.ts`

Let me know if I've missed something.


zettca (approved):
Thanks for the pull request! 🙏 

🚀

PR #3481, opened 2023-06-29
chore(POC): add useDefaultProps & usage in Avatar/ActionBar

zettca:
- add `useDefaultProps` hook to allow overriding component props at theme level
  - example `HvAvatar` usage for `size`/`variant` props
- add special logic for `classes` prop, so that users can pass `CSSInterpolation` and the name is computed
  - example `HvActionBar` usage for DS3 vs DS5 theming via `classes`


zettca:
Will re-open later with the finalised changes

PR #3482, opened 2023-06-30
chore: fix theme creator app not considering the removal of customisations from the code editor

plagoa:
whenever we were manually **removing** anything from the Code Editor those changes weren't being taken into consideration since we're diffing with the current changes with the previous changes and merging the result:

```
const diff = themeDiff(prev, changes);
return merge({}, prev, diff);
```

but the previous versions include what we're removing so they're never actually removed. I added a `isCodeEdit`prop because editing the code manually is actually different from making changes using the other controls.


zettca (approved):
LGTM 👍

plagoa:
@zettca I made some changes, also of some issues that I found. Please take a look and test all the variations of customizations you can think of (using the code editor, using the controls, mixed changes, etc etc). I'll keep this a draft while we test it.

PR #3485, opened 2023-07-04
fix: a menu item with a child with a link should also be a link

andrefrua:


zettca (approved):
Hmm, this looks like a regression from a recent change.

Thanks for the fix 👍 🙏

francisco-guilherme (approved):

PR #3493, opened 2023-07-05
fix(InlineEditor): losing variant styles when in edit mode

plagoa:
- moved the `getVarValue` function to a centralized place in the `styles` package (it was being defined differently in the `core` package and in the `app`).
- updated all uses of this function
- updated snapshots

### DIFF
--- a/packages/core/src/components/InlineEditor/InlineEditor.tsx
+++ b/packages/core/src/components/InlineEditor/InlineEditor.tsx
@@ -1,12 +1,7 @@
 import React, { useLayoutEffect, useRef, useState } from "react";
 import { HvBaseProps } from "@core/types";
 import { useControlled } from "@core/hooks";
-import {
-  ExtractNames,
-  getVarValue,
-  isKeypress,
-  keyboardCodes,
-} from "@core/utils";
+import { ExtractNames, isKeypress, keyboardCodes } from "@core/utils";
 import {
   HvButtonProps,
   HvTypographyVariants,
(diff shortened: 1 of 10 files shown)


packages/styles/src/utils.ts
```diff
@@ -130,3 +130,14 @@ export const getThemesVars = (themes: HvThemeStructure[]) => {
 
   return vars;
 };
+
+export const getVarValue = (cssVar: string, rootElementId?: string) => {
+  const root = document.getElementById(rootElementId || "hv-root");
```
zettca, line 135: It's a bit strange the `styles` package is manipulating/accessing the DOM.
Also, isn't this function a "hack" needed just in some core components?

packages/core/src/components/InlineEditor/InlineEditor.tsx
```diff
@@ -82,8 +82,12 @@ export const HvInlineEditor = ({
   const [editMode, setEditMode] = useState(false);
   const [cachedValue, setCachedValue] = useState(value);
   const inputRef = useRef<HTMLInputElement>();
+  const { rootId } = useTheme();
 
-  const typographyStyles = getTypographyStyles(theme.typography[variant] || {});
+  const typographyStyles = getTypographyStyles(
+    theme.typography[variant] || {},
+    rootId
+  );
```
zettca, line 90: Why not use the object value?
```suggestion
  const typographyStyles = activeTheme?.typography[variant] || {};
```

packages/styles/src/utils.ts
plagoa: Yeah, you're right, I'll revert that.

plagoa:
@zettca updated the PR with the changes you mentioned.

zettca (approved):
nice 👍

PR #3503, opened 2023-07-10
feat(Dialog): variant prop added for dialog status bar

MEsteves22:
- `variant` prop added to `HvDialog` to show a status bar at the top of the dialog.

### DIFF
--- a/packages/core/src/components/Dialog/Dialog.tsx
+++ b/packages/core/src/components/Dialog/Dialog.tsx
@@ -49,9 +49,12 @@ export interface HvDialogProps
   disableBackdropClick?: boolean;
   /** A Jss Object used to override or extend the styles applied to the component. */
   classes?: HvDialogClasses;
+  /** Variant of the dialog. Adds a status bar to the top of the dialog. If not provided, no status bar is added. */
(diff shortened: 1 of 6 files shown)


HQFOX (approved):

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -49,9 +49,12 @@ export interface HvDialogProps
   disableBackdropClick?: boolean;
   /** A Jss Object used to override or extend the styles applied to the component. */
   classes?: HvDialogClasses;
+  /** Variant of the dialog. Adds a status bar to the top of the dialog. If not provided, no status bar is added. */
+  variant?: "success" | "error" | "warning";
```
zettca, line 53: Alignment with Banner/Snackbar is nice 👍 

But what if users want a **catastrophic** Dialog? Should we allow any semantic color? 🤔 DS also calls these _Negative_, _Warning_, and _Positive_.

To me it's fine either way. If eventually we need to support any `semantic`/`color`, we could always deprecate this in favour of a more generic color prop

zettca (approved):
LGTM 👍

PR #3533, opened 2023-07-20
feat(grid): allow variable column number following design system specs

nantunes:


packages/core/src/components/Grid/Grid.tsx
```diff
@@ -5,11 +5,11 @@ import { HvBaseProps } from "@core/types";
...
-  xs: 2,
-  sm: 2,
-  md: 4,
-  lg: 4,
-  xl: 4,
+  xs: "15px",
+  sm: "15px",
+  md: "30px",
+  lg: "30px",
+  xl: "30px",
```
plagoa, line 12: Can you update the docs on this same file (around line 124) to reflect these changes?

nantunes: No, changes reverted, I was wrong. 32/16 is ok.

plagoa (approved):

PR #3549, opened 2023-08-01
chore: port avatar and badge to create classes

MEsteves22:
- The avatar and badge components were refactored to use the `createClasses` util.
- A small refactor was needed for the step navigation styles.
- In the badge component I changed the prefix `HvBadge-*` to `HvBadgeClasses-*` to maintain consistency with the other components.


packages/core/src/components/Avatar/Avatar.tsx
```diff
@@ -154,48 +155,28 @@ export const HvAvatar = ({
...
-          className={clsx(
-            className,
-            avatarClasses.root,
-            classes?.root,
-            avatarClasses.avatar,
-            classes?.avatar,
-            avatarClasses[size],
-            classes?.[size]
-          )}
+          className={cx(classes.root, classes.avatar, classes[size], className)}
```
zettca, line 171: this is definitely not the `root`... 🤷

zettca (approved):

packages/core/src/components/Avatar/Avatar.tsx
MEsteves22: Yup, it doesn't make much sense to have the `root` and `className` here in my opinion 😅 I didn't change this because this can lead to breaking changes for some users.

francisco-guilherme (approved):

PR #3570, opened 2023-08-04
refactor(Header): port to create classes

MEsteves22:


packages/core/src/components/Header/Actions/Actions.styles.tsx
```diff
@@ -1,19 +1,20 @@
...
+export const { staticClasses, useClasses } = createClasses("HvHeader-Actions", {
+  root: {
+    backgroundColor: "transparent",
+    display: "flex",
+    gap: theme.space.xs,
+    alignItems: "center",
+    marginLeft: "auto",
+    "& > *": {
+      marginLeft: theme.space.xs,
+    },
```
zettca, line 15: Unrelated to the refactor, but can we remove this? 😬

```suggestion
```

I noticed that we have duplicated spacing (`gap` + `marginLeft`)

packages/styles/src/theme.ts
```diff
@@ -59,6 +59,10 @@ const componentsSpec: DeepString<HvThemeComponents> = {
     secondLevelItemColor: "string",
     secondLevelSelectedItemColor: "string",
     secondLevelSelectedItemBackgroundColor: "string",
+    selectedItemBorderRadius: "string",
+    secondLevelSelectedItemBorderRadius: "string",
+    itemPadding: "string",
+    secondLevelItemPadding: "string",
```
zettca, line 65: It this API something we want to keep extending? 🤔

packages/core/src/components/Header/Navigation/MenuBar/Bar.styles.tsx
```diff
@@ -14,69 +10,57 @@ const show = {
...
+  },
+  hidden: { ...hide },
+  active: { ...show },
+  list: {
+    margin: 0,
+    padding: 0,
+    display: "inherit",
+    alignItems: "center",
+    height: "100%",
+    [`&:hover $active`]: {
```
zettca, line 50: No need for the computed keys

```suggestion
    "&:hover $active": {
```

zettca (approved):
👍
Some minor things that could be reviewed later

packages/styles/src/theme.ts
MEsteves22: Not really 😅 I only added this because we had some styling on the header component that only seemed to have been added for white labelling purposes (to get something like we have [here](https://lumada-design.github.io/uikit/master/?path=/docs/guides-theming-white-labeling--page)). 
Instead of removing the styling that was unnecessary I decided to add these tokens in case this is something we want to continue supporting. 
However, this is definitely something we need to look at when we review the theme.

packages/core/src/components/Header/Navigation/MenuBar/Bar.styles.tsx
MEsteves22: Thanks, I didn't notice! I will update 🙏

packages/core/src/components/Header/Actions/Actions.styles.tsx
MEsteves22: I'll take a look at this!

HQFOX (approved):

PR #3571, opened 2023-08-07
fix: add color contrast ignore rule to the pa11y config

plagoa:
we already have `color-contrast` as an ignore rule on our pa11y config file. This adds the actual rule that is being vioated on our color contrast violations.


zettca (approved):
👍

PR #3575, opened 2023-08-07
refactor: port Footer to create classes

plagoa:


packages/core/src/components/Footer/Footer.tsx
```diff
@@ -30,43 +26,37 @@ export const HvFooter = (props: HvFooterProps) => {
...
+  const { classes, cx } = useClasses(classesProp);
+
+  const isSmDown = useMediaQuery(muiTheme.breakpoints.down("sm"));
 
   return (
-    <StyledRoot
-      $breakpoints={muiTheme.breakpoints}
-      className={clsx(className, footerClasses.root, classes?.root)}
+    <footer
+      className={cx(className, classes.root, isSmDown && classes.small)}
```
zettca, line 40:
```suggestion
      className={cx(classes.root, className, {[classes.small]: isSmDown})}
```
- `className` needs to come after our internal styles (ie. `classes.root`)
- the `{[value]: condition}` `sx` syntax is preferable as it's easier to maintain when adding more conditionals (also, consistency).

This `classes.small` is a new class that will be exposed to the users now. (I'm not against it), but overriding `classes.small` would mean overriding multiple (unrelated) elements. We should keep `classes.X` on a single element type, otherwise it's confusing and not useful when overriding. We should only need `classes.small` here in the root

packages/core/src/components/Footer/Footer.styles.tsx
```diff
@@ -1,74 +1,51 @@
...
     backgroundColor: theme.colors.secondary,
     margin: `${theme.space.xs} ${theme.space.sm}`,
-    [$breakpoints.down("sm")]: {
-      display: "none",
-    },
-  })
-);
+    "&$small": { display: "none" },
+  },
+  small: {},
```
zettca, line 50: Regarding the point below about `classes.small` being in a single element, it would also simply styling here:
```suggestion
  small: {
    "&$root": {
      height: "unset",
      flexDirection: "column",
      padding: theme.space.xs,
    },
    "&$name": ...
  },
```

([example on another conditional class](https://github.com/lumada-design/hv-uikit-react/blob/master/packages/core/src/components/BaseCheckBox/BaseCheckBox.styles.ts#L15-L26))

packages/core/src/components/Footer/Footer.tsx
```diff
@@ -30,43 +26,37 @@ export const HvFooter = (props: HvFooterProps) => {
...
       {...others}
     >
-      <StyledName
-        $breakpoints={muiTheme.breakpoints}
-        variant="label"
-        className={clsx(footerClasses.name, classes?.name)}
+      <HvTypography
+        variant="highlightText"
+        className={cx(classes.name, isSmDown && classes.small)}
```
zettca, line 45:
```suggestion
        className={classes.name}
```

(see above)

plagoa:
@zettca updated the PR to address your comments

MEsteves22 (approved):

zettca (approved):

PR #3576, opened 2023-08-07
fix: polymorphic props for Typography and TimeAgo

MEsteves22:
- Polymorphic pops added for the `Typography` component. I also needed to add polymorphic pops to the `TimeAgo` component.


zettca (commented):
nice 👌 

Can we add a test for an `HvTypography`  with a custom `component` too? 😄

MEsteves22:
@zettca Yes, I will add the tests 🙏

zettca (approved):

PR #3578, opened 2023-08-08
refactor(Tabs): port to create classes

plagoa:


packages/core/src/components/Tabs/Tabs.tsx
```diff
@@ -31,15 +31,21 @@ export interface HvTabsProps
  * Tabs can be used as a navigational widget for switching between sets of documents.
  */
 export const HvTabs = (props: HvTabsProps) => {
-  const { classes, ...others } = useDefaultProps("HvTabs", props);
+  const {
+    classes: classesProp,
+    className,
```
MEsteves22, line 36: `className` is never used on MUI's `Tabs` component.

packages/core/src/components/Tabs/index.ts
```diff
@@ -1,3 +1 @@
-export { default as tabsClasses } from "./tabsClasses";
-export * from "./tabsClasses";
```
MEsteves22, line 2: I believe the `tabsClasses` file was not removed from the code base.

packages/core/src/components/Tabs/Tabs.styles.tsx
```diff
@@ -1,28 +1,26 @@
...
     height: 2,
   },
-  "& .MuiTabs-scroller": {
+  scroller: {
     overflow: "visible !important",
   },
-  "& .MuiTabs-flexContainer": {
-    "& button:first-of-type": {
+  flexContainer: {
+    "& button:first-child": {
```
zettca, line 23: We should avoid `:first-child` (see https://github.com/lumada-design/hv-uikit-react/pull/3399)

MEsteves22 (approved):

zettca (approved):

PR #3579, opened 2023-08-08
fix(Table): list sticky styles

zettca:


packages/core/src/components/Table/utils/utils.ts
```diff
@@ -1,49 +1,22 @@
 import { theme } from "@hitachivantara/uikit-styles";
 
 export const getBorderStyles = (
-  type: string,
-  color: string,
-  borderRadius?: string
+  type: "row",
+  color = theme.table.rowBorderColor,
+  rowBorderRadius = theme.table.rowBorderRadius
```
zettca, line 6: Just a refactor to reduce duplicated logic, but only `type: "row"` is being used 🤷

MEsteves22 (approved):

plagoa (approved):

PR #3583, opened 2023-08-08
feat(colorPicker): expose onChangeComplete property

HQFOX:


MEsteves22:
On my end, I noticed that if I use the onChangeComplete handler with a controller value, the color picker is very slow:

```
const [value, setValue] = useState<string | undefined>("#95AFE8");

    return (
      <div style={{ width: "134px" }}>
        <HvColorPicker
          {...args}
          value={value}
          onChangeComplete={(color) => setValue(color)}
        />
      </div>
    );
```

MEsteves22 (approved):

plagoa (approved):

PR #3587, opened 2023-08-08
refactor(GlobalActions): port to create classes

plagoa:


packages/core/src/components/GlobalActions/GlobalActions.styles.tsx
```diff
@@ -1,98 +1,106 @@
...
+  },
+  content: {
+    width: "100%",
+    padding: theme.space.xs,
+    display: "flex",
+    position: "relative",
+    alignItems: "center",
+    paddingTop: 17,
+    paddingBottom: 17,
+  },
```
MEsteves22, line 57: This `content` class doesn't seem to be used in the component.

packages/core/src/components/GlobalActions/GlobalActions.tsx
```diff
@@ -60,62 +50,56 @@ export const HvGlobalActions = (props: HvGlobalActionsProps) => {
...
+          width: `calc(100% - 2 * ${theme.spacing(4)})`,
+          marginLeft: `${theme.spacing(4)}`,
+          marginRight: `${theme.spacing(4)}`,
+        }
+      : isSmDown
+      ? {
+          width: `calc(100% - 2 * ${theme.spacing(2)})`,
+          marginLeft: `${theme.spacing(2)}`,
+          marginRight: `${theme.spacing(2)}`,
+        }
```
MEsteves22, line 70: This `isSmDown` styling should only be used if `positionProp` is fixed but in this case it's used even if it's not fixed.

MEsteves22 (approved):

packages/core/src/components/GlobalActions/GlobalActions.tsx
```diff
@@ -60,62 +51,58 @@ export const HvGlobalActions = (props: HvGlobalActionsProps) => {
...
-        position === "fixed" &&
-          clsx(classes?.positionFixed, globalActionsClasses.positionFixed),
-        variant === "global" &&
-          clsx(classes?.global, globalActionsClasses.global)
+        {
+          [classes.positionSticky]: position === "sticky",
+          [classes.positionFixed]: position === "fixed",
+          [classes.global]: variant === "global",
+        },
+        css(fixedPositionCss)
```
zettca, line 97: This should come before `className`

packages/core/src/components/GlobalActions/GlobalActions.stories.tsx
```diff
@@ -21,27 +21,25 @@ export const Main: StoryObj<HvGlobalActionsProps> = {
...
   decorators: [(Story) => <div style={{ height: 300 }}>{Story()}</div>],
-  render: ({ title, headingLevel, variant }) => {
+  render: (args) => {
     const BackButton = () => (
       <HvButton aria-label="Back" icon onClick={() => alert("Back!")}>
         <Backwards />
       </HvButton>
     );
 
+    console.log(args);
```
zettca, line 38: leftovers

packages/core/src/components/GlobalActions/GlobalActions.tsx
```diff
@@ -60,62 +51,58 @@ export const HvGlobalActions = (props: HvGlobalActionsProps) => {
...
+  const isSmDown = useMediaQuery(muiTheme.breakpoints.down("sm"));
+  const isUpMd = useMediaQuery(muiTheme.breakpoints.up("md"));
 
-  const { activeTheme } = useHvTheme();
-
-  const muiTheme = useTheme();
+  const fixedPositionCss =
+    positionProp === "fixed" &&
+    (isUpMd
+      ? {
```
zettca, line 63: An idea: we could move this to the styles file (outside `createClasses`) as an utility function, since this styling block is kinda large. something like

```ts
positionProp === "fixed" && getBreakpointStyles(isUpMd, isSmDown)
```

zettca (approved):

PR #3591, opened 2023-08-09
refactor(Tab): port to create classes

plagoa:


packages/core/src/components/Tab/Tab.tsx
```diff
@@ -26,27 +27,25 @@ export interface HvTabProps
...
       disableTouchRipple
-      // Exposes the global class HvIsFocusVisible as a marker not to
-      // be styled directly, but only as helper in specific css queries
-      focusVisibleClassName={clsx(
-        "HvIsFocusVisible",
-        clsx(tabClasses.focusVisible, classes?.focusVisible)
-      )}
+      // expose the global class HvIsFocusVisible as a marker
+      // not to be styled directly, only as helper in specific css queries
+      focusVisibleClassName={clsx("HvIsFocusVisible", classes.focusVisible)}
```
MEsteves22, line 48: We should use `cx` from `useClasses` instead of `clsx`.

packages/core/src/components/Tab/Tab.styles.tsx
```diff
@@ -1,64 +1,54 @@
...
-  "&.Mui-disabled": {
-    color: theme.colors.secondary_60,
-    cursor: "not-allowed",
-    pointerEvents: "all",
-    opacity: 1,
-    "&:hover": {
-      background: "none",
-    },
-  },
-  "& .MuiTab-iconWrapper": {
```
MEsteves22, line 61: This is missing from the refactor to remove the "extra" margin on the icon.

MEsteves22 (approved):

zettca (approved):

PR #3592, opened 2023-08-09
refactor(ListContainer): port to create classes

MEsteves22:
- `ListContainer` ported to create classes.
- Typing updated.
- Some files were updated to list `className` as the last class.


packages/core/src/components/ListContainer/ListItem/ListItem.tsx
```diff
@@ -176,54 +184,48 @@ export const HvListItem = (props: HvListItemProps) => {
...
       )}
-      $gutters={!disableGutters}
-      $interactive={interactive}
-      $disabled={disabled || false}
-      $selected={selected || false}
-      $startAdornment={startAdornment != null}
-      $endAdornment={endAdornment != null}
+      {...(!!interactive && {
+        onClick: handleOnClick,
+      })}
```
zettca, line 206: I'm not sure about making this change now... `HvListItem` is one of those components that are used everywhere, someone might be using `onClick` without `interactive`...

zettca (approved):
Unsure about the `onClick`, but not against it. Just thought I'd mention it. ✅

packages/core/src/components/ListContainer/ListItem/ListItem.tsx
MEsteves22: I'll change this and put it back as it was since we don't know what other people are doing

MEsteves22: Updated 🚀

plagoa (approved):

PR #3597, opened 2023-08-10
chore(ThemeCreator): refactor use of context

plagoa:
Updated the use of the context in the Theme Creator. Created a new hook that deals with validating if the context exists, allowing the `GeneratorContext` to be initialized as null instead of actually unnecessarily creating a new theme just for that purpose. The components that need to use the context use this custom hook instead of having to import `useContext` and `GeneratorContext` everytime.


zettca (approved):
👌

MEsteves22 (approved):

PR #3599, opened 2023-08-10
refactor(Tag): port to create classes

plagoa:


packages/core/src/components/Tag/Tag.styles.tsx
```diff
@@ -113,44 +44,95 @@ export const StyledChip = styled(
...
+      outline: "none",
+    },
+  },
+
+  semanticTextColor: {
+    color: theme.colors.base_light,
+  },
+  categoricalTextColor: {
+    color: theme.colors.secondary,
+  },
```
MEsteves22, line 137: The classes `categoricalTextColor`, `semanticTextColor`, and `titleOverflow` doesn't seem to be used in the component.

packages/core/src/components/Tag/Tag.tsx
```diff
@@ -94,34 +94,38 @@ export const HvTag = (props: HvTagProps) => {
...
     return (
-      <StyledButton
+      <HvButton
         classes={{
-          startIcon: clsx(tagClasses.tagButton, classes?.tagButton),
-          focusVisible: clsx(tagClasses.focusVisible, classes?.focusVisible),
-          root: clsx(tagClasses.button, classes?.button),
+          startIcon: classes.tagButton,
+          focusVisible: classes.focusVisible,
+          root: classes.primaryButton,
```
MEsteves22, line 116: We should continue to use the `button` class instead of `primaryButton` since it's a breaking change.

MEsteves22 (approved):

Issue #3525, opened 2023-07-14
[Header] - With more than two levels of items the first menu doesn't appear as selected


andrefrua:
### Which version of NEXT UI Kit are you opening this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

Currently when sending more than 2 levels of items to the Header component the menu on the top level isn't getting selected as you can see in the image below:
![image](https://github.com/lumada-design/hv-uikit-react/assets/26412841/b9f0d4cb-a135-4f62-a493-3c0f5e165b77)

The navigation data used to replicate the issue is the following:
```
[
  {
    id: "1",
    label: "Overview",
    data: [
      {
        id: "1-1",
        label: "Model Effectiveness 1",
        href: "/overview/model-effectiveness",
      },
      {
        id: "1-2",
        label: "Trend Analysis 1-2",
        href: "/overview/trend-analysis",
      },
    ],
  },
  {
    id: "2",
    label: "Events",
    href: "/events",
  },
  {
    id: "3",
    label: "Work Orders",
    data: [
      {
        id: "3-1",
        label: "Model Effectiveness 3-1",
        data: [
          {
            id: "3-1-1",
            label: "Another Sub Level 3-1-1",
            href: "/work-orders/model-effectiveness-3-1-1",
          },
          {
            id: "3-1-2",
            label: "Another Sub Level 3-1-2",
            href: "/work-orders/trend-analysis-3-1-2",
          },
        ],
      },
      {
        id: "3-2",
        label: "Trend Analysis 3-2",
        href: "/work-orders/trend-analysis",
      },
    ],
  },
  {
    id: "4",
    label: "Assets",
    href: "/assets",
  },
  {
    id: "5",
    label: "Analytics",
    data: [
      {
        id: "5-1",
        label: "Model Effectiveness 5-1",
        href: "/analytics/model-effectiveness",
      },
      {
        id: "5-2",
        label: "Trend Analysis 5-2",
        href: "/analytics/trend-analysis",
      },
    ],
  },
];
```

### Expected behavior 🤔

When a sub item is selected the top menu should also appear selected

zettca:
Sorry for the delay @andrefrua. We'll look into this 👍

plagoa:
@andrefrua please check the PR here:
https://github.com/lumada-design/hv-uikit-react/pull/3600

[...]

PR #3600, opened 2023-08-10
fix(Header): aria-current set to true if item has sub levels

MEsteves22:
This PR fixes https://github.com/lumada-design/hv-uikit-react/issues/3525:
- The `aria-current` is now set to `true` if the selected item has sub levels. Otherwise, it's set to `page`.
- The header has been limited to 2 levels since using more levels leads to weird behaviours and it's [not supported in DS](https://designsystem.hitachivantara.com/6c705d900/v/0/p/90fb01-navigation-system/b/07cb30).
- A new story was added to showcase how this can be used.


andrefrua:
LGTM 👍

PR #3601, opened 2023-08-10
feat(icons): add icon sprites

zettca:
**New stuff**
- add SVG sprite generator
  - add `icons.svg` and `pictograms.svg` sprites to bundle
- add `HvIconSprite` helper component to load sprites
- refactor code share logic between React icons & icon sprites
  - React icons logic was mostly build-time, sprites logic has to be runtime
  - replace `colorArray` with CSS vars

**Notes**
- `semantic` and `inverted` were not included in icon sprites, as they can be achieved using the `color` prop
- for the same reason, `semantic` and `inverted` were deprecated in favour of `color`


docs/foundation/Icons/Usage.stories.tsx
```diff
@@ -40,11 +41,19 @@ IconSize.parameters = {
...
 
 export const CustomColors = () => (
-  <Bookmark
-    color={["brand", "secondary"]}
-    iconSize="M"
-    aria-label="Click to bookmark"
-  />
+  <div style={{ display: "flex", gap: 8 }}>
+    <Bookmark
+      color={["brand", "secondary"]}
```
nantunes, line 46: For staying as it was, now it should be:
```suggestion
      color={["secondary", "brand"]}
```

Also, why did you add more icons to the sample? I'm ok with it (it showcases more), but then I would add them to the equivalent sample on Sprites stories.

docs/foundation/Icons/Usage.stories.tsx
```diff
@@ -67,7 +76,7 @@ DecorativeIcon.parameters = {
 };
 
 export const SemanticIcon = () => (
-  <Level4 role="img" title="Warning!" iconSize="M" semantic="negative" />
+  <Level4 role="img" title="Warning!" iconSize="M" color="negative" />
```
nantunes, line 79: Given the sample above, do we need to maintain this one?

packages/icons/src/utils/colors.ts
```diff
@@ -6,7 +6,8 @@ export const extractColors = (fileData: string) => {
   let matcher;
 
   const result = {
-    colorArray: [],
+    // ensure the main color (#414141) is always first (all users use it)
```
nantunes, line 9: Maybe:
```suggestion
    // ensure the main color (#414141) is always first (all icons use it)
```
?

nantunes (approved):
Great!

Maybe we can also package sprites with the most commonly used icons, but that can be evaluated in the future.

docs/foundation/Icons/Usage.stories.tsx
zettca: I'll keep this one and remove the others 👍

docs/foundation/Icons/Usage.stories.tsx
zettca: I moved them from the `inverted` sample and added an extra one, but honestly they don't add much, and maybe even _clutters_ the story (we're using 1 icon per sample). I'll remove

MEsteves22 (approved):

PR #3605, opened 2023-08-11
refactor(ThemeCreator): refactor unit slider

plagoa:


app/src/generator/Radii/Radii.tsx
```diff
@@ -16,19 +16,57 @@ const Radii = () => {
   const [currValues, setCurrValues] = useState<Map<string, string | number>>(
     new Map<string, string | number>()
   );
+  const currValuesRef = useRef<Map<string, string | number>>(currValues);
```
zettca, line 19: We could type the radius better, with `Radius`* instead of `string` (here and everywhere else)

```tsx
type Radius = keyof HvThemeTokens["radii"];
```
^ an helper alias

```suggestion
  const currValuesRef = useRef<Map<Radius, string | number>>(currValues);
```

Actually, here the type can be inferred
```suggestion
  const currValuesRef = useRef(currValues);
```

app/src/generator/Radii/Radii.tsx
```diff
@@ -16,19 +16,57 @@ const Radii = () => {
...
 
-  const valueChangedHandler = (spacing: string, value: string) => {
+  const valueChangedHandler = (radii: string, value: string) => {
```
zettca, line 21: same as above
```suggestion
  const valueChangedHandler = (radius: Radius, value: string) => {
```
This way we can avoid the evil `as` casts

app/src/generator/Radii/Radii.tsx
```diff
@@ -16,19 +16,57 @@ const Radii = () => {
...
+    currValuesRef.current = map;
     setCurrValues(map);
   };
 
   const setValueHandler = (radii: string) => {
-    const radiiValue = currValues.get(radii) || 0;
+    const value = currValuesRef.current.get(radii);
+    if (!value) return;
+
+    const radiiValue = parseInt(value.toString(), 10) || 0;
```
zettca, line 39:
```suggestion
    const radiiValue = Number(value) || 0;
```
`Number` is more modern - no need for `radix` and accept any type

zettca (approved):
LGTM 👍

MEsteves22 (approved):

app/src/generator/Radii/Radii.tsx
plagoa: Cool, I'll make those changes

PR #3613, opened 2023-08-16
fix(Icons): fix aria-label set on the wrong element

plagoa:
- pass a11y props to the `svg` element instead of the div container
  - use `splitIconProps` function to do this
  - also refactor focusable and size props to use `splitIconProps`
- fix tests using presentation role and label

### DIFF
--- a/packages/core/src/components/Badge/Badge.test.tsx
+++ b/packages/core/src/components/Badge/Badge.test.tsx
@@ -40,11 +40,11 @@ describe("Badge", () => {
       <HvBadge
         count={100}
         showCount
-        icon={<Alert role="presentation" aria-label="Alert" />}
+        icon={<Alert role="img" aria-label="Alert" />}
       />
     );
     expect(screen.queryByText("99+")).toBeInTheDocument();
-    expect(screen.getByLabelText("Alert")).toBeInTheDocument();
+    expect(screen.getByRole("img", { name: "Alert" })).toBeInTheDocument();
   });
 
   it("should render correctly with custom label", () => {
--- a/packages/core/src/components/Forms/Adornment/Adornment.test.tsx
+++ b/packages/core/src/components/Forms/Adornment/Adornment.test.tsx
@@ -7,13 +7,9 @@ import { HvAdornment } from "@core/components";
 describe("Adornment", () => {
   it("should render the passed icon", () => {
     render(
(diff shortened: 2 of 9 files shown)


packages/core/src/components/Badge/Badge.test.tsx
```diff
@@ -44,9 +44,7 @@ describe("Badge", () => {
       />
     );
     expect(screen.getByLabelText("99+")).toBeInTheDocument();
-    expect(
-      screen.getByRole("presentation", { name: "Alert" })
-    ).toBeInTheDocument();
+    expect(screen.getByLabelText("Alert")).toBeInTheDocument();
```
zettca, line 47: We should fix the test `icon`, `role="presentation"` with `aria-label` isn't correct - we should make it `role="img"`
```suggestion
    expect(screen.getByRole("img", {name: "Alert"})).toBeInTheDocument();
```

packages/core/src/components/Forms/Adornment/Adornment.test.tsx
```diff
@@ -11,9 +11,7 @@ describe("Adornment", () => {
         icon={<CloseXS role="presentation" aria-label="close icon" />}
       />
     );
-    expect(
-      screen.getByRole("presentation", { name: "close icon" })
-    ).toBeInTheDocument();
+    expect(screen.getByLabelText("close icon")).toBeInTheDocument();
```
zettca, line 14: same as above

packages/icons/src/utils/converter/generateComponent.ts
```diff
@@ -72,14 +72,20 @@ export const ${componentName} = ({
   semantic,
   inverted = false,
   svgProps,
+  "aria-label": ariaLabel,
+  "aria-labelledby": ariaLabelledBy,
+  "aria-describedby": ariaDescribedBy,
```
zettca, line 77: We also need the `role` to be passed down to the SVG.

We should refactor this code out of `generateComponent`. Code in this template string get no proper validation/highlighting and will make the package size increase.

We could refactor it into an util, to split `others` props and `svgProps` - something like:

```tsx
const {svgProps, others} = getSvgProps(props);
```

zettca (changes requested):

zettca (approved):

PR #3614, opened 2023-08-16
This is a fix to the current implementation, that follows ARIA 1.1.

nantunes:
With the release of ARIA 1.2, the combobox role underwent significant changes. Given that the ARIA 1.1 pattern had limited support and was not fully 
compatible with assistive technologies, it's beneficial if we adopt the new specification in the near term.


MEsteves22 (approved):

zettca (approved):
👍

PR #3615, opened 2023-08-16
refactor(icons): remove div container

zettca:
testing the impact of replacing the `div` container with `svg`


zettca:
there is a big impact, as expected

PR #3627, opened 2023-08-21
chore: fix ProgressBar stories a11y issues

plagoa:


packages/core/src/components/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -118,8 +118,8 @@ export const Determinate: StoryObj<HvProgressBarProps> = {
           <HvProgressBar
             value={100}
             status="completed"
+            aria-label="Example Determined Progress Bar"
```
zettca, line 121: Can we remove the `labelProps` values? I don't think they're correct.

`role=progressbar` [shouldn't have semantic children](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role#all_descendants_are_presentational), so these `aria-*` should probably live alongside the `progressbar` (like you just did for the `aria-label`) - **but** `progressbar` [already has an implicit `aria-live`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes)

zettca (approved):
👌

PR #3517, opened 2023-07-12
docs: update stories theme when updating docs color mode

MEsteves22:
As discussed, the stories theme is updated when the docs color mode is changed.

PR #3628, opened 2023-08-21
docs(storybook): upgrade to `storybook@7`

zettca:
- upgrade `storybook` dependencies to v7
  - make necessary API changes
- ~replace internal `mode-selector` plugin with [`storybook-dark-mode`](https://storybook.js.org/addons/storybook-dark-mode)~
- reduce CSS styles overrides, as they're more brittle and harder to maintain and upgrade (a few classes changed in SB7)
  - replace with [storybook theming API](https://storybook.js.org/docs/react/configure/theming#create-a-theme-quickstart) whenever possible
  - move component styling overrides from global to decorators or inline
  -


MEsteves22:
LGTM 🙏 I just noticed the following:

- The label has the same color as the button here:

Screenshot 2023-08-24 at 14 46 42

- When I select "ds3 dawn" or "ds5 dawn" and then go to another page, the selected theme/mode changes to "ds3 wicked" and "ds5 wicked" respectively. This happens because I have the docs in dark mode. If I have the docs in light mode, the opposite happens: if I select "ds3 wicked" and go to another page, the selected theme/mode changes to "ds3 dawn". 

- The tooltips on the charts are cut by the story container (we previously has some CSS to avoid this on storybook):

Screenshot 2023-08-24 at 15 04 41

zettca:
Regarding the Theme Switcher thing, `storybook-dark-mode` emits more "switch theme" events than it should, which is leading to the reset.
I decoupled the dark-mode and theme-switcher events, now they won't change. Let me know what you think

Alternatively, I can just drop af956e01f0a446a8505dac9787b58d2657dd18b7 to keep the previous behaviour and components
Let me know what you think!

MEsteves22:
Great 🙏 

I just noticed that we lost a "feature" with the decoupling. Basically, when the user changes the docs color mode (by clicking on the sun/moon button), the stories' color mode should also change. It was added [here](https://github.com/lumada-design/hv-uikit-react/pull/3517).

I also noticed some other small things:
- We should increase the story container's height for the wizard (I'm unable to see the wizard when I open it) and the snackbar main story (the banner is cut out). 
- The links have the same color has the main text. I believe the styles created on the `docs.ts` file are not being applied. Something like this should solve the problem on this file I think:
```
return Object.entries(styles).reduce((acc, curr) => {
    acc[`${curr[0]}:not(.hv-story-sample *):not(.sb-unstyled *)`] = curr[1];

    return acc;
  }, {});
```
- The white labelling sample story is "broken". I believe using the `Unstyled` component should fix it:
```
<Unstyled>
  <WhiteLabeling />
</Unstyled>
```

zettca:
Closing this for now as `eyes-storybook`'s dev-server mode [isn't working](https://github.com/lumada-design/hv-uikit-react/actions/runs/5998528745/job/16267269889?pr=3628) in Storybook 7.
Take a look at https://github.com/lumada-design/hv-uikit-react/pull/3647 for a cherry-pick of some of the upgrade/improvements

PR #3630, opened 2023-08-22
chore: fix Avatar a11y issues

plagoa:


packages/core/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -258,9 +258,10 @@ export const Actions: StoryObj<HvAvatarProps> = {
           overrideIconColors={false}
           component="a"
           href="#profile-url"
+          aria-label="External link"
         >
           <HvAvatar size="md">
-            <LinkIcon color="base_dark" aria-label="External link" />
+            <LinkIcon color="base_light" />
```
zettca, line 264: This way it doesn't work in wicked 😅 This needs to be `atmo1`, like the default Avatar icon

image


```suggestion
            <LinkIcon color="atmo1" />
```

plagoa: Updated!

zettca (approved):

PR #3631, opened 2023-08-22
feat(bulkActions): update do ds5 styles

HQFOX:
The dark mode styles were updated to this specification that I got from DS 
image (3)

### DIFF
--- a/packages/core/src/components/BulkActions/BulkActions.tsx
+++ b/packages/core/src/components/BulkActions/BulkActions.tsx
@@ -126,11 +126,6 @@ export const HvBulkActions = (props: HvBulkActionsProps) => {
   const anySelected = numSelected > 0;
   const isSemantic = semantic && anySelected;
 
-  const dividerColor =
-    activeTheme?.colors?.modes[selectedMode].type === "dark"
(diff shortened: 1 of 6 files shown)


MEsteves22 (approved):

packages/core/src/components/BulkActions/BulkActions.tsx
```diff
@@ -122,6 +126,11 @@ export const HvBulkActions = (props: HvBulkActionsProps) => {
   const anySelected = numSelected > 0;
   const isSemantic = semantic && anySelected;
 
+  const dividerColor =
+    activeTheme?.colors?.modes[selectedMode].type === "dark"
+      ? theme.colors.secondary
+      : theme.colors.atmo4;
```
zettca, line 132: We shouldn't be doing this. This should be solved with the color palette (as we've always done), otherwise we're opening the door to maintainability and customisation issues.
If we need a new color for this, let's add it. Otherwise, we could keep `atmo4`, which is the _specs_ we have been following (zeroheight dawn)

zettca (changes requested):

zettca (approved):

PR #3633, opened 2023-08-22
feat(ConfusionMatrix): confusion matrix added

MEsteves22:
- Confusion matrix component created.
   -  Formats available: `square` (default) or `landscape`.
   -  The confusion matrix uses a default color scale but it can be overwritten. If it's a delta matrix, the `positive`/`negative` colors are used. 
   - Leverages the visual roles to create the confusion matrix. 
      - Should we enable users to provide the matrix and rows/columns values directly in the future?
- More customisations added to the charts: chart width and height, axis position, and axis name customisations.

### DIFF
--- a/packages/viz/src/components/ConfusionMatrix/ConfusionMatrix.tsx
+++ b/packages/viz/src/components/ConfusionMatrix/ConfusionMatrix.tsx
@@ -1,0 +1,271 @@
+import { useMemo } from "react";
+
+import { Arrayable, ExtractNames } from "@hitachivantara/uikit-react-core";
+
+import * as echarts from "echarts/core";
+import { HeatmapChart } from "echarts/charts";
+import {
+  VisualMapComponent,
+  GridComponent,
+  TooltipComponent,
+} from "echarts/components";
+
+import {
+  HvVisualMapHookProps,
+  useData,
+  useGrid,
+  useTooltip,
+  useVisualMap,
+  useXAxis,
+  useYAxis,
+} from "@viz/hooks";
+import { getGroupKey } from "@viz/utils";
+
+import { HvBaseChart } from "../BaseChart";
+import { useClasses } from "./ConfusionMatrix.styles";
+import {
+  HvChartCommonProps,
+  HvChartXAxis,
+  HvChartYAxis,
+} from "../../types/common";
+import { HvChartTooltip } from "../../types/tooltip";
+import { HvConfusionMatrixMeasure } from "../../types/measures";
+import { useColorScale, useGridLayout, useSeries } from "./utils";
+import {
+  HvConfusionMatrixFormat,
+  HvConfusionMatrixColorScale,
+  HvConfusionMatrixValuesProps,
+} from "./types";
+
+// Register chart components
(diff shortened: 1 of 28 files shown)


packages/viz/src/components/ConfusionMatrix/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+                    value: [data.array(filterKey)[i], c, rv != null ? rv : "-"],
+                    ...(hasDeltaScale && {
+                      visualMap: false,
+                      itemStyle: {
+                        color: getDeltaColor(rv, i === j),
+                      },
+                    }),
+                  },
+                ];
+              }, []);
```
zettca, line 116: I worry about the performance in this block, since `data` can grow big and this is quadratic.

Can we use the `array.push(newEl)` instead of `[...array, newEl]`? See: https://jpcamara.com/2023/03/07/making-tanstack-table.html#using-push-directly

packages/viz/src/components/ConfusionMatrix/ConfusionMatrix.tsx
```diff
@@ -0,0 +1,299 @@
...
+  classes?: HvConfusionMatrixClasses;
+  /**
+   * Color scale of the confusion matrix. The first and second values are the lower and upper ends of the scale, respectively.
+   *
+   * The `colorScale` will be ignored when `semanticScale` is provided.
+   * If `delta` is not provided, a default color scale is used when `colorScale` is not defined: `[base-light, cat3]`.
+   */
+  colorScale?: [string, string];
+  /** Semantic scale used to distinguish between "good" and "bad" values. */
+  semanticScale?: HvConfusionMatrixSemanticScale[];
```
zettca, line 85: Wouldn't it be simpler to merge these two into one, since only one of them will be applied and they kinda do the same thing? We would handle the differences internally.

```suggestion
  colorScale?: [string, string] | HvConfusionMatrixSemanticScale[];
```

Also, calling this `semantic` and "good" and "bad" values seems too specific (we also accept atmo colors). What if we want to plot political parties? Which is good, which is bad? 😈 😅
What do you think about `HvConfusionMatrixScale` or `HvConfusionMatrixColorScale`?

packages/viz/src/components/ConfusionMatrix/ConfusionMatrix.tsx
```diff
@@ -0,0 +1,299 @@
...
+    show: !!semanticScale,
+    type: semanticScale ? "piecewise" : "continuous",
+    ...(semanticScale && {
+      pieces: semanticScale.reduce(
+        (
+          prev: (Omit<HvConfusionMatrixSemanticScale, "color"> & {
+            color?: string;
+          })[],
+          curr
+        ) => {
```
zettca, line 146: The typing completions for `color` is nice, but maybe we should just allow any `string`. If it's in our palette, we use it, else we use it as it (like we're doing elsewhere with our `getColor` utils). Would allow passing more colors

zettca (approved):
Looking great 👌

PR #3637, opened 2023-08-24
chore: fix pa11y issues on card, stack, and kpi

MEsteves22:
- Pa11y issues fixed on the card component.
   - Consequently the Pa11y issues with the KPI and stack components were also fixed.
- `semantic` prop updated to `color` for icons since the former is deprecated.

### DIFF
--- a/packages/core/src/components/Kpi/Kpi.stories.tsx
+++ b/packages/core/src/components/Kpi/Kpi.stories.tsx
@@ -98,7 +98,7 @@ export const AverageService: StoryObj<HvKpiProps> = {
         className={css(styles.root)}
         statusColor="positive"
         bgcolor="atmo2"
-        icon={<Level0Good title="Good" semantic="positive" />}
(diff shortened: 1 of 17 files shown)


plagoa (approved):

packages/core/src/components/Kpi/Kpi.stories.tsx
```diff
@@ -200,8 +200,16 @@ export const IOPS: StoryObj<HvKpiProps> = {
         selectable
         selected={selected}
         onClick={() => setSelected(!selected)}
+        tabIndex={0}
+        role="button"
+        onKeyDown={(event) => {
+          if (event.code === "Enter" || event.code === "Space") {
+            setSelected(!selected);
+          }
+        }}
```
zettca, line 209: We should consider creating a BaseButton for these types of cases 🤔

zettca (approved):
Nice 💯

PR #3639, opened 2023-08-24
fix(DropdownMenu): pa11y issues

MEsteves22:
- `role="presentation"` replaced by `aria-hidden` on the dropdown menu component to fix a bunch of these accessibility issues we have on various stories. `aria-hidden` seemed more appropriate since the icon is unimportant and not needed for the accessibility tree.

Screenshot 2023-08-24 at 12 13 41

### DIFF
--- a/packages/core/src/components/DropDownMenu/DropDownMenu.tsx
+++ b/packages/core/src/components/DropDownMenu/DropDownMenu.tsx
@@ -145,7 +145,7 @@ export const HvDropDownMenu = (props: HvDropDownMenuProps) => {
         >
           {icon || (
             <MoreOptionsVertical
-              role="presentation"
+              aria-hidden
               color={disabled ? "secondary_60" : undefined}
             />
           )}


packages/core/src/components/DropDownMenu/DropDownMenu.tsx
```diff
@@ -145,7 +145,7 @@ export const HvDropDownMenu = (props: HvDropDownMenuProps) => {
         >
           {icon || (
             <MoreOptionsVertical
-              role="presentation"
+              aria-hidden
```
zettca, line 148: `role="none"` [would also fix the "issue"](https://github.com/lumada-design/hv-uikit-react/pull/3632), even though it's supposed to be exactly [the same as `presentation`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role) 

This is most likely a bug in `pa11y`/axe. The `svg` has no semantic children.

PR #3640, opened 2023-08-24
fix(BaseChart): chart width and height

MEsteves22:
- I simplified the chart width and height logic because we don't need all this code after all. Also, the following was detected on storybook when changing the theme on the line, bar and donut charts:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/042d1ce4-d3dd-44d7-b707-bdc6c46c3638


plagoa (approved):

zettca (approved):
I never understood why it was necessary 😝

PR #3643, opened 2023-08-25
fix(FileUploader): fix a11y issues

MEsteves22:
The accessibility issues for the file uploader component were fixed:
- All the logic from the `div` with `role="button"` was transferred to the `input` component to only have one interactive element. The tests were updated because of this change.
- The unnecessary `aria-label` from the label/description container was removed.
- I also made a small change to the styling in order for the text inside the file uploader to only take two lines like on DS.

Screenshot 2023-08-25 at 12 11 32


MEsteves22:
The changes on Applitools are expected and will be accepted after this PR is accepted.

MEsteves22:
The changes on Applitools are expected and will be saved after the PR is accepted.

zettca (approved):
Nice 👌

PR #3644, opened 2023-08-25
fix(StepNavigation): fix a11y issues

plagoa:


MEsteves22 (approved):
In this component, we have the following `aria-label` for the steps `aria-label={step-${title}}` which looks like and id. I think it would be better to only use the `title` and remove `step-`.

plagoa:
> In this component, we have `aria-label={step-${title}}` for the steps and looks like an id. I think it would be better to only use the `title`.

Yes, I'll update.

zettca (approved):

PR #3647, opened 2023-08-25
docs(storybook): theming & sample improvements

zettca:
_cherry-pick_ some stuff on https://github.com/lumada-design/hv-uikit-react/pull/3628 to make upgrading simpler

- prepare for storybook@7 upgrade
  - migrate stuff that breaks in v7
  - leverage SB theming capabilities

### DIFF
--- a/.storybook/addons/theme-selector/ThemeSelector.tsx
+++ b/.storybook/addons/theme-selector/ThemeSelector.tsx
@@ -68,7 +68,12 @@ const ThemeSelector = ({ api }) => {
         trigger="click"
         tooltip={<TooltipLinkList links={links} />}
       >
-        <IconButton key={ADDON_ID} active={false} title={"Select theme"}>
+        <IconButton
+          key={ADDON_ID}
+          active={false}
+          title="Select theme"
+          style={{ width: 140 }}
+        >
           {`Theme: ${selectedTheme?.label}`}
         </IconButton>
       </WithTooltip>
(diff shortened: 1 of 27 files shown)


.storybook/addons/theme-selector/ThemeSelector.tsx
```diff
@@ -18,14 +18,16 @@ import {
...
 
 const ThemeSelector = ({ api }) => {
-  const managerStyles = getManagerStyles();
   const themesList = getThemesList(themes);
   const initialTheme = getInitialTheme(themesList);
   const [selectedTheme, setSelectedTheme] = useState(initialTheme);
 
-  const switchMode = (mode: string) => {
-    const currentTheme = getLocalTheme();
+  const currentTheme = getLocalTheme();
```
MEsteves22, line 25: LGTM 👌🏻 Just one small thing:

I think we should switch this code to the way it previously was because:
- `getManagerStyles` doesn't need the parameter
- and if we switch the docs color mode, the wrong theme (ds3/ds5) is selected on the theme selector sometimes because it's using and old stored value:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/f695d7f7-d00f-4955-9ab3-f19d22772394

zettca: I've reverted this and a few more things 👍

MEsteves22 (approved):

PR #3648, opened 2023-08-28
fix(a11y): tooltip labels

zettca:
HvTooltip (actually MuiTooltip) passes aria-label to the tooltip target We shouldn't be wrapping it around divs

- Explicitly remove `aria-label` from OverflowTooltip, as it's incorrect & redundant
- Add roles to elements in HvTooltip samples
- Also remove some redundant/incorrect `aria-labels`

### DIFF
--- a/packages/core/src/components/OverflowTooltip/OverflowTooltip.tsx
+++ b/packages/core/src/components/OverflowTooltip/OverflowTooltip.tsx
@@ -97,7 +97,7 @@ export const HvOverflowTooltip = (props: HvOverflowTooltipProps) => {
     ]
   );
 
-  return open || isOverflowing ? (
+  return (
     <HvTooltip
       id={id}
       disableHoverListener={!isOverflowing}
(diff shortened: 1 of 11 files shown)


packages/core/src/components/OverflowTooltip/OverflowTooltip.tsx
```diff
@@ -97,22 +96,19 @@ export const HvOverflowTooltip = (props: HvOverflowTooltipProps) => {
...
       id={id}
       disableHoverListener={!isOverflowing}
       open={open}
       placement={placement}
-      title={
-        <HvTypography className={classes.tooltipData} variant="body">
-          {data}
-        </HvTypography>
-      }
+      title={data}
```
MEsteves22, line 105: I don't really like the `HvTypography` here too since data is a `React.Node`. However, by doing this we lose the classes `tooltipData` so if someone is using it to style the title it won't work anymore.

zettca: You're right, I overlooked this 😞

We're using `HvTypography` to override MUI text styles... but they're already being overriden, except for the `fontFamily`

MEsteves22 (approved):

PR #3650, opened 2023-08-28
fix(BaseDropdown): aria & role on header element

zettca:


packages/core/src/components/BulkActions/BulkActions.test.tsx
```diff
@@ -120,15 +104,10 @@ describe("BulkActions", () => {
       expect(button2).toBeEnabled();
       expect(button3).toBeEnabled();
 
-      // Open tooltip
+      // Open actions
       fireEvent.click(button3);
-
-      const tooltip = getByRole("tooltip");
-
-      expect(tooltip).toBeInTheDocument();
```
zettca, line 128: We were already using `"tooltip"` as a selector here, teams might be too
But besides being wrong, I can't see a case which the selector would help - the goodies are in the `"menu"`, tooltip is completely redundant here

MEsteves22 (dismissed):

MEsteves22 (approved):

PR #3654, opened 2023-08-30
fix(TimePicker): fix a11y issues

MEsteves22:
- Missing `aria-labelledby` and `aria-describedby` added to the base dropdown on the time picker component. This was leading to accessibility issues on the query builder.
- I removed the `aria-label` from the time picker stories because since the `label` is set it shouldn't necessary to have both. They were previously added to fix the accessibility issues on the time picker stories. This PR fixes these previous issues without having the need to set both the `aria-label` and `label`.


zettca (approved):
Nice 💯

PR #3655, opened 2023-08-30
fix(Table): sort button hover color

MEsteves22:
The hover color for the sort button on the tables was fixed.


packages/core/src/components/Table/TableHeader/TableHeader.styles.tsx
```diff
@@ -102,7 +102,7 @@ export const { staticClasses, useClasses } = createClasses("HvTableHeader", {
       boxShadow: "none",
       backgroundColor: "transparent",
       "&:hover": {
-        backgroundColor: "red",
```
zettca, line 105: 🤣

PR #3658, opened 2023-08-30
fix(Table): a11y issues fixed 

MEsteves22:
- The table samples were switched to the "Visualizations" tab. Because of this pa11y now tests the table samples which was not happening previously. Applitools was not testing the table also.
   - This change increased the number of accessibility issues to more than 200 😅
- Several accessibility issues were fixed on the table (not all of them) and we now have 15 issues on our report 🎉
   - There's only one issue remaining for the table regarding the table header group.
   - The renderers should be reviewed on a next major because they use different logic for accessibility:
      - The switch was using a `switchLabel` but the dropdown an `id`. For the progress line I ended up only using the column id to simplify things.

Screenshot 2023-09-01 at 10 41 31

### DIFF
--- a/packages/core/src/components/Table/TableHeader/TableHeader.tsx
+++ b/packages/core/src/components/Table/TableHeader/TableHeader.tsx
@@ -15,7 +15,7 @@ import { theme } from "@hitachivantara/uikit-styles";
 import { HvTypography, HvTypographyProps } from "@core/components/Typography";
 import { useTheme } from "@core/hooks/useTheme";
 import { ExtractNames } from "@core/utils/classes";
-import { HvButton } from "@core/components/Button";
+import { HvButton, HvButtonProps } from "@core/components/Button";
 
 import TableContext from "../TableContext";
 import TableSectionContext from "../TableSectionContext";
@@ -71,6 +71,8 @@ export interface HvTableHeaderProps
   resizerProps?: HTMLAttributes<HTMLDivElement>;
   /** A Jss Object used to override or extend the styles applied to the component. */
   classes?: HvTableHeaderClasses;
+  /** Extra props to be passed onto the sort button in the header. */
+  sortButtonProps?: HvButtonProps;
 }
 
 const defaultComponent = "th";
@@ -102,6 +104,7 @@ export const HvTableHeader = forwardRef<HTMLElement, HvTableHeaderProps>(
       resizerProps = {},
       resizable = false,
       resizing = false,
+      sortButtonProps,
       ...others
     },
     externalRef
@@ -183,6 +186,8 @@ export const HvTableHeader = forwardRef<HTMLElement, HvTableHeaderProps>(
               className={classes.sortButton}
               icon
               overrideIconColors={false}
+              aria-label="Sort"
+              {...sortButtonProps}
             >
               <Sort className={classes.sortIcon} />
             </HvButton>
--- a/packages/core/src/components/Table/renderers/renderers.tsx
+++ b/packages/core/src/components/Table/renderers/renderers.tsx
@@ -176,6 +176,8 @@ export function hvTagColumn<
   };
 }
 
+// TODO - Review accessibility on the next renderers because they all differ
+
 export function hvSwitchColumn<
   D extends object = Record<string, unknown>,
   H extends HvTableHeaderRenderer | undefined = HvTableHeaderRenderer
@@ -213,7 +215,7 @@ export function hvDropdownColumn<
   H extends HvTableHeaderRenderer | undefined = HvTableHeaderRenderer
 >(
   col: HvTableColumnConfig<D, H>,
-  id: string,
+  id: string | undefined,
   placeholder: string,
   disabledPlaceholder: string,
   onChange?: (identifier: string, value: HvListValue) => void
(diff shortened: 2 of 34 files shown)


packages/core/src/components/Table/renderers/renderers.tsx
```diff
@@ -229,7 +228,8 @@ export function hvDropdownColumn<
           onChange={(val) => onChange?.(row.id, val)}
           disabled={dsbld}
           dropdownProps={{
```
MEsteves22, line 230: I'm not sure about this one because it can lead to breaking changes. It was previously using `setId(id, column.id)` to concat the id provided by the user and the column id. However, it didn't make a lot of sense to be because the user would have to manually set this concat id on the tablet header. 
I can't leave it as it was and only fix the samples. What do you think?

.config/pa11yci.config.js
```diff
@@ -8,7 +8,12 @@ const filterStories = (stories) => {
   const exclude = [
     "Overview/",
     "Foundation/",
-    "Guides/",
+    "Guides/Provider",
+    "Guides/Theming",
+    "Guides/Styling",
+    "Guides/Forms",
+    "Guides/Visualizations",
+    "Guides/Layout",
```
zettca, line 16: What do you think about moving the `HvTable` docs to `Components/`?
It feels really out of place under "Guides", as it's a component and its docs the typical component props + samples, unlike the other guides that we have here

Besides, this is also not being includes in Applitools

packages/core/src/components/Table/TableHeader/TableHeader.tsx
```diff
@@ -183,6 +183,8 @@ export const HvTableHeader = forwardRef<HTMLElement, HvTableHeaderProps>(
               className={classes.sortButton}
               icon
               overrideIconColors={false}
+              tabIndex={-1}
+              aria-hidden
```
zettca, line 187: Could we get rid of the `HvButton`?  🤔

.config/pa11yci.config.js
MEsteves22: Yes, makes sense! but maybe we should add the table stories to the visualisations. In the project status, this component is under visualisations, and also in DS.

packages/core/src/components/Table/TableHeader/TableHeader.tsx
MEsteves22: I left the button because we need the styling from the `HvButton` but I can use `component="div"` to remove the button role all altogether.

zettca: Actually, we need the button behaviour as-is, because the parent `th` doesn't have any tab-stop (maybe it should?), therefore we need the `tabIndex` (therefore we need the role).

.config/pa11yci.config.js
zettca: Yeah, sure! It's where we previously had it

packages/core/src/components/Table/TableHeader/TableHeader.tsx
MEsteves22: Yup, I was looking at the problem the wrong way. I'll leave the button and just add some `aria-label` because that was the accessibility issue we were having.

packages/core/src/components/Table/renderers/renderers.tsx
zettca: If `column.id` automatically correctly labels the element, then fine by me - less work for the user 👍

We could just leave `id` there (unused), that way we wouldn't be _technically_ changing anything in the API

MEsteves22:
I updated my PR with the suggestions and reviewed what was previously done. Let me know what you think.
Regarding `hvDropdownColumn` I left it as it was since the changes I previously did were breaking changes. I just did my some small changes to use the column id in case `setId` is `undefined`.

packages/core/src/components/Table/TableHeader/TableHeader.tsx
```diff
@@ -183,6 +186,7 @@ export const HvTableHeader = forwardRef<HTMLElement, HvTableHeaderProps>(
               className={classes.sortButton}
               icon
               overrideIconColors={false}
+              aria-label={sortButtonAriaLabel}
```
zettca, line 189: can we do `sortProps`, similar to the other elements here? it's more flexible

```suggestion
              aria-label="Sort"
              {...sortProps}
```

MEsteves22: Yup, I'll remove `sortButtonAriaLabel` and a `sortButtonProps`

MEsteves22: Updated

MEsteves22:
The baselines failed because the data is always changing. I'll add some ignore regions once the PR is approved.

MEsteves22:
I removed the random data that was generated on the table samples. This was breaking Applitools. I think it's preferable to visually test the table samples and not have a bunch of ignore regions instead of having all this random data. 

What do you think?

When approved, I'll accept the new baselines.

packages/core/src/components/Table/stories/storiesUtils.tsx
```diff
@@ -46,8 +44,6 @@ export interface AssetEvent {
 // If a Cell gets a value, it has to return a react element
 const getCell = (value: string) => value as unknown as React.ReactElement;
 
-const rand = new Random();
```
zettca, line 49: Strange 🤔 
This "Random" was supposed to generate stable values

zettca (approved):
🚀

PR #3660, opened 2023-09-01
feat(theme): add `colors` helper to `useTheme`

zettca:
making it simpler to get colors from the `activeTheme` & `selectedMode`'s palette

### DIFF
--- a/packages/core/src/hooks/useTheme.ts
+++ b/packages/core/src/hooks/useTheme.ts
@@ -1,9 +1,18 @@
-import { useContext } from "react";
+import { useContext, useMemo } from "react";
+import { HvThemeColorModeStructure } from "@hitachivantara/uikit-styles";
 import {
   HvThemeContext,
   HvThemeContextValue,
 } from "@core/providers/ThemeProvider";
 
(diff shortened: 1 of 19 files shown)


packages/core/src/hooks/useTheme.ts
```diff
@@ -1,9 +1,18 @@
...
 import {
   HvThemeContext,
   HvThemeContextValue,
 } from "@core/providers/ThemeProvider";
 
-export const useTheme = (): HvThemeContextValue => {
-  return useContext(HvThemeContext);
+export const useTheme = (): HvThemeContextValue & {
+  /** Colors of the currently active theme and mode */
+  colors?: HvThemeColorModeStructure;
```
zettca, line 10: Adding this here instead of the context to avoid changing `uikit-react-shared`

MEsteves22 (approved):
LGTM 👌🏻

We just need to update the following text to say that `useTheme` will also return `colors` with all the selected theme colors:

Screenshot 2023-09-01 at 11 58 28

HQFOX (approved):

PR #3662, opened 2023-09-01
refactor(FileUploader): add the progressbar component to remove dupli…

HQFOX:
…cated code.

The `HvFile` component was implementing a progressbar similar to the one that we have has a component. To keep consistency I thought it would be better if it implemented our component. In order to do this, and not alter the look, I had to add a prop to the `HvProgressBar` in order to hide the text label.

### DIFF
--- a/packages/core/src/components/FileUploader/File/File.styles.tsx
+++ b/packages/core/src/components/FileUploader/File/File.styles.tsx
@@ -3,22 +3,12 @@ import { theme } from "@hitachivantara/uikit-styles";
 
 export const { staticClasses, useClasses } = createClasses("HvFile", {
   root: {},
-  progressbar: {
+  progressBar: {
     position: "absolute",
     top: "-1px",
-    width: "80%",
-    height: theme.fileUploader.file.progressHeight,
-    border: `${theme.fileUploader.file.borderWidth} solid ${theme.colors.secondary}`,
-
-    "&::-moz-progress-bar": {
-      background: theme.colors.secondary,
-    },
   },
-  progressbarBack: {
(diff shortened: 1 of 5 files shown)


packages/core/src/components/FileUploader/File/File.styles.tsx
```diff
@@ -3,22 +3,12 @@ import { theme } from "@hitachivantara/uikit-styles";
 
 export const { staticClasses, useClasses } = createClasses("HvFile", {
   root: {},
-  progressbar: {
+  progressBar: {
```
MEsteves22, line 6: I would leave the class names as they were because changing them will lead to breaking changes if someone is using them.

MEsteves22 (commented):
👌🏻

zettca (commented):
The "loading bar" color seem wrong in Wicked

image

zettca (approved):
LGTM 👍

packages/core/src/components/FileUploader/File/File.tsx
```diff
@@ -138,15 +140,17 @@ export const HvFile = ({
...
-        <progress
-          className={classes.progressbar}
-          aria-valuemin={0}
-          aria-valuemax={100}
-          aria-valuenow={currentProgress}
-          style={{ width: `${currentProgress}%` }}
+        <HvProgressBar
+          classes={{
+            root: classes.progressbar,
+            progressBarContainer: cx(
```
HQFOX, line 147: This was the less confusing way I found to keep the old classes and minimize breaking changes

MEsteves22 (approved):

PR #3663, opened 2023-09-04
docs: Input, DatePicker, GlobalActions review

zettca:
- consolidate `HvInput` samples
- move sample container styles to decorators
- fix `HvDatePicker` missing decorators
- change `styled` samples to use `css` instead
- rename `HvDatePicker` morbid sample 💀 
- update deprecated input controls API

### DIFF
--- a/packages/core/src/components/DatePicker/DatePicker.stories.tsx
+++ b/packages/core/src/components/DatePicker/DatePicker.stories.tsx
@@ -1,11 +1,11 @@
 import { useEffect, useState } from "react";
 
-import styled from "@emotion/styled";
+import { Global } from "@emotion/react";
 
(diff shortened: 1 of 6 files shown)


packages/core/src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -20,14 +20,40 @@ import {
   theme,
 } from "@hitachivantara/uikit-react-core";
 
-const Decorator = ({ children }) => {
-  return <div style={{ width: 340, height: 600, padding: 10 }}>{children}</div>;
-};
+const overflowDecorator: DecoratorFn = (Story) => (
```
zettca, line 23: We probably should find a place to put some "story utilities", that's easily accessible from all packages

HQFOX (approved):

PR #3667, opened 2023-09-05
docs: swr for data fetching in templates

MEsteves22:


packages/core/src/components/Table/stories/TableSamples/LoadingContainer.tsx
```diff
@@ -10,50 +8,32 @@ import { hexToRgbA } from "@core/utils/hexToRgbA";
...
-      const { clientHeight, clientWidth, offsetTop, offsetLeft } = ref.current;
-      setOverlayPosition({
-        top: offsetTop,
-        left: offsetLeft,
-        height: clientHeight,
-        width: clientWidth,
-      });
-    }
-  }, [children]);
+  const { ref } = useResizeDetector();
```
zettca, line 14: Probably much better than the `useEffect`, but JS for a simple container overlay seems overkill - this can easily be done with just CSS

MEsteves22: Yup, I was actually updating this at the moment to remove this `useResizeDetector`

MEsteves22:
Updated with review.

Applitools never seems to run with the changes in this PR... It's weird because the templates are not even tested.

MEsteves22:
Applitools is now running ✅ I could use `mdx` files for some reason. Switched back to `stories.tsx` files.

zettca (approved):
Looks great 👌

francisco-guilherme (approved):

PR #3668, opened 2023-09-05
fix(Theme): link shadow constant to theme header shadow property

loonatic:
DS3 theme has on the header-shadow property the shadow constant set while that same property on DS5 is set to "none" causing the header to not have a shadow as defined on DS mocks.


MEsteves22 (approved):

zettca (approved):
Thanks for the contribution 💪

PR #3673, opened 2023-09-12
docs(AssetInventory): refactor template

plagoa:


templates/AssetInventory/CardView/CardView.tsx
```diff
@@ -69,6 +69,12 @@ export const CardView = ({ id, instance }: CarViewProps) => {
                 </HvCardContent>
               </div>
             </div>
+            <HvCardMedia
+              component="img"
+              alt="Compressor"
```
MEsteves22, line 74: The `alt` attribute is always "Compressor". I think we should get this value from `row.original` in order to have an `alt` value that changes with the image.

MEsteves22 (commented):
LGTM 🙏 

I just noticed the following:

zettca (approved):

PR #3676, opened 2023-09-13
docs(storybook): upgrade to storybook@7

zettca:
- upgrade storybook dependencies to v7
  - make necessary API changes
- simplify CSS styles overrides


MEsteves22 (commented):
LGTM 👌🏻

I just noticed the following:

- The `templateDecorator` used on the templates is no longer working since the class names changed. This was added to remove the message under the samples when no args table is displayed. We should update the decorator to keep this behaviour.
- We should increase the sample height for the drawer because it's too small now since the drawer is rendered inside the sample container. 
- I noticed the subcomponents were all removed from the stories. Shouldn't we keep them to have their API listed in the docs? 
- The table stories should be under the visualisations tab. This was updated a few weeks ago when we noticed their weren't being tested by applytools and pa11y. I also have the impression that some samples are missing. Also, the look at the hooks and renderers page, the text is not the right one. We previously had detailed information about the hooks and renderers.

packages/core/src/components/Table/stories/TableRenderers/TableRenderers.stories.tsx
```diff
@@ -30,6 +30,12 @@ import {
 
 import { makeRenderersData, NewRendererEntry } from "../storiesUtils";
 
+export default {
```
MEsteves22, line 33: Do we need this? We already have `TableRenderers.stories.mdx`. 
This is creating a new "Samples" page inside the "Table Renderers" folder repeating the samples we already have inside the "Renderers" folder.

packages/core/src/components/Table/stories/TableSamples/TableSamples.stories.tsx
```diff
@@ -28,8 +28,14 @@ import {
 import { Delete, Lock, Unlock } from "@hitachivantara/uikit-react-icons";
 import { makeData, getColumns, AssetEvent } from "../storiesUtils";
 
-import { TableComplete } from "./TableComplete.stories";
-import TableCompleteRaw from "./TableComplete.stories?raw";
+import { TableComplete } from "./TableCompleteSample";
```
MEsteves22, line 31: Some information and table samples are missing:

- The main table sample is missing information: "A table gathers relational data. It displays..." 
- The following samples are missing from the main table page: no data, simple table, grouped rows, responsive table, list row.
- The complete table is missing: https://lumada-design.github.io/uikit/master/?path=/docs/visualizations-table--complete-story
- We have a "Table Samples" page where we have repeated samples included in the table hooks: Alternative Layout Story, Column Resize, etc

packages/core/src/components/Table/stories/TableRenderers/TableRenderers.stories.tsx
zettca: `.stories.tsx` must have a meta default export with meta 😕
removing `.stories` from name

packages/core/src/components/Drawer/Drawer.stories.tsx
```diff
@@ -17,28 +14,19 @@ import {
 const meta: Meta<typeof HvDrawer> = {
   title: "Widgets/Drawer",
   component: HvDrawer,
+  decorators: [(Story) => <div style={{ minHeight: 600 }}>{Story()}</div>],
 };
 export default meta;
 
 export const Main: StoryObj<HvDrawerProps> = {
```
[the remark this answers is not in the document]
zettca: Fixed

packages/core/src/components/Table/stories/TableHooks/TableHooks.stories.tsx
```diff
@@ -41,6 +41,12 @@ import {
   AssetEvent,
 } from "../storiesUtils";
 
+export default {
```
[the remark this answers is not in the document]
zettca: removed the extension

MEsteves22 (approved):
🚀🎉

PR #3680, opened 2023-09-14
docs(DetailsView): review template

zettca:
- Replace KPI `HvCard`'s with `MetadataItem` (slimmer border/padding) according to DS template
- Replace Kpi1/2/3 with `MetadataItem` usages - they're not that different
- Modify data entries a bit
- Add Suspense
- Add "Edit mode" for a few entries. Doesn't update the "server data"


plagoa (approved):

MEsteves22 (commented):
LGTM 🙏

I just noticed that in "edit mode" the tags inside the tags input are hidden.

zettca:
> LGTM 🙏
> 
> I just noticed that in "edit mode" the tags inside the tags input are hidden.

It's an issue with TagsInput 😞 
I've opened a follow-up issue

MEsteves22 (approved):

PR #3682, opened 2023-09-14
fix(GlobalActions): empty space same as container

zettca:
`GlobalActions` grows according to the content, but its `::before` element which pushes the content down doesn't (has fixed width)

Example:
image

### DIFF
--- a/packages/core/src/components/GlobalActions/GlobalActions.styles.tsx
+++ b/packages/core/src/components/GlobalActions/GlobalActions.styles.tsx
@@ -17,16 +17,17 @@ export const { staticClasses, useClasses } = createClasses("HvGlobalActions", {
     zIndex: `calc(${theme.zIndices.banner} - 2)`,
     top: 0,
     left: 0,
+    padding: theme.spacing(1, 0),
+    backdropFilter: "blur(1px)",
+
     "&:before": {
       content: "''",
       display: "flex",
-      width: "100%",
-      height: 72,
-      top: 0,
+      position: "absolute",
+      inset: 0,
       background: theme.colors.atmo2,
       opacity: "75%",
     },
-    backdropFilter: "blur(1px)",
   },
   wrapper: {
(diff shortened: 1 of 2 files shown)


plagoa (approved):

MEsteves22:
I noticed the following when looking at the layout page:

Screenshot 2023-09-14 at 17 48 47

The `::before` component is on top of the content (without scrolling) when it previously pushed the content.

packages/core/src/components/GlobalActions/GlobalActions.styles.tsx
```diff
@@ -20,9 +20,8 @@ export const { staticClasses, useClasses } = createClasses("HvGlobalActions", {
     "&:before": {
```
MEsteves22, line 20: We no longer see the `::before` pseudo element behind the main component to blur the background like this: 

https://github.com/lumada-design/hv-uikit-react/assets/43220251/57b92785-3e29-4a3c-b246-a6a2f80c8e00


I believe something like this would fix it (I removed the `marginTop` from `globalWrapperComplement` and changed `global`):

```
global: {
    zIndex: `calc(${theme.zIndices.banner} - 2)`,
    top: 0,
    left: 0,

    paddingTop: theme.space.xs,
    paddingBottom: theme.space.xs,

    "&:before": {
      content: "''",
      display: "flex",
      width: "100%",
      background: theme.colors.atmo2,
      opacity: "75%",
      
      height: "100%",
      top: 0,
      position: "absolute",
    },
    backdropFilter: "blur(1px)",
  },
  globalWrapperComplement: {
    position: "relative",
    top: 0,
    left: 0,
    background: theme.colors.atmo1,
    width: "100%",
  },
```

zettca: I've changed it to `padding` in `global` 👍

MEsteves22 (approved):
🚀

PR #3683, opened 2023-09-15
fix(TagsInput): fix incorrect tags layout

plagoa:


zettca (commented):
What do you think about reducing this min-width to something like `60px`? `100px` seems overkill

image


Also, we could make the Input's text size match the labels (ie. smaller) - that way it'd match better and we'd be able to fit more text in the same space

image

plagoa:
> What do you think about reducing this min-width to something like `60px`? `100px` seems overkill and we'll keep the issue in the template
> 
> image
> 
> Also, we could make the Input's text size match the labels (`12px`/`caption1`) - that way it'd match better and we'd be able to fit more text in the same space
> 
> image

yeah, I think both make sense. I'll give it a try.

plagoa:
@zettca updated the PR with your comments.

zettca (approved):
Looks great 👌 

image

PR #3687, opened 2023-09-18
fix(Tag): a11y review, styling, disabled behaviour

zettca:
- replace delete button with icon only
  - Tags are focusable if there's `onClick` or `onDelete`
  - `onDelete` is triggered with keyboard Delete key
  - `onClick` is triggered with Enter
- improve tests
- stying & disabled behaviour review
- improve docs decorators


packages/core/src/components/Tag/Tag.tsx
```diff
@@ -84,82 +85,51 @@ export const HvTag = (props: HvTagProps) => {
...
-          color={disabled ? disabledSemanticColor : "base_dark"}
-        />
-      </HvButton>
-    );
-  };
-
-  const inlineStyle = {
-    ...style,
-  };
+  const defaultDeleteIcon = (
```
MEsteves22, line 94: We lost the hover effect on the delete button.

MEsteves22 (approved):
LGTM 🙏 

I'm just a little bit concerned about the potential breaking changes mainly because `deleteButtonProps` sightly changed. However, I believe this should not me too impactful since the new prop type should cover most attributes used by users.

zettca:
> LGTM 🙏
> 
> I'm just a little bit concerned about the potential breaking changes mainly because `deleteButtonProps` sightly changed. However, I believe this should not me too impactful since the new prop type should cover most attributes used by users.

Yes, I understand. That's why we put the issue on hold.
Initially, I changed `deleteButtonProps` to be anything (`Record<string, any>`); but I think it's preferable to not lie and pass the correct type. In the small chance a user is using a prop that exists on `button`/`HvButton` but not on `div` - they'll be informed of the issue, instead of it silently behaving incorrectly

plagoa (approved):

PR #3690, opened 2023-09-18
fix(Dropdown): role review & alignment

zettca:
Change `HvList` roles to be `list`+`listem` when selectors are enabled, avoiding nested interactible elements
Also aligns `HvList` items (had a random`5px` padding 🤷 )

### DIFF
--- a/packages/core/src/components/Dropdown/List/List.tsx
+++ b/packages/core/src/components/Dropdown/List/List.tsx
@@ -342,7 +342,8 @@ export const HvDropdownList = ({
                     maxHeight ??
                     `calc(${height}px - 32px - ${theme.space.xs} - ${theme.space.sm})`,
                   overflow: "auto",
-                  padding: 5,
+                  padding: 4,
+                  margin: -4,
                 }),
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Dropdown/List/List.tsx
```diff
@@ -342,7 +342,6 @@ export const HvDropdownList = ({
                     maxHeight ??
                     `calc(${height}px - 32px - ${theme.space.xs} - ${theme.space.sm})`,
                   overflow: "auto",
-                  padding: 5,
```
MEsteves22, line 345: If we remove the padding, the outline added when an item is focused is cut out. We could add a negative margin to fix the problem with the outline as well as the wrong alignment of the list: 

```
 padding: 5,
 margin: -5,
```

zettca: That's nice - nice catch 👍

Added it as `4px`, which is the size of our border
We probably should have this as an utility/var and use it elsewhere - it's a problem I've seen quite a few times

PR #3693, opened 2023-09-18
fix(VerticalNavigation): fix invalid 'aria-controls' attribute value

loonatic:
When the VerticalNavigation component is collapsed, the items with children would still have reference to an ID of the hvtreeview which doesn't exist since in this mode a popup is used instead.

image

`Invalid ARIA attribute value: aria-controls="hvtreeview3-8-group"`


zettca (approved):
Thanks for this contribution 💪

PR #3701, opened 2023-09-20
feat(defaultProps): replace theme variables with classes

HQFOX:
I have replaced most of the theme component variables with classes to the `defaultProps`.
Some values I wasn't able to remove, these were mostly values used with the `activeTheme` ( mostly typography variants ).
Other values I left because they seemed useful to customize the component and provided some abstraction, these were mostly values that are used throughout the component in different classes or sub components. 
See if you agree with the ones I left out or the ones I left in.


MEsteves22 (commented):
Looks good 🙏 

I just identified some things while reviewing. 

Also, I noticed some components have two distinct areas:
- (1) one where the props are defined and used by `useDefaultProps`:

```
components: { "HvCard": { (...) }, (...) }
```

- (2) and one where we have some more styling and/or variants defined (and these values are mapped out to CSS vars):

```
card: {
     titleVariant: "label",
     subheaderVariant: "caption1",
     subheaderColor: theme.colors.secondary,
}
```

I believe there are still some variables in (2) that could be mapped out to (1). I identified those in my review. Also, when necessary, I also think we should map out the variables related to CSS styling by creating new classes. For example, regarding the `card` component, we could create a new class for the `subheaderColor` to remove it from (2). The only values related to CSS styling I see remaining in (2) are the `header` height and second level height because these seem like variables users would want to access.

Also, having variants as CSS vars is something I believe we should try to remove in this PR since these are not valid values (ex: `titleVariant` and `subheaderVariant` for the card component). Thus, I think we should review our mapping to omit them.

What of you think?

packages/core/src/components/Header/Brand/Brand.styles.tsx
```diff
@@ -10,4 +10,5 @@ export const { staticClasses, useClasses } = createClasses("HvHeader-Brand", {
     margin: theme.space.xs,
     backgroundColor: theme.colors.secondary,
   },
+  brandName: {},
```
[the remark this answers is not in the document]
HQFOX: that color is already being applied by default so I remove it.

packages/core/src/components/Drawer/Drawer.tsx
```diff
@@ -94,39 +84,14 @@ export const HvDrawer = (props: HvDrawerProps) => {
...
 
   const closeButtonDisplay = () => <Close role="none" />;
 
   const CloseButtonTooltipWrapper = buttonTitle
     ? withTooltip(closeButtonDisplay, buttonTitle, "top")
     : closeButtonDisplay;
 
-  const [backgroundColor, setBackgroundColor] = useState(
-    getBackgroundColor(backgroundColorValue)
-  );
```
zettca, line 112: We don't need this anymore?

packages/core/src/components/Card/Card.tsx
```diff
@@ -61,8 +61,7 @@ export const HvCard = (props: HvCardProps) => {
         "HvIsCardGridElement",
         classes.root,
         css({
-          backgroundColor:
-            (bgcolor && theme.colors[bgcolor]) || theme.card.backgroundColor,
+          backgroundColor: getColor(bgcolor),
```
zettca, line 64: I think we should include these improvements & missing `useDefaultProps` in a separate PR, to reduce the scope & changes of this huge PR

packages/core/src/components/Card/Header/Header.tsx
```diff
@@ -27,17 +27,17 @@ export interface HvCardHeaderProps
...
+export const HvCardHeader = (props: HvCardHeaderProps) => {
+  const {
+    classes: classesProp,
+    className,
+    title,
+    subheader,
+    icon,
+    onClick,
+    ...others
+  } = useDefaultProps("HvCardHeader", props);
```
zettca, line 39: I think we should include these improvements & missing `useDefaultProps` in a separate PR, to reduce the scope & changes of this huge PR

zettca (commented):
What a huge effort 😅 💪 

Looking great overall. I still want to make a deeper review, but I'll leave these comments for now:

- I think we should separate the component improvements & missing `useDefaultProps` to a separate PR, to reduce the scope and size of this PR
- It would be nice if we could extract more common color keys like we did for `containerBackgroundHover`

packages/core/src/components/Drawer/Drawer.tsx
HQFOX: The drawer backdrop color isn't working probably for a while, so it wasn't making a difference, I tried for a few minutes to fix it but decided it's probably best to open a separate issue

packages/core/src/components/Card/Header/Header.tsx
HQFOX: In this case it was needed to remove on `titleVariant` theme variable

packages/core/src/components/Drawer/Drawer.tsx
zettca: agreed 👍

packages/core/src/components/Card/Header/Header.tsx
zettca: I think we should definitely expose `HvCardHeader` - it's part of our API. I missed it 😞

MEsteves22 (commented):
Just noticed two small things👌🏻

MEsteves22 (approved):
LGTM 🙏

We'll need to review the "White Labelling" page because a lot of the variables used there are no longer available. But I think we can do that in another PR.

As future work I also think that we should try to remove variants as CSS vars since these are not valid values.

packages/styles/src/themes/ds3.ts
```diff
@@ -228,291 +229,1107 @@ const ds3 = makeTheme((theme) => ({
...
-    borderRadius: theme.radii.base,
-    padding: theme.spacing("0", "xs"),
-    marginIconRight: "0px",
-    marginIconLeft: "-8px",
-    semanticColor: "rgba(251, 252, 252, 0.3)",
-    semanticColorDisabled: "rgba(251, 252, 252, 0.1)",
-    hoverColor: theme.colors.atmo3,
-    secondaryBackgroundColor: theme.colors.atmo1,
-    secondarySubtleBorderColor: theme.colors.atmo4,
   },
```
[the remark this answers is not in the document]
HQFOX: wouldn't move the type definition to here remove completions if you create a new theme?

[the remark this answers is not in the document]
zettca: This wouldn't be "defining" the types, but merely making it more "strict" for the `CSSProperties` completions.
But I don't oppose using `CSSProperties` on the other side (although we might lose some `CSSInterpolation` "features")

zettca (approved):
Awesome 👌

PR #3706, opened 2023-09-22
docs: document title fixed

MEsteves22:
- This [PR](https://github.com/lumada-design/hv-uikit-react/pull/3702) was not able to fix the following problem:

Screenshot 2023-09-22 at 14 42 45

- In this PR the title is changed directly on the `dist` folder by running a script after building the docs. I saw Fluent UI using this method to update their title. Unfortunately, from what I've seen, Storybook doesn't have a better way of doing this.

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -1,7 +1,4 @@
 <!-- Storybook override -->
-<script>
-  document.title = "NEXT UI Kit";
-</script>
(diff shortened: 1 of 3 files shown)


zettca (approved):
If I knew it would be this complicated, I probably wouldn't haven't suggesting fixing this 😅 

Linking to https://github.com/storybookjs/storybook/issues/22211 for context

PR #3707, opened 2023-09-22
feat(Flow): add Node component

plagoa:


packages/lab/src/components/Flow/Node/Node.tsx
```diff
@@ -0,0 +1,192 @@
...
+        className
+      )}
+    >
+      <div
+        className={cx(css({ backgroundColor: color }), classes.headerContainer)}
+      >
+        <div className={classes.titleContainer}>
+          {icon}
+          <HvTypography variant="title4" className={classes.title}>
+            {groupId}
```
MEsteves22, line 115: Instead of using the `groupId` here, I think we should use the group `label` (`nodeTypes?.[type].meta?.label`). Using an `id` seems confusing specially if someone is not careful about the ids they are providing to the groups.
Also, I think we should rename the `titleContainer` class used here. It can be confusing since we have a `title` prop that we use below. 
What do you think?

packages/lab/src/components/Flow/Node/Node.tsx
```diff
@@ -0,0 +1,192 @@
...
+            onClick={() => setShowParams((p) => !p)}
+            style={{
+              cursor: hasParams ? "pointer" : "not-allowed",
+            }}
+          >
+            <Down />
+          </HvButton>
+        </div>
+      </div>
+      <div className={classes.descriptionContainer}>
```
MEsteves22, line 135: I think we should rename this `descriptionContainer` class since this is wrapping the `title`.  It can be confusing since we have a `description` prop that we use above.

MEsteves22:
LGTM 👌🏻

I just identified some small things.

packages/lab/src/components/Flow/Node/Node.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const { classes, cx, css } = useClasses(classesProp);
+
+  const { nodeGroups, nodeTypes } = useFlowContext();
+  const edges = useStore((s) => s.edges);
+  const nodes = useStore((s) => s.getNodes());
+
+  const node = nodes.find((n) => n.id === id);
+
+  const groupId = nodeTypes?.[type].meta?.groupId;
+  const groupLabel = nodeTypes?.[type].meta?.label;
```
MEsteves22, line 74: Sorry, in my previous review I induced you in error 😅 We should get the group label by doing:

```
const groupLabel = groupId && nodeGroups && nodeGroups[groupId].label)
```

Otherwise, we'll get the node type label and that's not what we want.

packages/lab/src/components/Flow/stories/ToyStory.tsx
```diff
@@ -0,0 +1,32 @@
+import { theme } from "@hitachivantara/uikit-styles";
+import { HvFlowNode } from "../Node/Node";
+
+export const ToyStory = (props) => {
+  return (
+    <HvFlowNode
```
MEsteves22, line 6: I think we want "Toy Story" and "Toy Story Description" for the `title` and `description`, respectively. 
Also, `color` is no longer needed (same for the other nodes).

packages/lab/src/components/Flow/stories/Insights.tsx
```diff
@@ -0,0 +1,26 @@
+import { theme } from "@hitachivantara/uikit-styles";
+import { HvFlowNode } from "../Node/Node";
+
+export const Insights = (props) => {
+  return (
+    <HvFlowNode
```
MEsteves22, line 6: I think we want "Insights Dashboard" for the `title`.
Also, `color` is no longer needed (same for the other nodes).

MEsteves22 (approved):
🚀👌🏻

zettca (approved):
LGTM 🚀 

Some notes for later:

PR #3708, opened 2023-09-22
fix: layout page and links fixed in docs

MEsteves22:
- "Simple Grid" sample in the layout page fixed: breakpoints were undefined
- "Packages" removed from the "Project Status" table of contents since this section doesn't exist
- Wrong URLs fixed
- Links fixed in the docs:
   - Anchor links to navigate to a page section were opening a new iframe
   - Links linking to other storybook pages were broken and leading to "Not found" pages

### DIFF
--- a/.storybook/blocks/DocsContainer.tsx
+++ b/.storybook/blocks/DocsContainer.tsx
@@ -7,6 +7,7 @@ import { addons } from "@storybook/addons";
 import {
   HvProvider,
   HvTypography,
+  HvTypographyProps,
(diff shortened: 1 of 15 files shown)


zettca (approved):
Nice 👍 

Maybe we should consider using Storybook's `LinkTo` component.
The syntax isn't as nice as `[]()` and adds more coupling to Storybook, but it's more future-proof and the navigation is _smoother_.

But for now, this is great 👍

PR #3710, opened 2023-09-25
chore: utility classes POC with `UnoCSS`

zettca:
This PR showcases the usage of utility classes in the `/app` project using `UnoCSS` Vite plugin. The `presetMini` preset is used, and configured with our theme (minimal configuration).

The (partial) `/app` refactor commit results in a substantial reduction of code:
- 13 fewer files (removal of `styles.ts` files)
- +189 lines added, -651 lines removed


app/src/components/common/Tutorial/Tutorial.tsx
```diff
@@ -28,15 +24,7 @@ export const Tutorial = ({
...
-    <div
-      className={classes.root}
-      style={{
-        backgroundColor: hexToRgbA(
-          activeTheme?.colors.modes[selectedMode].base_dark,
-          0.5
-        ),
-      }}
-    >
+    <div className="screen overflow-hidden position-fixed inset-0 z-modal bg-base_dark:50">
```
zettca, line 27: This is quite a long class, which doesn't look great.
But these cases aren't all that common, and typically mean some kind of abstraction is in order.
In this case, an `Overlay` or even `HvOverlay` component.

app/src/pages/Instructions/Instructions.tsx
```diff
@@ -55,138 +67,76 @@ const Instructions = () => {
...
+              </div>
             }
           />
           <HvCardContent>
-            <HvTypography variant="title2" className={classes.subSection}>
+            <HvTypography variant="title2" className="mb-sm">
               Tools
             </HvTypography>
-            <HvBox className={classes.tool}>
-              <HvBox className={classes.toolTitle}>
```
zettca, line 69: The refactor of these `classes.tool` and `classes.toolTitle` leads to multiple repeated classes, which means that a [refactor is in order](https://tailwindcss.com/docs/reusing-styles#extracting-components-and-partials), hence the `Entry` component, reducing duplication

packages/core/src/components/Table/TableBody/TableBody.tsx
```diff
@@ -38,7 +38,7 @@ export interface HvTableBodyProps
 
 const tableSectionContext = {
   type: "body",
-  filterClassName: "grid",
+  filterClassName: "_grid",
```
zettca, line 41: This _internal_ `grid` class collides with a utility class

zettca:
See #3855

PR #3726, opened 2023-10-09
docs: links in read me page fixed

MEsteves22:


plagoa (approved):

zettca (commented):
Can we fix this too?

https://github.com/lumada-design/hv-uikit-react/blob/master/docs/overview/welcome/Header/Header.tsx#L27

zettca (approved):

PR #3729, opened 2023-10-11
docs: add Flow sample

plagoa:


MEsteves22 (approved):
👌🏻📈

packages/lab/package.json
```diff
@@ -45,6 +45,7 @@
     "@emotion/css": "^11.11.0",
     "@hitachivantara/uikit-react-core": "^5.28.1",
     "@hitachivantara/uikit-react-icons": "^5.6.7",
+    "@hitachivantara/uikit-react-viz": "^5.5.12",
```
zettca, line 48: We don't want this. We don't need this.
If ESLint complains in the stories files, then it's wrong and we should ignore it

zettca (approved):

package-lock.json
```diff
@@ -31660,6 +31660,7 @@
         "@emotion/css": "^11.11.0",
         "@hitachivantara/uikit-react-core": "^5.28.1",
         "@hitachivantara/uikit-react-icons": "^5.6.7",
+        "@hitachivantara/uikit-react-viz": "^5.5.12",
```
MEsteves22, line 31663: There's some leftovers of the viz package in the `package-lock`

PR #3732, opened 2023-10-12
feat(Flow): allow overriding default edge options

plagoa:


zettca (approved):
👍

packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -235,6 +236,15 @@ export const HvDroppableFlow = ({
 
   const validEdges = validateEdges(edges, nodes, nodeTypes);
 
+  const defaultEdgeOptions = {
```
MEsteves22, line 239: I don't know if I'm reading too much into this 😅 but maybe we should do something like shown below in order for the `markerEnd` values we have defined to not be overwritten if someone defines the `markerEnd` on `defaultEdgeOptions` just to change the `color` let's say:

```
 const defaultEdgeOptions = {
     markerEnd: {
       type: MarkerType.ArrowClosed,
       height: 20,
       width: 20,
       ...defaultEdgeOptionsProp?.markerEnd,
     },
     ...defaultEdgeOptionsProp,
 };
```

MEsteves22: As discussed, we'll leave this as it is since the `markerEnd` can be either a string or an object.

MEsteves22 (approved):

PR #3735, opened 2023-10-13
feat(Flow): add empty state support

plagoa:


packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -39,6 +39,8 @@ export interface HvDroppableFlowProps<
   nodes?: Node<NodeData, NodeType>[];
   /** Flow edges. */
   edges?: Edge[];
+  /** Flow empty state. */
+  emptyState?: React.ReactNode;
```
MEsteves22, line 43: We no longer need the `emptyState` prop in the droppable flow (and all the related code added).

packages/lab/src/components/Flow/Empty/Empty.tsx
```diff
@@ -0,0 +1,37 @@
+import { css } from "@emotion/css";
+import {
+  HvEmptyState,
+  HvEmptyStateProps,
+} from "@hitachivantara/uikit-react-core";
+import { Fail } from "@hitachivantara/uikit-react-icons";
+import { theme } from "@hitachivantara/uikit-styles";
+import { useStore } from "reactflow";
+
+export interface HvFlowEmptyProps {
```
MEsteves22, line 10: I think we should extend `HvEmptyState` to enable the customisation of the empty state component without any limitations (icon, classes, etc):

```
export interface HvFlowEmptyProps extends HvEmptyStateProps { }
```

and then we can do something like:

```
import { useClasses } from "./Empty.styles";

export const HvFlowEmpty = ({ className, ...others }: HvFlowEmptyProps) => {
   const { classes, css } = useClasses();
   
   const nodes = useStore((state) => state.getNodes());
   
   return (
     !nodes ||
     (nodes.length === 0 ? (
       <HvEmptyState 
            className={cx(classes.root, className)}
            {...others} 
       />
     ) : null)
   );
 };
```

```
// Empty.styles.tsx

export const { staticClasses, useClasses } = createClasses("HvFlowEmpty", {
  root: {
     width: "100%",
     height: "100%",
     alignItems: "center",
     justifyContent: "center",
     backgroundColor: theme.colors.backgroundColor,
     position: "absolute",
  },
});

``` 

I'm also using the `css` util from `useClasses` to use our emotion cache.

plagoa: Yep, that's a good idea, thanks.

plagoa:
@MEsteves22 updated the PR and the documentation.

MEsteves22 (commented):
Looks good 🙏 

We just need to remove the `emptyState` prop from the `HvDroppapleFlow` since we no longer need it.

plagoa:
> LGTM 🙏
> 
> We just need to remove the `emptyState` prop from the `HvDroppapleFlow` since we no longer need it.

Done!

MEsteves22 (approved):

zettca (approved):
Looking great 👌

Issue #3715, opened 2023-09-28
CSS import is breaking the build on Next.js


urkopineda:
### Which version of NEXT UI Kit are you opening this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

Since `@hitachivantara/uikit-react-lab@5.5.0`, importing `HvStepNavigation` is breaking the build on Next.js. The error shown when building the project seems to be related to a CSS import. The last working version with Next.js is `@hitachivantara/uikit-react-lab@5.4.50`.

```
./node_modules/reactflow/dist/style.css
Global CSS cannot be imported from within node_modules.
Read more: https://nextjs.org/docs/messages/css-npm
Location: node_modules/@hitachivantara/uikit-react-lab/dist/esm/components/Flow/DroppableFlow.js

Import trace for requested module:
./node_modules/reactflow/dist/style.css
./node_modules/@hitachivantara/uikit-react-lab/dist/esm/components/Flow/DroppableFlow.js
./node_modules/@hitachivantara/uikit-react-lab/dist/esm/components/Flow/Flow.js
./node_modules/@hitachivantara/uikit-react-lab/dist/esm/index.js
./src/components/shared/Wizard/Wizard.tsx

./node_modules/reactflow/dist/style.css
Module build failed: Error: Final loader (./node_modules/next/dist/build/webpack/loaders/error-loader.js) didn't return a Buffer or String
    at processResult (/Users/upinedaolmo/Projects/file-administrator/ch2-nas-manager-ui/node_modules/next/dist/compiled/webpack/bundle5.js:28:398847)
    at /Users/upinedaolmo/Projects/file-administrator/ch2-nas-manager-ui/node_modules/next/dist/compiled/webpack/bundle5.js:28:400317
    at /Users/upinedaolmo/Projects/file-administrator/ch2-nas-manager-ui/node_modules/next/dist/compiled/loader-runner/LoaderRunner.js:1:8727
    at iterateNormalLoaders (/Users/upinedaolmo/Projects/file-administrator/ch2-nas-manager-ui/node_modules/next/dist/compiled/loader-runner/LoaderRunner.js:1:5565)
    at iterateNormalLoaders (/Users/upinedaolmo/Projects/file-administrator/ch2-nas-manager-ui/node_modules/next/dist/compiled/loader-runner/LoaderRunner.js:1:5650)
[...]

PR #3738, opened 2023-10-16
fix(Flow): react flow styles import

MEsteves22:
Fixes: https://github.com/lumada-design/hv-uikit-react/issues/3715

- The React Flow styles from `reactflow/dist/style.css` were copied to a local file, `base.ts`, and we use this file to set the flow's global styles using `Global`:
   - I created a `postinstall` script that fetches the React Flow styles making sure the styles are updated whenever the project is installed (or when the React Flow package is upgraded for example). Let me know if you think this is unnecessary work. 
   - I tried to import the styles using `?raw` as discussed but this was leading to the following error on Next.js so I decided to copy the styles in the end since I couldn't find any other solution: 

```
Module not found: Package path ./dist/style.css?raw is not exported from package /Users/user/Desktop/hv-uikit-react/examples/uikit-next/node_modules/reactflow (see exports field in /Users/user/Desktop/hv-uikit-react/examples/uikit-next/node_modules/reactflow/package.json) 
 ```


plagoa (approved):

plagoa:
Nice touch with the post install script

MEsteves22:
Upon further reflection I decided to remove the `postinstall` script since I think this can potentially lead to unwanted behaviours: this script would run when installing the dependencies during the build process (nightly). Nonetheless, I decided to keep script to have a faster way of updating the styles. 
Let me know if you think otherwise.

zettca (approved):

zettca:
> Upon further reflection I decided to remove the `postinstall` script since I think this can potentially lead to unwanted behaviours: this script would run when installing the dependencies during the build process (nightly). Nonetheless, I decided to keep script to have a faster way of updating the styles. Let me know if you think otherwise.

Agreed. `npm install` modifying source files strange to say the least 😅

PR #3741, opened 2023-10-17
build: change build and compilerOptions target to es2021 

HQFOX:
This pr changes our build target to `ES6` witch I think is the best [fit](https://caniuse.com/?search=es6) to our advertised in our [docs](https://lumada-design.github.io/uikit/master/?path=/docs/overview-project-status--docs). 
Changing this target did not polyfill the code and many ECMAScript recent features would be left as is. To fix this I added babel to polyfill the results of the rollup. This babel addon  only runs the last stable version of `corejs` wich is `3`. This meant `ES2023` and `ES2022` features like `Object.hasOwn()` wouldn't be polyfilled. For this reason I changed the `tsconfig` `lib` to `ES2021` so that the typescript compiler would show an error when using most recent features. 

These changes result in a much higher build time from about 1 minute to about 5 in my machine.

### DIFF
--- a/.config/vite.config.js
+++ b/.config/vite.config.js
@@ -3,6 +3,7 @@ import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
 import dts from "vite-plugin-dts";
 import tsconfigPaths from "vite-tsconfig-paths";
+import { getBabelOutputPlugin } from "@rollup/plugin-babel";
 
 const pkg = require(path.resolve(process.cwd(), "package.json"));
 
@@ -26,6 +27,7 @@ export default defineConfig({
     tsconfigPaths({ loose: true }),
   ],
   build: {
+    target: "ES6",
     minify: false,
     emptyOutDir: true,
     lib: {
@@ -33,6 +35,19 @@ export default defineConfig({
       entry: path.resolve(process.cwd(), "src/index.ts"),
     },
     rollupOptions: {
+      plugins: [
+        getBabelOutputPlugin({
+          plugins: [
+            [
+              "@babel/plugin-transform-runtime",
+              {
+                usebuiltIns: "usage",
+                corejs: 3,
+              },
+            ],
+          ],
+        }),
+      ],
       output: [
         {
           format: "esm",
--- a/.config/vite.config.js
+++ b/.config/vite.config.js
@@ -3,7 +3,6 @@ import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
 import dts from "vite-plugin-dts";
 import tsconfigPaths from "vite-tsconfig-paths";
-import { getBabelOutputPlugin } from "@rollup/plugin-babel";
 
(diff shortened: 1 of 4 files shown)


.config/vite.config.js
```diff
@@ -26,13 +27,27 @@ export default defineConfig({
     tsconfigPaths({ loose: true }),
   ],
   build: {
+    target: "ES6",
```
zettca, line 30: Why `ES6`? It's pretty old (2015). We'll be getting unnecessary polyfills (larger, less efficient bundles)

Our supported browsers seem to support at least [ES2020](https://caniuse.com/sr_es11) and [ES2021](https://caniuse.com/sr_es12) (didn't exhaustively search), so it should be safe to use `ES2021` here.

And if that's the case, then there's nothing to polyfill; because we're disallowing newer features statically with TS's `lib` `ES2021`

MEsteves22: I think we should keep `ES2017` if `ES2018` is not possible because of a browser version we say we support. Changing the browsers we support seems like something that should only happen on a major release.  

I also looked at the build time and it's around 2-3 min now in my machine.

zettca: I think we need to be pragmatic here.

The problems that arose were due to usage of "bleeding-edge" (ES2022) features, not old ones. We're already not polyfilling to these older (ES2017/ES2018) versions and we haven't had issues regarding this.

By polyfilling everything back to ES2017/ES2018 we're generating extra code that:
- Takes longer to build
- Makes our bundles larger (longer downloads, loading times, etc.)
- Runtime isn't as efficient/optimised (nothing better than native, right?)

And what do we stand to gain for this? Compatibility with uncommon language features (eg. [lookbehind](https://caniuse.com/js-regexp-lookbehind)) that we don't use and probably never will? I don't think the trade-off make sense

zettca: > I also looked at the build time and it's around 2-3 min now in my machine.

Build times aren't all that relevant, compared to the generated user code. I looked at the `ES6` transpiled code for the `HvButton`, and it almost doubled - with unnecessary polyfills, as browsers already supported it

MEsteves22: Okay, let's change it if you guys agree 👍 I don't have a strong opinion about this. I just wasn't sure about the browser compatibility change but, like you said, we are already not polyfilling to these versions.

package-lock.json
```diff
@@ -2512,6 +2512,35 @@
         "@babel/core": "^7.0.0-0"
       }
     },
+    "node_modules/@babel/plugin-transform-runtime": {
```
MEsteves22, line 2515: We just need to clean up the `package-lock` since we no longer need these changes.

MEsteves22 (approved):

MEsteves22:
Can we change the title of the PR since we are no longer transpiling to ES2017?

zettca (approved):
👌

PR #3742, opened 2023-10-17
chore: add aria-label to header menu items

loonatic:
As of [PR](https://github.com/lumada-design/hv-uikit-react/pull/3184), menu items are duplicated due to styling issues but the duplicated ones are hidden from screen readers and do not appear in the accessibility tree. However, playwright+chrome is detecting both causing the locator to have the name also duplicated (like `getByRole("link", { name: "menu_name menu_name"})`.

This PR adds the aria-label attribute to the header navigation menu items so that we can find it without having to duplicate the name on the locator.


nantunes (approved):

packages/core/src/components/Header/Navigation/MenuItem/MenuItem.tsx
```diff
@@ -146,6 +146,7 @@ export const HvHeaderMenuItem = (props: HvHeaderMenuItemProps) => {
           target={itemTarget}
           {...itemProps}
           aria-current={isCurrent}
+          aria-label={item.label}
```
zettca, line 149: This label is already "visible" in the children, so this seems redundant 🤔

Looking at either Chrome DevTools or https://testing-playground.com using `getByRole("link", {name})` the label doesn't seem to be duplicated. Maybe this is a bug in Playwright...

Can we add a unit test validating this works as expected? And a small inline comment explaining why this is needed, so that it's not removed for its _apparent_ redundancy?

loonatic: Added the comment to provide more context. 
As for the unit test, I'm not entirely sure what it should test with the adding of the aria attribute. Any recommendation?

zettca: A `getByRole("link", { name: "menu_name"})` like you mentioned would be more than enough 👌

loonatic: Done 👍

zettca (approved):

PR #3744, opened 2023-10-17
feat(Flow): data added to flow node metadata

MEsteves22:
In this PR, the `data` prop was added to the `HvFlowNode` metadata. This prop is then used to set the initial node `data` when the node is added to the flow.


zettca (approved):
👍

plagoa (approved):

PR #3747, opened 2023-10-18
fix: reduce specifity on baseline styles

plagoa:


zettca (approved):
We should have a test for this

MEsteves22 (approved):

PR #3750, opened 2023-10-18
feat(Section): add Section component

plagoa:


packages/lab/src/components/Section/Section.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+        { id: "2", title: "Section 2", content: "Section 2 content." },
+        { id: "3", title: "Section 3", content: "Section 3 content." },
+        { id: "4", title: "Section 4", content: "Section 4 content." },
+        { id: "5", title: "Section 5", content: "Section 5 content." },
+        { id: "6", title: "Section 6", content: "Section 6 content." },
+      ],
+      []
+    );
+
+    useEffect(() => {
```
MEsteves22, line 151: Nothing major but I think we should remove this since since it's doing unnecessary work in the sample.

packages/lab/src/components/Section/Section.test.tsx
```diff
@@ -0,0 +1,114 @@
+import { fireEvent, render } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import { HvSection } from "./Section";
+
+describe("StepNavigation", () => {
```
MEsteves22, line 5: Wrong component name here. It should be "Section".

packages/lab/src/components/Section/Section.tsx
```diff
@@ -0,0 +1,127 @@
...
+  useControlled,
+  HvBaseProps,
+  ExtractNames,
+  HvButtonProps,
+} from "@hitachivantara/uikit-react-core";
+import { Down, Up } from "@hitachivantara/uikit-react-icons";
+
+import { staticClasses, useClasses } from "./Section.styles";
+
+export { staticClasses as emptyStateClasses };
```
MEsteves22, line 17: Wrong static classes name: should be `sectionClasses`.

MEsteves22:
At the moment the component is in the lab package. Shouldn't we add it to the core package as a widget?

packages/core/src/components/Section/Section.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const [isOpen, setIsOpen] = useControlled(expanded, Boolean(defaultExpanded));
+
+  const actsVisible = actions?.slice(0, maxVisibleActions);
+  const actsDropdown = actions?.slice(maxVisibleActions);
+
+  const showContent = expandable ? !!isOpen : true;
+
+  return (
+    <div className={cx(classes.root, className)} {...others}>
```
MEsteves22, line 75: We are missing the `id` that we extracted:

```suggestion
    <div id={id} className={cx(classes.root, className)} {...others}>
```

packages/core/src/components/Section/Section.tsx
```diff
@@ -0,0 +1,141 @@
...
+    expandButtonProps,
+    children,
+    ...others
+  } = useDefaultProps("HvSection", props);
+
+  const { classes, cx } = useClasses(classesProp);
+
+  const [isOpen, setIsOpen] = useControlled(expanded, Boolean(defaultExpanded));
+
+  const contentId = setId(id, "content");
```
MEsteves22, line 72: I think we should use the `useUniqueId` here. `setId` sets `contentId` to `undefined` if no `id` is provided by the user.

We could do something similar to what we have in other components:

```
const elementId = useUniqueId(id, "hvSection");
const contentId = setId(elementId, "content");
```

and then use the elementId at the root:

```
<div id={elementId} className={cx(classes.root, className)} {...others}>
(...)
```

MEsteves22 (commented):
Looks good 👌🏻

The tests just need to be updated because they are failing and I noticed the following:

MEsteves22 (approved):

packages/core/src/components/Section/Section.test.tsx
```diff
@@ -0,0 +1,114 @@
+import { fireEvent, render } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import { HvSection } from "./Section";
+
+describe("Section", () => {
+  it("should contain all the steps", () => {
+    const { getByText } = render(<HvSection title="Section Title" />);
+    expect(getByText("Section Title")).toBeInTheDocument();
```
zettca, line 8: Using `screen` [is better](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen)
```suggestion
    render(<HvSection title="Section Title" />);
    expect(screen.getByText("Section Title")).toBeInTheDocument();
```

packages/core/src/components/Section/Section.tsx
```diff
@@ -0,0 +1,143 @@
...
+              aria-expanded={isOpen}
+              aria-controls={contentId}
+              aria-label={isOpen ? "Collapse" : "Expand"}
+              {...expandButtonProps}
+            >
+              {isOpen ? <Up /> : <Down />}
+            </HvButton>
+          </div>
+        )}
+        <HvTypography variant="title4">{title}</HvTypography>
```
zettca, line 101: I don't think we should assume the typography variant will always be `title4`.
Users will eventually ask for extra configuration. First the `variant`, then any typography props...
What about a sub-title?

We can eliminate the problem by simply not assuming anything.

```suggestion
        {title}
```

packages/core/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "outDir": "dist",
     // "noImplicitAny": true,
     "paths": {
-      "@core/*": ["./*"]
+      "@core/*": ["./*"],
+      "@viz/*": ["../../viz/src/*"]
```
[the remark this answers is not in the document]
MEsteves22: This was added because of [this error](https://github.com/lumada-design/hv-uikit-react/actions/runs/6562065132/job/17823583163). For some reason, `typecheck` was running locally but it was breaking in the workflow. 
We can try to remove this and see if it's break again because it didn't the first time around. 

But we have this kind of config in other packages when a package is importing from another one. For example here, the `viz` package needs to config the `core` aliases because otherwise it wouldn't build as `viz` doesn't know how to resolve the aliases.

```
"paths": {
      "@core/*": ["../../core/src/*"],
      "@viz/*": ["./*"]
}
```

It would be nice to have a better solution but adding the alias in the `tsconfig` was the solution we found at the time to make all of this work.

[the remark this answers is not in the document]
zettca: All right, thanks for the explanation 👍 

It would've been even nicer if we didn't have a `@viz/*` alias 😜

zettca (approved):
Looking good 👌 . Nice tests btw 😄 

Some minor docs comments

PR #3751, opened 2023-10-18
fix(QueryBuilder): export types and default values

HQFOX:
Should we add the HV or HVQuerybuilder prefix to there subcomponents?

### DIFF
--- a/docs/overview/migration/MigrationV5.stories.mdx
+++ b/docs/overview/migration/MigrationV5.stories.mdx
@@ -461,31 +461,44 @@ In order to use them, you'll need to update the necessary imports. Please find a
 
 In v5.x some components and types were renamed as detailed in the table below.
 
-| Type or component | v4.x                   | v5.x                       |
-| ----------------- | ---------------------- | -------------------------- |
-| Component         | HvBreadcrumb           | HvBreadCrumb               |
-| Component         | HvImageCarousel        | HvCarousel                 |
-| Type              | NavigationItemProp     | HvHeaderNavigationItemProp |
-| Type              | ListValueProp          | HvListValue                |
-| Type              | ListLabelsProp         | HvListLabels               |
-| Type              | BreadCrumbPathElement  | HvBreadCrumbPathElement    |
-| Type              | HvActionContainerProps | HvActionBarProps           |
-| Type              | ActionGeneric          | HvActionGeneric            |
-| Type              | FileUploaderProps      | HvFileUploaderProps        |
-| Type              | FileUploaderLabelsProp | HvFileUploaderLabels       |
-| Type              | FilesAddedEvent        | HvFilesAddedEvent          |
-| Type              | FilesRemovedEvent      | HvFilesRemovedEvent        |
-| Type              | File                   | HvFileData                 |
-| Type              | FileProps              | HvFileProps                |
-| Type              | PaginationLabelsProp   | HvPaginationLabels         |
-| Type              | SimpleGridProps        | HvSimpleGridProps          |
-| Type              | StackDirection         | HvStackDirection           |
-| Type              | InputSuggestion        | HvInputSuggestion          |
-| Type              | TagSuggestion          | HvTagSuggestion            |
-| Type              | HvUiKitThemeNames      | HvThemeColorMode           |
-| Type              | KnobProperty           | HvKnobProperty             |
-| Type              | MarkProperty           | HvMarkProperty             |
-| Type              | FilterValue            | HvFilterGroupValue         |
+| Type or component | v4.x                   | v5.x                             |
+| ----------------- | ---------------------- | -------------------------------- |
+| Component         | HvBreadcrumb           | HvBreadCrumb                     |
+| Component         | HvImageCarousel        | HvCarousel                       |
+| Type              | NavigationItemProp     | HvHeaderNavigationItemProp       |
+| Type              | ListValueProp          | HvListValue                      |
+| Type              | ListLabelsProp         | HvListLabels                     |
+| Type              | BreadCrumbPathElement  | HvBreadCrumbPathElement          |
+| Type              | HvActionContainerProps | HvActionBarProps                 |
+| Type              | ActionGeneric          | HvActionGeneric                  |
+| Type              | FileUploaderProps      | HvFileUploaderProps              |
+| Type              | FileUploaderLabelsProp | HvFileUploaderLabels             |
+| Type              | FilesAddedEvent        | HvFilesAddedEvent                |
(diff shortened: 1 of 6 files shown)


MEsteves22:
I think what we are missing are the types and some default values. I believe the subcomponents shouldn't be exported (they were not exported in [v4.x](https://github.com/lumada-design/hv-uikit-react/blob/v4.x/packages/core/src/QueryBuilder/index.d.ts)) since the user only needs the `HvQueryBuilder` component.  

```
export { defaultCombinators, defaultLabels, defaultOperators } from "./Context";
export * from "./types";
```

Regarding the prefix, I think we should add the `HvQuerybuilder` prefix to the types and default values (`hvQuerybuilder`) since they are too generic at the moment.

packages/core/src/components/QueryBuilder/types.ts
```diff
@@ -1,123 +1,123 @@
...
 
-export interface QueryOperator {
+export interface HvQueryBuilderQueryOperator {
   operator: string;
   label: string;
   combinators: string[];
 }
 
-interface DialogLabels {
+interface HvQueryBuilderDialogLabels {
```
zettca, line 56: no need to change this. this is internal and not being exported.

packages/core/src/components/QueryBuilder/types.ts
```diff
@@ -1,123 +1,123 @@
...
+  | HvQueryBuilderAddRemoveAction
+  | HvQueryBuilderSetCombinatorAction
+  | HvQueryBuilderSetAttributeAction
+  | HvQueryBuilderSetOperatorAction
+  | HvQueryBuilderSetValueAction;
 
-export interface AskAction {
-  actions: QueryAction[];
-  dialog: DialogLabels;
+export interface HvQueryBuilderAskAction {
```
zettca, line 118: no need to change this. this is for internal use only.

packages/core/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -33,32 +33,32 @@ export { staticClasses as queryBuilderClasses };
...
+  combinators?: HvQueryBuilderQueryCombinator[];
   /**
    * The initial query representation.
    */
-  query?: Query;
+  query?: HvQueryBuilderQuery;
   /**
    * Callback fired when query changes.
-   * @param {Query} value - the query representation.
+   * @param {HvQueryBuilderQuery} value - the query representation.
```
zettca, line 51: TS already takes care of this
```suggestion
   * @param value - the query representation.
```

packages/core/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -88,9 +88,9 @@ export const HvQueryBuilder = (props: HvQueryBuilderProps) => {
   } = useDefaultProps("HvQueryBuilder", props);
   const { classes } = useClasses(classesProp);
 
-  const [pendingAction, askAction] = useState<AskAction>();
+  const [pendingAction, askAction] = useState<HvQueryBuilderAskAction>();
```
zettca, line 91: This (and all its associated types) is for internal use only

packages/core/src/components/QueryBuilder/index.ts
```diff
@@ -1 +1,7 @@
 export * from "./QueryBuilder";
+export {
+  defaultCombinators as hvQuerybuilderDefaultCombinators,
+  defaultLabels as hvQuerybuilderDefaultLabels,
+  defaultOperators as hvQuerybuilderDefaultOperators,
```
zettca, line 5:
```suggestion
  defaultCombinators as hvQueryBuilderDefaultCombinators,
  defaultLabels as hvQueryBuilderDefaultLabels,
  defaultOperators as hvQueryBuilderDefaultOperators,
```

docs/overview/migration/MigrationV5.stories.mdx
```diff
@@ -785,3 +786,8 @@ this prop was deprecated by the underlining component from material ui and you s
 - The deprecated props `hours`, `minutes`, and `seconds` were removed.
   - Users should now use either `defaultValue` or `value` to set the time, in an uncontrolled or controlled way respectively (which follows the `HvTimePickerValue` type).
 - The `HvTimePickerValue` no longer has the `period` property, as it uses the 24-hour format for consistency
+
+### Query Builder <a id="query-builder" />
+
+- All of the types exported from this component now have the `HvQueryBuilder` prefix. (e.g. the type: `Attribute` is now `HvQueryBuilderAttribute`).
+- The default values from this component have also change to have a `hvQueryBuilder` prefix. (e.g. `defaultLabels` is now `hvQuerybuilderDefaultLabels`)
```
MEsteves22, line 793: Just a typo:

```suggestion
- The default values from this component have also changed (...)
```

docs/overview/migration/MigrationV5.stories.mdx
```diff
@@ -47,6 +47,7 @@ Here's what there is to know about this migration:
   - [Time Picker](#time-picker)
   - [Table Renderer - Expand Column](#table-renderer-expand-column)
   - [Drawer](#drawer)
+  - [Query Builder](#query-builder)
```
MEsteves22, line 50: Regarding the migration guide I think we should also add the renamed types and defaults values in the "Renamed components and types" table to maintain consistency 🙏

packages/core/src/components/QueryBuilder/index.ts
```diff
@@ -1 +1,7 @@
 export * from "./QueryBuilder";
+export {
+  defaultCombinators as hvQueryBuilderDefaultCombinators,
+  defaultLabels as hvQueryBuilderDefaultLabels,
+  defaultOperators as hvQueryBuilderDefaultOperators,
+} from "./Context";
+export * from "./types";
```
MEsteves22, line 7: We are currently making public all the types exported in `types` but we only want to export some of them (not `QueryAction`, `AskAction`, and `ValueComponentProps`). Thus, we should do something like: 

```suggestion
export type { HvQueryBuilderAttribute, HvQueryBuilderNumericRange, HvQueryBuilderDateTimeStrings, HvQueryBuilderDateTimeRange, HvQueryBuilderQueryRuleValue, HvQueryBuilderQuery, HvQueryBuilderQueryRule, HvQueryBuilderQueryCombinator, HvQueryBuilderQueryOperator, HvQueryBuilderLabels } from "./types";
```

MEsteves22 (commented):
👌🏻

MEsteves22 (approved):

francisco-guilherme (approved):

PR #3752, opened 2023-10-19
chore: add visual test for scoped theme providers

plagoa:


.storybook/manager.ts
```diff
@@ -11,4 +11,11 @@ addons.setConfig({
   enableShortcuts: true,
   showNav: true,
   showPanel: true,
+  sidebar: {
+    filters: {
+      patters: (item) => {
+        return !item.tags.includes("test");
+      },
+    },
+  },
```
zettca, line 20: There's a built-in feature to hide specific stories: `docs: { disable: true }`
See https://github.com/lumada-design/hv-uikit-react/blob/master/docs/foundation/Icons/Usage.stories.tsx#L105

zettca: apparently it doesn't work for this case

zettca (approved):
👍

MEsteves22 (approved):

Issue #3717, opened 2023-09-29
Some components are not wrapped with forwardRef what causes warnings


Quernest:
### Which version of NEXT UI Kit are you opening this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

https://github.com/lumada-design/hv-uikit-react/blob/05db5a640d352bf7f44caffc109394fef5be37aa/packages/core/src/components/Dropdown/Dropdown.tsx#L214

It looks like the component is not wrapped in forwardRef which causes a warning:
<img width="1191" alt="Screenshot 2023-09-29 at 17 39 06" src="https://github.com/lumada-design/hv-uikit-react/assets/21086792/069e267a-0550-4944-82b5-49f2ef838bc1">

The same issue for `HvCheckbox`, `HvSlider`

### Expected behavior 🤔

Components should use `forwardRef`. There shouldn't be console warnings

francisco-guilherme:
Hi @Quernest, 
We opened an [issue](https://hv-eng.atlassian.net/) to review the usage of the `forwardRef` on all our Form Element components.

MEsteves22:
Hello @Quernest, this was fixed on version [@hitachivantara/uikit-react-core@5.31.0](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.31.0). Thank you for opening this issue!

PR #3754, opened 2023-10-19
fix: forwardRef added to form elements

MEsteves22:
Fixes: https://github.com/lumada-design/hv-uikit-react/issues/3717

`forwardRef` was added to all form elements:
- Checkbox
- Checkbox group
- Date picker
- Dropdown
- Input (already had `forwardRef`)
- Radio
- Radio group
- Slider
- Switch
- Tags input
- Text area
- Time picker
- Color picker
- Filter group
- Selection list

Base components were also updated:
- Base checkbox
- Base dropdown
- Base radio
- Base switch


zettca (approved):
I'm not completely sure which element should be getting the `ref`, but this makes sense to me 👌 

Nice work 💪

plagoa (approved):

PR #3756, opened 2023-10-20
feat(WarningText): align with DS 5.7

MEsteves22:
The warning text was aligned with DS 5.7:
- The text color was updated to `negative120` and the variant to `caption1`
- The icon container's size was updated to `24px`:
   - Using Figma I concluded that I had to use the `XS` size for the icon but change the container's size to `24px` because [this icon size usually uses `32px` for the container according to DS](https://designsystem.hitachivantara.com/6c705d900/p/31e481-dawn-icons/t/28a0f9). I found this a bit strange since I had to override the default specs for the icons' sizes. 
- I did not update the border and icon color since they were already using `negative`. 

I also fixed the following:
- I removed the `paddingLeft` that was unnecessary for the `warningText`
- The `paddingTop` value for `topGutter` in DS3 was updated since the text was not aligned

Before:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/5667c864-8bc8-428b-ac77-91d3aac05d6f

After:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/04c2ef79-2b3c-4b6c-8c9a-09ab7e62efbc

### DIFF
--- a/packages/core/src/components/Forms/WarningText/WarningText.tsx
+++ b/packages/core/src/components/Forms/WarningText/WarningText.tsx
@@ -67,7 +67,7 @@ export const HvWarningText = (props: HvWarningTextProps) => {
   const showWarning = localVisible && !localDisabled;
   const content = showWarning ? children : "";
   const localAdornment = adornment || (
-    <Fail className={classes.defaultIcon} color="negative" />
+    <Fail iconSize="XS" className={classes.defaultIcon} color="negative" />
   );
 
   return (
@@ -88,6 +88,7 @@ export const HvWarningText = (props: HvWarningTextProps) => {
           [classes.topGutter]: !disableGutter,
           [classes.hideText]: hideText,
         })}
+        variant="caption1"
         role="status"
         aria-live="polite"
         aria-relevant="additions text"
--- a/packages/styles/src/themes/ds3.ts
+++ b/packages/styles/src/themes/ds3.ts
@@ -421,6 +421,18 @@ const ds3 = makeTheme((theme) => ({
         },
       },
     },
+    HvBreadCrumbPage: {
+      classes: {
+        link: {
+          "&:hover": {
+            backgroundColor: theme.colors.atmo3,
+          },
+          "&:focus": {
+            backgroundColor: theme.colors.atmo3,
+          },
+        },
+      },
+    },
     HvBulkActions: {
       classes: {
         root: {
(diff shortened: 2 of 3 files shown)


MEsteves22:
This PR will break several baselines in Applitools but it's expected.

packages/core/src/components/Forms/WarningText/WarningText.tsx
```diff
@@ -67,7 +67,7 @@ export const HvWarningText = (props: HvWarningTextProps) => {
   const showWarning = localVisible && !localDisabled;
   const content = showWarning ? children : "";
   const localAdornment = adornment || (
-    <Fail className={classes.defaultIcon} color="negative" />
+    <Fail iconSize="XS" className={classes.defaultIcon} color="negative" />
```
zettca, line 70: `XS` makes it SVG icon `12px`, which looks too small. It (looks like) is `16px` in the mocks.
From my understanding, only the container changes, not the SVG size 🤔 

```suggestion
    <Fail className={classes.defaultIcon} color="negative" />
```

packages/styles/src/themes/ds3.ts
```diff
@@ -1317,6 +1317,23 @@ const ds3 = makeTheme((theme) => ({
...
+          color: theme.colors.negative,
+        },
+        topGutter: { paddingTop: "8px" },
+        defaultIcon: {
+          width: "32px",
+          height: "32px",
+          "& svg": {
+            height: "16px",
+            width: "16px",
+          },
```
zettca, line 1333: We should need this

zettca (commented):
The icon padding seems off. See below for a suggestion

image

Also, can we do `feat(WarningText):`? It's a component like any other :D

packages/core/src/components/Forms/WarningText/WarningText.tsx
MEsteves22: I will ask DS system about the size of the icon to be sure

zettca: To clarify, to me it looks like a `16px` (`iconSize: "S"`) SVG icon on a `24px` container

MEsteves22: DS responded and said the icon size is `12px`. In this case, I think I should maintain `iconSize="XS"` and the override  for the `svg` dimensions in DS3 (comment below). Do you agree or suggest an alternative?

packages/styles/src/themes/ds3.ts
```diff
@@ -1317,6 +1317,24 @@ const ds3 = makeTheme((theme) => ({
         },
       },
     },
+    HvWarningText: {
```
zettca, line 1320: Sorting these (roughly) alphabetically helps reducing diff conflicts (and those with OCD too 😅)

packages/core/src/components/Forms/WarningText/WarningText.tsx
zettca: If it's `12px`, then it's `12px` 👍  - I guess I'm bad at counting pixels

zettca (approved):
Looking good 👌

packages/styles/src/themes/ds3.ts
MEsteves22: Sorted!

plagoa (approved):

PR #3757, opened 2023-10-20
feat(InlineEditor): align with DS 5.7

MEsteves22:
- Default background color updated to `atmo1`
- Disabled state added


plagoa (approved):

packages/core/src/components/InlineEditor/InlineEditor.styles.tsx
```diff
@@ -39,12 +39,12 @@ export const { staticClasses, useClasses } = createClasses("HvInlineEditor", {
     justifyContent: "flex-start",
     alignItems: "center",
 
-    backgroundColor: "transparent",
+    backgroundColor: theme.colors.atmo1,
```
zettca, line 42: This makes it look less "inline", but it's DS's call

zettca (approved):
Thanks, I hate it 😅

PR #3759, opened 2023-10-20
refactor(CodeEditor): port to createClasses

MEsteves22:
I noticed the code editor was still using our "old styling" so I updated it.


packages/code-editor/src/components/CodeEditor/CodeEditor.stories.tsx
```diff
@@ -11,8 +17,6 @@ import {
   PopUp,
   Fullscreen,
 } from "@hitachivantara/uikit-react-icons";
-import { useState } from "react";
-import { Modal } from "@mui/material";
```
zettca, line 15: We need prettier/ESLint plugin to settle on an import order 😄

zettca (approved):
👌

packages/code-editor/src/components/CodeEditor/CodeEditor.stories.tsx
MEsteves22: Yes, I agree. I'll open a PR with this.

PR #3760, opened 2023-10-23
chore: import order configured in ESLint 

MEsteves22:
ESLint was configured to standardise the way we do imports in the codebase (they were all over the place 😅). The idea comes from this [PR](https://github.com/lumada-design/hv-uikit-react/pull/3759).
 
- The order is the following:
   - (1) - Imports from building packages
   - (2) - Imports from `react`
   - (3) - Imports from external packages
   - (4) - Internal imports with relative paths (aliases)
   - (5) - Internal import with absolute paths 

If you think another order should be implemented let me know. 

Sorry for the huge amount of files changed, the changes are all related to imports 😅

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -85,6 +85,39 @@
       }
     ],
     "import/no-cycle": "error",
+    "import/order": [
+      "error",
+      {
+        "groups": [
+          "builtin",
+          "external",
+          "internal",
+          ["parent", "sibling", "index"]
+        ],
+        "pathGroups": [
+          {
+            "pattern": "react",
+            "group": "external",
+            "position": "before"
+          },
+          {
+            "pattern": "@hitachivantara/**",
+            "group": "external"
+          },
+          {
+            "pattern": "@core/**",
+            "group": "internal"
+          },
+          {
+            "pattern": "@viz/**",
+            "group": "internal"
+          }
+        ],
+        "distinctGroup": false,
+        "pathGroupsExcludedImportTypes": ["builtin", "object"],
+        "newlines-between": "always-and-inside-groups"
+      }
+    ],
     // TODO: review this
     "react/no-array-index-key": "off",
     "no-restricted-syntax": "off", // loops are fine?
--- a/app/.eslintrc
+++ b/app/.eslintrc
@@ -1,6 +1,51 @@
 {
   "rules": {
     "no-alert": "off",
-    "no-console": "off"
+    "no-console": "off",
+    "import/order": [
+      "error",
+      {
+        "groups": [
+          "builtin",
+          "external",
+          "internal",
+          ["parent", "sibling", "index"]
+        ],
+        "pathGroups": [
+          {
+            "pattern": "react",
+            "group": "external",
+            "position": "before"
+          },
+          {
+            "pattern": "assets/**",
+            "group": "internal"
+          },
+          {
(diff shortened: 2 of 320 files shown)


plagoa (approved):

.eslintrc
```diff
@@ -85,6 +85,42 @@
       }
     ],
     "import/no-cycle": "error",
+    "import/order": [
+      "error",
+      {
+        "groups": [
+          ["external", "builtin"],
```
zettca, line 92: Why have `builtin`'s after `external`? 🤔 It's the default

I too like to see `"react"` on top, but if there's a NodeJS import in a react file, probably something's wrong - and it being on top helps identifying it

app/.eslintrc
```diff
@@ -1,6 +1,53 @@
...
+            "pattern": "types/**",
+            "group": "internal"
+          }
+        ],
+        "pathGroupsExcludedImportTypes": ["react"],
+        "newlines-between": "always",
+        "alphabetize": {
+          "order": "asc",
+          "caseInsensitive": true
+        }
```
zettca, line 49: I personally don't have a strong opinion on this, but... the fact that there are so many changes makes me think this is a bit too opinionated.

Some statistics:
- With the current config, there are `1780 errors`
- Without `alphabetize`, it'd be `831 errors`
- With `"newlines-between": "always-and-inside-groups"` (a bit more permissive), `637 errors`
- Without this block, `145 errors`

.eslintrc
MEsteves22: Just a personal preference. I'll change this.

app/.eslintrc
MEsteves22: I'll remove the `alphabetize` block to reduce the number of errors. It was just a personal preference since the code feels more "clean" with it.

app/.eslintrc
```diff
@@ -1,6 +1,50 @@
...
+            "position": "before"
+          },
+          {
+            "pattern": "assets/**",
+            "group": "internal"
+          },
+          {
+            "pattern": "components/**",
+            "group": "internal"
+          },
```
zettca, line 27: this doesn't scale well - but I'll review it in https://github.com/lumada-design/hv-uikit-react/pull/3764

examples/uikit-vite-ts/src/main.tsx
```diff
@@ -1,6 +1,8 @@
 import React from "react";
+
```
zettca, line 2: I don't understand where in the config it states that `"react"` needs to be in its own group - but ok 🤔

zettca (approved):
LGTM 👍

examples/uikit-vite-ts/src/main.tsx
MEsteves22: [From what I've read](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#distinctgroup-boolean.) this happens by default when setting the `position` for `"react"` inside `pathGroups`  and `newlines-between` is set to `always`. However, we can disable this by using [`distinctGroup`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#distinctgroup-boolean). I set `distinctGroup` to `false` in order for `"react"` to not have its own group. 
Let me know what you think.

PR #3761, opened 2023-10-23
docs(Flow): refactor samples

MEsteves22:
The Flow samples were refactored because they were confusing for users and the available code was useless sometimes:
- Each sample has its own folder to reduce confusion
- The imports were updated to import from the `lab` package instead of an absolute path (when possible)
- The samples show the code in a `raw` format: the code samples are now more legible and useful
- Comments were added with a link to find the code for the values and components that can only be imported from absolute paths

### DIFF
--- a/packages/lab/src/components/Flow/stories/Base/Dashboard.tsx
+++ b/packages/lab/src/components/Flow/stories/Base/Dashboard.tsx
@@ -1,4 +1,4 @@
-import { HvFlowNode } from "../Node/Node";
+import { HvFlowNode } from "@hitachivantara/uikit-react-lab";
 
 export const Dashboard = (props) => {
   return (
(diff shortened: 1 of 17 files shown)


MEsteves22:
The changes on Applitools are expected.

plagoa:
This looks perfectly fine for me, my only concert is that the sample `InitialState` is getting its styles from the `Base` sample, which means that if someone copies the code from the sample the flow won't show up because its missing the styles and that can lead to some confusion as everything's working but there's nothing showing.

MEsteves22:
> This looks perfectly fine for me, my only concert is that the sample `InitialState` is getting its styles from the `Base` sample, which means that if someone copies the code from the sample the flow won't show up because its missing the styles and that can lead to some confusion as everything's working but there's nothing showing.

All samples are getting `classes` from `Base` but I'll change this because of the concern you mentioned. However, the samples are still getting some values from `Base` like the `nodeTypes`, `nodeGroups`, etc. I think in these cases it makes sense to have them in the same file to avoid repeating code. I added comments with links to where they can find the code for these values and components.

MEsteves22:
Updated with the comments

plagoa (approved):

PR #3765, opened 2023-10-23
fix(DatePicker): Navigation alignment & a11y

zettca:
- Fix navigations buttons
  - alignment
  - refactor to use `HvButton` (deduplicate)
- Remove `CalendarWeekLabels`
  - the component is just a map - inlined in `SingleCalendar`
  - easier to customise `calendarDay`


MEsteves22 (approved):
Looks good!

As future work we could add a way of overriding the `aria-label` for the next and previous buttons.

plagoa (approved):

PR #3770, opened 2023-10-25
feat(facetSearch): add facet search sample

HQFOX:


packages/core/src/components/FacetSearch/FacetSearch.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default meta;
+
+export const Sample: StoryObj = {
+  parameters: {
+    docs: {
+      source: { code: FacetSearchRaw },
+      description: {
+        story:
+          "A Sample of an example implementation of the Facet Search design patterned.",
```
MEsteves22, line 17: Typo: "pattern"

packages/core/src/components/FacetSearch/FacetSearch.stories.tsx
```diff
@@ -0,0 +1,22 @@
+import { Meta, StoryObj } from "@storybook/react";
+import FacetSearchRaw from "./FacetSearch?raw";
```
MEsteves22, line 2: ESLint error: we just need a space between the first two imports

packages/core/src/components/FacetSearch/FacetSearch.stories.tsx
MEsteves22, line 17: This description is not showing on storybook. I think it's because this sample is the main story and for some reason storybook doesn't show the description 😅 Maybe using an `mdx` file like we have for some stories would be better.

Also, we have this "Nothing found" label since we don't have an args table:

Screenshot 2023-10-25 at 16 18 43

We can remove this with a decorator (example in the file `docs/templates/templateDecorator`).

packages/core/src/components/FacetSearch/FacetSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+  HvCheckBoxGroup,
+  HvCheckBox,
+  HvButton,
+  HvTypography,
+  HvInput,
+} from "@hitachivantara/uikit-react-core";
+import { css } from "@emotion/css";
+import { theme } from "@hitachivantara/uikit-styles";
+
+export const FacetSearch = () => {
```
zettca, line 13: This is a bit confusing.
We're putting the sample in the same place we put components (`components/<CompName>/<CompName>.tsx`) and even exporting it in `index.ts`

Looking at how we do things elsewhere, this should probably be under a `stories` folder (ie. `stories/FacetSearch.tsx`)

MEsteves22, line 13: I found the following behaviour with the search:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/3c4af12b-0212-4102-a4ab-fe26a0bffba0

"Coats" is not checked after cleaning the search input

packages/core/src/components/FacetSearch/FacetSearch.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { FacetSearch } from "./FacetSearch";
+
+const meta: Meta = {
```
zettca, line 5: The docs page looks a bit strange without the component props table... but I don't have any good alternative 😐 
We could make this a sample of another component, but which one? Accordion? List? CheckboxGroup?

zettca (commented):
The sample looks nice 👍

packages/core/src/components/FacetSearch/FacetSearch.stories.tsx
HQFOX: I agree, the Accordion is the one that makes the most sense to me.

MEsteves22: Yes, I think adding this sample to the Accordion page seems to make sense to me as well

packages/core/src/components/FacetSearch/FacetSearchSample.tsx
```diff
@@ -0,0 +1,99 @@
...
+          classes={{ root: css({ marginBottom: theme.spacing(1) }) }}
+        />
+        <HvCheckBoxGroup
+          id={Object.keys(data)[1]}
+          name={Object.keys(data)[0]}
+          showSelectAll
+          selectAllLabel="Select All"
+        >
+          {Object.entries(data.Closet)
+            .filter((entry) =>
```
MEsteves22, line 85: I just found some problems with the search here:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/75f2e135-8dd9-4d39-8ecb-8c1f2c877dec

I believe this will fix it:

```
{Object.entries(data.Closet).map(([word, options], index) => (
   <HvCheckBox
      label={`${word} (${options.length})`}
      value={options}
      key={index}
      className={css({
         display: word.toLowerCase().includes(searchStr.toLowerCase()) ? "flex" : "none",
       })}
   />
))}
```

MEsteves22 (commented):
👌🏻

packages/core/src/components/FacetSearch/FacetSearchSample.tsx
HQFOX: Nice catch!

docs/overview/status/data.json
```diff
@@ -310,6 +310,12 @@
       "status": "✅",
       "notes": ""
     },
+    {
+      "name": "Facet Search",
+      "features": "",
+      "status": "Supported",
+      "notes": "See the Facet Search Sample"
```
MEsteves22, line 317: I think we should just specify that the facet search sample in the accordion section: "See the Facet Search Sample located in the Accordion section" (or something similar).

MEsteves22 (approved):

francisco-guilherme (approved):

PR #3775, opened 2023-10-26
feat(DatePicker): align with DS 5.7

MEsteves22:
In this PR, the `DatePicker` component was aligned with DS 5.7:
- Only the `CalendarHeader` subcomponent needed to be aligned:
   - This component was not using our `HvInput` component so I refactored it to simply things and to add missing styles:
      - The `CalendarHeader` is exported by the `core` package so I tried to avoid breaking changes. If you notice anything thing that seems like one, let me know. In my opinion, this component needs more refactoring in another PR or major. 
- When refactoring `CalendarHeader`, I noticed `HvInput` needed some changes:
   - Alignment with DS v5.7: the border color is `secondary` 
   - The styles in DS3 were not overriding DS5 in some cases
   -  The disabled and read only variants in DS5 had the wrong styling: this seems to have changed in DS v5.2


packages/core/src/components/BaseInput/BaseInput.styles.tsx
```diff
@@ -159,17 +158,17 @@ export const { staticClasses, useClasses } = createClasses("HvBaseInput", {
...
 
-    "& $inputRootReadOnly": {
-      backgroundColor: "transparent",
+    "&$inputRootReadOnly": {
+      backgroundColor: theme.colors.atmo3,
     },
   },
   inputRootDisabled: {
-    background: theme.colors.atmo2,
+    background: theme.colors.atmo3,
```
zettca, line 166: maybe these could be set in the parent. fewer overrides, easier to change

```suggestion
    background: "none",
```

packages/core/src/components/BaseInput/BaseInput.styles.tsx
```diff
@@ -122,14 +121,14 @@ export const { staticClasses, useClasses } = createClasses("HvBaseInput", {
   inputRootInvalid: { borderColor: theme.colors.negative },
   inputRootReadOnly: {
     borderColor: theme.colors.secondary_60,
-    backgroundColor: "transparent",
+    backgroundColor: theme.colors.atmo3,
```
zettca, line 124: This looks like `atmo2` [in the spec](https://designsystem.hitachivantara.com/6c705d900/p/6700f7-input/b/50959e) 🤔

zettca (commented):
Nice refactor 💯

packages/core/src/components/BaseInput/BaseInput.styles.tsx
MEsteves22: I agree, I'll look into this!

MEsteves22: I'll revert back the background color. It will most likely be `atmo3` in the future but DS is still making some tests. Once it's official we'll change it.

packages/core/src/components/BaseInput/BaseInput.styles.tsx
MEsteves22: I'll revert back the background color. It will most likely be `atmo3` in the future but DS is still making some tests. Once it's official we'll change it.

plagoa (approved):

francisco-guilherme (approved):

PR #3776, opened 2023-10-26
refactor(flow): decoupling from digital advisor design decisions

nantunes:


packages/lab/src/components/Flow/Node/index.ts
```diff
@@ -1 +1,3 @@
 export * from "./Node";
+export * from "./ParamsNode";
+export * from "./DaNode";
```
zettca, line 3: The refactor into multiple components is nice, but I'm not sure about the naming...

We shouldn't an exported API so specific to a particular use-case that it even has its name. If it really is only suitable for that case, it should be a sample. If it's more generic than that, let's give it a more generic name.

We should have a `HvFlowBaseNode` which assumes very little (like the new `HvFlowNode`), similar to our other `Base*` components

packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -66,7 +67,7 @@ export const getNode = (nodes: Node[], nodeId: string) => {
 const validateEdge = (
   nodes: Node[],
   edge: Edge,
-  nodeTypes: HvFlowNodeTypes<string> | undefined
+  nodeMetaRegistry: Record<string, HvFlowNodeMeta>
```
zettca, line 70: maybe this registry should have its own type

packages/lab/src/components/Flow/Node/Node.tsx
```diff
@@ -88,6 +89,12 @@ export const HvFlowNode = ({
   className,
   children,
 }: HvFlowNodeProps<unknown>) => {
+  const { registerNode, unregisterNode } = useNodeMetaRegistry();
+  useEffect(() => {
+    registerNode(id, { title, inputs, outputs });
```
plagoa, line 94: I think this should be something like:
`registerNode(id, { label: title, inputs, outputs });`

since the `HvFlowNodeMeta` doesn't have a `title` prop right?

```
export interface HvFlowNodeMeta {
  label: string;
  inputs?: HvFlowNodeInput[];
  outputs?: HvFlowNodeOutput[];
}
```

packages/lab/src/components/Flow/Node/index.ts
MEsteves22: I think the `DaNode` should only exist in our samples and not as an exported component by the UI Kit since it's specific to a particular use-case.
Also, I agree that we should have a `HvFlowBaseNode` by renaming the new `HvFlowNode`.

packages/lab/src/components/Flow/Node/Node.tsx
```diff
@@ -59,63 +64,58 @@ const isInputConnected = (id, type, idx, edges) => {
       (e) => e.source === id && e.sourceHandle === idx.toString()
     );
   }
+
+  return false;
 };
 
+const defaultActions: HvFlowNodeAction[] = [
```
MEsteves22, line 71: Since these are now the default actions for the node, we can remove `defaultActions` from our stories.

PR #3781, opened 2023-10-31
fix(Drawer): add missing backdrop

plagoa:


packages/core/src/components/Drawer/Drawer.styles.tsx
```diff
@@ -10,7 +11,9 @@ export const { staticClasses, useClasses } = createClasses("HvDrawer", {
     overflow: "auto",
     boxShadow: theme.colors.shadow,
   },
-  background: {},
+  background: {
+    background: hexToRgbA(theme.colors.atmo4, 0.8),
```
zettca, line 15: This `hexToRgbA` doesn't work with CSS variables, right? No use of adding it here.

packages/core/src/components/Drawer/Drawer.tsx
```diff
@@ -81,41 +88,61 @@ export const HvDrawer = (props: HvDrawerProps) => {
...
+        >
+          <CloseButtonTooltipWrapper />
+        </HvButton>
+        {children}
+      </MuiDrawer>
+      <MuiBackdrop
+        open={!!open}
+        onClick={handleBackdropClick}
+        className={css({
+          background: hexToRgbA(colors?.atmo4 || theme.colors.atmo4, 0.8),
```
zettca, line 143: Same as above - the fallback doesn't do anything. Also, `0.8` is already the default
```suggestion
          backgroundColor: hexToRgbA(colors?.atmo4),
```

We should also add `classes.background` here - it's currently unused 🤔

packages/core/src/components/Drawer/Drawer.tsx
```diff
@@ -81,41 +88,61 @@ export const HvDrawer = (props: HvDrawerProps) => {
...
+  const { colors } = useTheme();
 
   const closeButtonDisplay = () => <Close role="none" />;
 
   const CloseButtonTooltipWrapper = buttonTitle
     ? withTooltip(closeButtonDisplay, buttonTitle, "top")
     : closeButtonDisplay;
 
+  const handleBackdropClick = useCallback(
+    (event) => {
```
zettca, line 105: Can we type the function? `event` is any like this. Alternatively, we could inline the function 😄 
The memo seems unnecessary too.

packages/core/src/components/Drawer/Drawer.styles.tsx
```diff
@@ -10,7 +11,9 @@ export const { staticClasses, useClasses } = createClasses("HvDrawer", {
     overflow: "auto",
     boxShadow: theme.colors.shadow,
   },
-  background: {},
+  background: {
```
MEsteves22, line 14: We should leave the `background` class empty since using `hexToRgbA` with a CSS var doesn't work and we already define the `background` inside the `Drawer.tsx` file.

Screenshot 2023-11-02 at 10 33 55

plagoa: Yes, these are leftovers. I've removed it.

MEsteves22 (approved):

PR #3782, opened 2023-11-02
feat(icons): align icons with ds 5.7.1

HQFOX:


plagoa (approved):

MEsteves22 (commented):
👌🏻

I just noticed the following:
- New icons that seem to be missing: Check, NewProject.
- Icons that seem to have changed: Validated, Completed. 
- RightPanel and LeftPanel are identical. Is it supposed to? I didn't find them.
- ThemeSwitcher changed to Contrast and Champion to Trophy. Should we add Contrast and Trophy?

HQFOX:
> 👌🏻
> 
> I just noticed the following:
> 
>     * New icons that seem to be missing: Check, NewProject.
> 
>     * Icons that seem to have changed: Validated, Completed.
> 
>     * RightPanel and LeftPanel are identical. Is it supposed to? I didn't find them.
> 
>     * ThemeSwitcher changed to Contrast and Champion to Trophy. Should we add Contrast and Trophy (and keep the old ones)?

The RightPanel and LeftPanel was on accident, I'm going to fix it now.
The Check and NewProject already exist they are under the CompletedStep and New (although i just noticed a slight change in the plus sign, I'm going to update it now as well ) The Validated and Completed just switch names. I have been keeping the old names to minimize the breaking changes and thought that adding duplicates with different names would cause confusion specially because they change frequently. I think that we should keep note to align the names in the next major.

MEsteves22:
> > 👌🏻
> > I just noticed the following:
> > ```
> > * New icons that seem to be missing: Check, NewProject.
> > 
> > * Icons that seem to have changed: Validated, Completed.
> > 
> > * RightPanel and LeftPanel are identical. Is it supposed to? I didn't find them.
> > 
> > * ThemeSwitcher changed to Contrast and Champion to Trophy. Should we add Contrast and Trophy (and keep the old ones)?
> > ```
> 
> The RightPanel and LeftPanel was on accident, I'm going to fix it now. The Check and NewProject already exist they are under the CompletedStep and New (although i just noticed a slight change in the plus sign, I'm going to update it now as well ) The Validated and Completed just switch names. I have been keeping the old names to minimize the breaking changes and thought that adding duplicates with different names would cause confusion specially because they change frequently. I think that we should keep note to align the names in the next major.

Okay, great! I didn't noticed that their name changed. It makes sense to me keeping the old names for now and updating everything in the next major 🙏

PR #3794, opened 2023-11-09
fix(CardHeader): icon alignment

zettca:
- fix `HvCardTitle` title-icon alignment
  - remove `position: absolute` so that text doesn't override icon
- align `HvCardTitle` with DS spec:
  - outer padding
  - default text variant (was using MUI's)


MEsteves22 (commented):
Looks good 🙏 

We just need to override the CSS (title font and spacing) for DS3 since the header is supposed to look like this:

Screenshot 2023-11-10 at 09 46 14

The font is bigger and we need more padding.

MEsteves22 (approved):

PR #3797, opened 2023-11-10
fix(BaseChart): allow for dynamic height on the BaseChart

plagoa:


zettca (approved):

MEsteves22:
The samples need to be reviewed since they are over overflowing in some cases

packages/viz/src/components/BaseChart/stories/utils.tsx
```diff
@@ -7,6 +7,7 @@ export const vizDecorator: DecoratorFn = (Story) => (
     className={css({
       backgroundColor: theme.colors.atmo1,
       padding: 20,
+      height: 300,
```
MEsteves22: Adding something like this should do the trick for most cases:

```
display: "flex",
flexDirection: "column",
```

However, we need to fix some specific charts by using an higher height than 300px.

Also, for the confusion matrix, we don't need the new values in the `vizDecorator`. By removing them, the samples are all fixed.

plagoa:
> The samples need to be reviewed since they are over overflowing in some cases

Updated to allow more space for each chart.

MEsteves22 (approved):

PR #3798, opened 2023-11-10
fix(theme): spacing

MEsteves22:
The `spacing` util was using the `space` token. This means that in DS3 this util was always using the DS5 values. Also, if someone was providing custom `space` values to the theme, the `spacing` util  was not using them. This was fixed by using the CSS vars instead.


plagoa (approved):

zettca (approved):
Nice catch 👌

PR #3808, opened 2023-11-14
chore(app): dashboard layout configuration switched to node

MEsteves22:
- Dashboard layout configuration switched to the dashboard node 
- Layout configuration removed from the dashboard preview 
- Multiple selects added to node types
- Initial flow state added to the dashboard sample
- 👉 **Flow fix:** All `useStore` usage was replaced by `useNodes` and `useEdges` because the former hook was decreasing the performance of the flow component significantly. I noticed this with the dashboard sample since we have a lot of params. The flow was lagging because `useStore` was repeatedly called.


packages/lab/src/components/Flow/stories/CustomDrop/LineChart.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { NodeProps, useStore } from "reactflow";
+import { NodeProps, useEdges, useNodes } from "reactflow";
 
 import { data } from "./data";
 
 export const LineChart = (props: NodeProps) => {
   const { id } = props;
-  const nodes = useStore((state) => state.getNodes());
-  const edges = useStore((state) => state.edges);
+  const nodes = useNodes<any>();
```
HQFOX, line 10: I think we should try to remove this `any`, but i couldn't find a quick replacement

HQFOX (approved):

plagoa (approved):

packages/lab/src/components/Flow/stories/CustomDrop/LineChart.tsx
MEsteves22: Yes, you are right. I'll take a look at this 🙏

PR #3815, opened 2023-11-15
fix(Flow): prevent calling onFlowChange on initial render

plagoa:


MEsteves22 (approved):

packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -130,6 +130,7 @@ export const HvDroppableFlow = ({
   ...others
 }: HvDroppableFlowProps) => {
   const { classes, cx } = useClasses(classesProp);
+  const [initialRender, setInitialRender] = useState(true);
```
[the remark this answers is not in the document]
plagoa: Oh perfect, I'll use that. Thanks!

packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -224,10 +225,14 @@ export const HvDroppableFlow = ({
       const nds = applyNodeChanges(changes, nodes);
       setNodes(nds);
 
-      handleFlowChange(nds, edges);
-      onNodesChangeProp?.(changes);
+      if (!initialRender.current) {
```
MEsteves22, line 228: This way, if we have an empty canvas and add one node to it, the `onFlowChange` won't be triggered. I believe setting the `ref` when the component's mount as before would solve this.

plagoa: Yeah, I'll keep the `ref` but revert back to the original solution. The `useIsMounted` hook doesn't work in what we're trying to achieve here.

packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -144,6 +145,10 @@ export const HvDroppableFlow = ({
     id: elementId,
   });
 
+  useEffect(() => {
```
MEsteves22, line 148: It doesn't seem to be working when testing the "Initial state" sample. The `ref` is set before `onNodesChange` is triggered. 

I believe one way of solving this is using the [`onInit`](https://reactflow.dev/api-reference/react-flow#on-init) callback:

```
const handleInit = useCallback(= (i) => {
   initialRender.current = false;
   onInitProp?.(i) // Extracted from props passed by the used
}, [onInitProp])

<ReactFlow
   (...)
   onInit={handleInit}
>
{children}
</ReactFlow>
```

plagoa:
@MEsteves22 @zettca as discussed I removed the "first render" logic from the component as it looked a bit sketchy and actually prevented the user from being notified of that first render. I updated the "InitialState" sample to include the behavior we were looking for but have it on the application side.

MEsteves22:
> @MEsteves22 @zettca as discussed I removed the "first render" logic from the component as it looked a bit sketchy and actually prevented the user from being notified of that first render. I updated the "InitialState" sample to include the behavior we were looking for but have it on the application side.

Great 🙏 We just need to remove the code from the droppable flow

MEsteves22 (approved):

zettca (approved):
👍

Issue #3810, opened 2023-11-15
`HvOverflowTooltip` paragraph mode doesn't work


stephaniegaspar:
### Which version of NEXT UI Kit are you opening this bug for?

None

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

When `HvOverflowTooltip` is rendered with property `paragraphOverflow: true`, the text won't truncate with ellipsis when overflows a parent component.

### Expected behavior 🤔

The property should take effect on every scenario when a text is expected to be presented with 2 or more lines and be truncated when overflows a parent component. A `Tooltip` should also be shown.

### Steps to reproduce 🕹

Enter sandbox link: 

And/or the steps to reproduce:

![image](https://github.com/lumada-design/hv-uikit-react/assets/108751625/ef613891-1bd0-457d-9793-974fde7e2924)
![image](https://github.com/lumada-design/hv-uikit-react/assets/108751625/2c5e40aa-fe09-4c37-a4cc-e6c4e35656e9)

### Your Environment 🌎

Device: [e.g. Laptop, Mobile, ....]
Browser [e.g. stock browser, safari, chrome]
Other details [e.g. screen resolution 1920x1080, ...]

stephaniegaspar:
Related work [MLAAS-2895](https://hv-eng.atlassian.net/browse/MLAAS-2895).

zettca:
Thanks for reporting this @stephaniegaspar 🙏 
We'll look into it

PR #3820, opened 2023-11-16
fix(OverflowTooltip): paragraph overflow

MEsteves22:
- `HvOverflowTooltip` fixed when `paragraphOverflow` is set to `true`: line clamp was not working and the tooltip was not showing.
- New sample added to the `HvSnackbar` docs to showcase `paragraphOverflow`.
- Since the issue was similar, I took this opportunity to refactor `HvAppSwitcher` to leverage `HvOverflowTooltip`: the tooltip was not showing on the app `title` and code was duplicated unnecessarily. 
   - I don't believe there are breaking changes. The only difference is that the `title` now uses a `div` instead of a `HvTypography` as a `span`

Addresses: https://github.com/lumada-design/hv-uikit-react/issues/3810


plagoa:
I think the use case that lead to this fix actually required a 3 line clamp. Maybe we can have this be set via property and clamped to a maximum of 3 lines. Maybe default to 2 but also accept 3?

packages/core/src/components/AppSwitcher/Action/Action.styles.tsx
```diff
@@ -41,19 +41,10 @@ export const { staticClasses, useClasses } = createClasses(
...
-
-      textAlign: "left",
-
-      overflow: "hidden",
       whiteSpace: "normal",
-      textOverflow: "ellipsis",
-      display: "-webkit-box",
-      "-webkit-line-clamp": "2",
-      "-webkit-box-orient": "vertical",
-      textWrap: "balance",
```
zettca, line 54: Can't we keep `textWrap: "balance"`?

zettca (approved):
Nice 👌

packages/core/src/components/AppSwitcher/Action/Action.styles.tsx
MEsteves22: Yup, I'll bring in back

MEsteves22:
> I think the use case that lead to this fix actually required a 3 line clamp. Maybe we can have this be set via property and clamped to a maximum of 3 lines. Maybe default to 2 but also accept 3?

Okay, I was not aware of this need. I'll add a prop to set this!

MEsteves22:
As discussed, the line clamp was updated to [3 lines as defined by DS](https://designsystem.hitachivantara.com/6c705d900/p/7761d5-notification/b/82dd7c). The CSS was overridden in the App Switcher component to only use 2 lines.

packages/core/src/components/AppSwitcher/Action/Action.styles.tsx
MEsteves22: Updated ✅

PR #3823, opened 2023-11-17
chore: vite typings & use import.meta.env

zettca:
- add vite typing references
- use import.meta.env instead of process.env
- wrap missing console warnings
  - remove unused `isValidLocale`


MEsteves22 (approved):
👌🏻

PR #3827, opened 2023-11-17
chore: add Table node to the app flow

plagoa:
Added a `Table` node to the `Flow` in the app. I think it gives it a bit more variety and is useful to illustrate scenarios with tables in dashboards. The node itself has a filter for a specific field (that part could be improved, probably) that's used to filter what is shown on the table in the dashboard.

### DIFF
--- a/app/src/pages/Flow/DashboardPreview/Renderers/Renderer.tsx
+++ b/app/src/pages/Flow/DashboardPreview/Renderers/Renderer.tsx
@@ -7,12 +7,14 @@ import {
 import { useData } from "../../utils";
 import { ChartContainer } from "./ChartContainer";
 import { Kpi } from "./Kpi";
+import { Table } from "./Table";
 
 export type DashboardContentType =
   | "line-chart"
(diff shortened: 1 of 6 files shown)


packages/core/src/components/Table/stories/TableHooks/UseHvBulkActions.tsx
```diff
@@ -46,6 +46,8 @@ export const UseHvBulkActions = () => {
     useHvBulkActions
   );
 
+  console.log(data);
```
MEsteves22, line 49: Left over to remove

MEsteves22 (approved):
Really nice 👌🏻

I just left a minor comment. 
Also, I think the table renderer and node are closely tied to the steelwheels dataset, i.e., if a new dataset is added with different data, the table renderer and node would need to be updated. However, I don't think it's a problem for now since we don't have other datasets (and I don't know if we will).

zettca (approved):
LGTM 👌

PR #3828, opened 2023-11-17
chore(Icons): improve color types & generation

zettca:
- ~add color completions~ there's issues with `dts` linking to styles packages
- reduce types size from 109kB to 38kB
- cleanup tsconfig


MEsteves22 (commented):
👌🏻

When we build the icons package we get the following error:

Screenshot 2023-11-20 at 10 26 51

For some reason the absolute path of the `styles` package is used. However, I couldn't find a quick solution 😅

zettca:
> 👌🏻
> 
> When we build the icons package we get the following error:
> 
> Screenshot 2023-11-20 at 10 26 51
> For some reason the absolute path of the `styles` package is used. However, I couldn't find a quick solution 😅

Nice catch 🙏

I tried a couple of things, but that didn't work. I've removed the `HvColorAny`

MEsteves22 (approved):

PR #3831, opened 2023-11-20
fix: button-name and heading-order a11y issues

MEsteves22:
6 a11y issue, i.e. all `button-name` and `heading-order` issues, fixed:
   - `expandParamsButtonProps` prop added to the `HvFlowNode` to be able to customise the `aria-label` (as well as other props)


zettca (approved):
👌

PR #3836, opened 2023-11-21
docs: add `GlobalActions` samples

plagoa:


packages/core/src/components/GlobalActions/GlobalActions.stories.tsx
```diff
@@ -183,3 +201,113 @@ export const SectionVariant: StoryObj<HvGlobalActionsProps> = {
     );
   },
 };
+
+export const Breadcrumb: StoryObj<HvGlobalActionsProps> = {
```
zettca, line 205: According to our guidelines:

> Avoid adding too many similar stories; each story should have a unique purpose.

How do these `Breadcrumb` and `Tabs` relate to `GlobalActions`? If the goal is to show how users can pass "custom content" to GlobalActions, then the two stories are redundant and we should merge them


And to that effect, we should probably disable Applitools too

packages/core/src/components/GlobalActions/GlobalActions.stories.tsx
```diff
@@ -183,3 +201,113 @@ export const SectionVariant: StoryObj<HvGlobalActionsProps> = {
...
+  argTypes: {
+    classes: { control: { disable: true } },
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "The `title` property can be a custom component. For example, here we use a `HvBreadCrumb` component.",
```
zettca, line 213: Not a component, element/ReactNode
```suggestion
          "The `title` property can be a custom element. For example, here we use a `HvBreadCrumb` component.",
```

packages/core/src/components/GlobalActions/GlobalActions.stories.tsx
```diff
@@ -183,3 +201,113 @@ export const SectionVariant: StoryObj<HvGlobalActionsProps> = {
...
+  },
+  render: () => {
+    const customTitle = (
+      <HvBreadCrumb
+        listRoute={data}
+        id="breadcrumb1"
+        aria-label="Breadcrumb"
+        maxVisible={5}
+      >
+        List
```
zettca, line 225: `HvBreadCrumb` doesn't use `children`

plagoa:
@zettca updated with comments.

packages/core/src/components/GlobalActions/GlobalActions.stories.tsx
```diff
@@ -183,3 +192,73 @@ export const SectionVariant: StoryObj<HvGlobalActionsProps> = {
     );
   },
 };
+
+export const CustomContent: StoryObj<HvGlobalActionsProps> = {
+  argTypes: {
```
MEsteves22, line 197: Some left overs: we should also remove `argTypes`

packages/core/src/components/GlobalActions/GlobalActions.stories.tsx
```diff
@@ -183,3 +192,73 @@ export const SectionVariant: StoryObj<HvGlobalActionsProps> = {
...
+    );
+
+    return (
+      <HvGlobalActions
+        title={customTitle}
+        classes={{
+          wrapper: classes.wrapper,
+          actions: classes.actions,
+        }}
+        backButton={
```
MEsteves22, line 242: Should we deprecate the `backButton` (with the intention of removing it in v6), like we briefly mentioned, and create a new one more generic since not only a back button can be used here?

MEsteves22 (approved):

PR #3838, opened 2023-11-22
fix(Table): mark empty header on the table with `aria-hidden` when using the `useRowSelection` hook

plagoa:


zettca (approved):
👍 

If this is approach is desirable generically speaking, it would be nice if we could do this for every empty column headers, not just useRowSelection's selectors.
For instance, in the [lockedSelection example](https://lumada-design.github.io/uikit/pr-3838/?path=/story/visualizations-table-table-hooks-usehvrowselection--locked-selection-story) we still have the a11y issue since the actions column doesn't have the `aria-hidden`

MEsteves22 (approved):
🙏

PR #3840, opened 2023-11-22
feat(Flow): input and output id and auxiliary hooks

nantunes:


packages/lab/src/components/Flow/hooks/useFlowNode.ts
```diff
@@ -1,24 +1,53 @@
-import { Edge, Node, useEdges, useNodes } from "reactflow";
+import { useMemo } from "react";
+import { Node, Edge, ReactFlowState, useStore } from "reactflow";
 
-export function useFlowNode<T = any>(id: string) {
-  const nodes = useNodes<T>();
-  const edges = useEdges();
+export function useFlowNode<T extends Node = Node>(
+  id: string
+): Node | undefined {
```
zettca, line 6:
```suggestion
 {
```
By casting the return value to `Node`, we're losing the generic "feature". TS's inference already gives us the right result

packages/lab/src/components/Flow/hooks/useFlowNode.ts
```diff
@@ -1,24 +1,53 @@
...
+  const nodeSelector = useMemo(
+    () => (state: ReactFlowState) =>
+      state.getNodes().find((n: Node) => n.id === id) as T | undefined,
+    [id]
+  );
+  return useStore<T | undefined>(nodeSelector);
```
zettca, line 12: These types are also already inferred. also, `useCallback`
same for the other hooks
```suggestion
  const nodeSelector = useCallback(
    (state: ReactFlowState) =>
      state.getNodes().find((n): n is T => n.id === id),
    [id]
  );
  return useStore(nodeSelector);
```

zettca (approved):
LGTM 👍 

Just some notes about the types

MEsteves22 (approved):

PR #3843, opened 2023-11-22
fix(Table): mark empty header on the table with `aria-hidden` when using the `useRowSelection` hook

plagoa:


MEsteves22 (approved):
🙏

packages/core/src/components/Table/hooks/useRowSelection.tsx
```diff
@@ -120,6 +120,8 @@ const DEFAULT_LABELS = {
   selectRowCheckBoxAriaLabel: "Select this row",
 };
 
+const hideHeaderVariants = ["checkbox", "actions"];
```
zettca, line 123: We can't assume the column is going to be called `"actions"` every time, and this would only work when using `rowSelection`

Regarding the checkbox column, renaming the other column to `"checkbox"` might lead to collisions in column ids on users; that's why we have a `_hv_` prefix. Also, there are apps and examples currently using the `_hv_selection` string. It's an internal name, yes - but we shouldn't just rename it without a good reason

zettca (changes requested):

packages/core/src/components/Table/hooks/useRowSelection.tsx
plagoa: @zettca not sure I understand what you're saying. I'm not renaming anything, I just started using the `variant` prop on the `column` object instead of the `id`.

zettca: Sorry, I didn't notice it was targeting `column.variant` and not `column.id` 🤦

zettca (approved):

PR #3846, opened 2023-11-24
 fix(CheckboxGroup): fix incorrect select all or none behavior

plagoa:
- This change addresses DS specification on the flow of selection of items: when only some items are selected, pressing select all should remove all selections.


MEsteves22 (approved):
👌🏻

zettca (approved):
👌

PR #3853, opened 2023-11-27
feat(QueryBuilder): opt-out of confirmation dialogs

MEsteves22:
- `disableConfirmation` prop added to the `HvQueryBuilder` to be used when the user wants to opt-out of the confirmation dialogs shown before removing rules and rule groups:
   - Defaults to `false` (current behaviour)
- "InitialQuery" sample updated to control the `disableConfirmation` prop

<s>➡️❓ Let me know if you think the prop should have another name.</s> => Updated to `disableConfirmation`

### DIFF
--- a/packages/core/src/components/QueryBuilder/QueryBuilder.tsx
+++ b/packages/core/src/components/QueryBuilder/QueryBuilder.tsx
@@ -50,6 +50,8 @@ export interface HvQueryBuilderProps {
   readOnly?: boolean;
   /** Renderers for custom attribute types. */
   renderers?: HvQueryBuilderRenderers;
+  /** Whether to opt-out of the confirmation dialogs shown before removing rules and rule groups. Default to `false`. */
+  disableConfirmation?: boolean;
   /** A Jss Object used to override or extend the styles applied. */
   classes?: HvQueryBuilderClasses;
 }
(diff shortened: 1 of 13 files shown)


packages/core/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -50,6 +50,8 @@ export interface HvQueryBuilderProps {
   readOnly?: boolean;
   /** Renderers for custom attribute types. */
   renderers?: HvQueryBuilderRenderers;
+  /** Whether to opt-out of the confirmation dialogs shown before removing rules and rule groups. Default to `false`. */
+  unrestricted?: boolean;
```
zettca, line 54: `unrestricted` feels a little generic - when I read it, it's not clear what it does

> Whether to **opt-out of the confirmation** dialogs

Inferring from the docs... what do you think about `disableConfirmation`?

zettca (approved):
LGTM 👌 

I left a naming suggestion

packages/core/src/components/QueryBuilder/QueryBuilder.tsx
MEsteves22: Yup, it seems better! Updated ✅

HQFOX (approved):

PR #3854, opened 2023-11-27
docs: add ignore rules to a11y tests

plagoa:


packages/lab/src/components/Dashboard/Dashboard.stories.tsx
```diff
@@ -84,6 +84,11 @@ export const DataDriven: StoryObj<HvDashboardProps> = {
       },
     },
     eyes: { include: false },
+    a11y: {
+      config: {
+        rules: [{ id: "aria-prohibited-attr", enabled: false }],
```
MEsteves22, line 89: Can we also add the color contrast rule (`{ id: "color-contrast", enabled: false }`)? 🙏 We don't have a11y issues with the dawn color mode but we have it with wicked.

MEsteves22 (approved):

packages/lab/src/components/Dashboard/Dashboard.stories.tsx
plagoa: Sure, PR updated.

zettca (approved):

PR #3855, opened 2023-11-28
feat(styles): add `uikit-uno-preset` preset package

zettca:
- Add `@hitachivantara/uikit-uno-preset` package
  - feel free to suggest better names for the package or main export
  - versioned as `0.x` to allow us to make any needed (breaking) changes before stabilising the API
- Based on [`@unocss/preset-uno`](https://unocss.dev/presets/uno) as a base.
  - overrides theme (instead of extending it) to set NEXT tokens/values
  - remove non-DS colors/tokens
- Use package in some `app` components for dogfooding

Currently uses `ds5` theme values to create the utility classes, for best IDE completions. `wicked` is added as a `dark` theme variant


MEsteves22 (commented):
Really nice 🙏👌🏻

I just noticed the following:

- The colors don't seem to be working with CSS vars. If we take a look at the samples on the new doc page, the styles related to color are not applied. It seems there's a problem between CSS vars (which are hex colors) that can be read by. the `rgb` util used behind the scenes. I was trying to find a solution but I didn't find one for now.
- I don't know if you guys agree, but we could add a reference to the autocomplete UnoCSS plugin in the docs. It may be helpful to users.

packages/core/src/components/Table/TableBody/TableBody.tsx
```diff
@@ -38,7 +38,7 @@ export interface HvTableBodyProps
 
 const tableSectionContext = {
   type: "body",
-  filterClassName: "grid",
+  filterClassName: "_grid",
```
MEsteves22, line 41: Do we really did this change? I couldn't understand it.

MEsteves22 (approved):
Looks good 👌🏻

Just some suggestions that can be added to the docs later:
- Add the package to the [introduction page](https://lumada-design.github.io/uikit/pr-3855/?path=/docs/overview-introduction--docs) where the packages are listed.
- We could tell the users that this package can undergo changes in the future (and breaking changes) just to make them more aware of that. 
- We could add a reference to the autocomplete UnoCSS plugin in the docs. It may be helpful to users.

packages/core/src/components/Table/TableBody/TableBody.tsx
zettca: Yes, it's an internal class that we're using in `HvFocus` - it shouldn't have such a generic/common name, as it might lead to collisions in `className`s

zettca:
> * Add the package to the [introduction page](https://lumada-design.github.io/uikit/pr-3855/?path=/docs/overview-introduction--docs) where the packages are listed.

Added 👍 

> * We could tell the users that this package can undergo changes in the future (and breaking changes) just to make them more aware of that.

That's already implied by versioning with [SemVer's `0.x`](https://semver.org/#spec-item-4), which we document in [Releases/Versioning](https://lumada-design.github.io/uikit/master/?path=/docs/overview-community-releases--docs#versioning). Isn't that enough?

> * We could add a reference to the autocomplete UnoCSS plugin in the docs. It may be helpful to users.

It's already mentioned in a small note - do you think it should be more visible?

MEsteves22:
> > * Add the package to the [introduction page](https://lumada-design.github.io/uikit/pr-3855/?path=/docs/overview-introduction--docs) where the packages are listed.
> 
> Added 👍
> 
> > * We could tell the users that this package can undergo changes in the future (and breaking changes) just to make them more aware of that.
> 
> That's already implied by versioning with [SemVer's `0.x`](https://semver.org/#spec-item-4), which we document in [Releases/Versioning](https://lumada-design.github.io/uikit/master/?path=/docs/overview-community-releases--docs#versioning). Isn't that enough?
> 
> > * We could add a reference to the autocomplete UnoCSS plugin in the docs. It may be helpful to users.
> 
> It's already mentioned in a small note - do you think it should be more visible?

Sorry, I didn't notice the small note for the plugin 😅 My bad. The note is enough for me 🙏 

Yeah, I was wondering if using the semantic version is enough but you're right since we already document this.

LGTM 👍

plagoa (approved):

PR #3856, opened 2023-11-28
docs: Add `Table` section sample

plagoa:


plagoa:
The Applitools test failure is expected as I slightly reduced the width of the `Title` column to keep things tidier.

packages/core/src/components/Table/stories/TableSamples/Section.tsx
```diff
@@ -0,0 +1,234 @@
...
+    borderBottom: `1px solid ${theme.colors.atmo3}`,
+  }),
+  sectionActions: css({
+    right: theme.space.sm,
+  }),
+  pagination: css({
+    margin: 0,
+    backgroundColor: theme.colors.atmo2,
+    padding: theme.space.xs,
+    borderRadius: `0 ${theme.radii.round} ${theme.radii.round} 0`,
```
MEsteves22, line 50: The left bottom border radius is missing. It should be: `0 0 ${theme.radii.round} ${theme.radii.round}`

MEsteves22 (commented):
Looks good 👌🏻

I just noticed the following:

packages/core/src/components/Table/stories/TableSamples/Section.tsx
plagoa: fixed

MEsteves22 (approved):

PR #3861, opened 2023-11-29
fix(Grid): item container spacing

MEsteves22:
- When using `spacing`, `rowSpacing`, or `columnSpacing` as an object, a width is set (`xs`, `sm`, etc...), and the grid is a `container` as well as an `item`, the following MUI error is shown: `MUI: Expected spacing argument to be a number or a string, got [object Object].`. It seems that this is a [MUI limitation](https://github.com/mui/material-ui/issues/29349) and this was leading to unexpected grid behaviours.
   - Since this [PR](https://github.com/lumada-design/hv-uikit-react/pull/3533), the error started to appear more frequently because, when `spacing` is set to `auto`, which is the default value, `spacing` is converted to an object (instead of a number like it previously was). I believe this was done to remove the need for `useWidth` and to avoid re-rendering the grid on width changes unnecessarily. 

This PR fixes the situation mentioned above when any spacing is set to `auto` and the grid is a `container` as well as an `item`. In this case, the `useWidth` hook is used to provide a number to spacing and not an object.

### DIFF
--- a/packages/core/src/components/CheckBox/CheckBox.stories.tsx
+++ b/packages/core/src/components/CheckBox/CheckBox.stories.tsx
@@ -162,39 +162,38 @@ export const ExternalErrorMessage: StoryObj<HvCheckBoxProps> = {
 
     return (
       <HvGrid container>
-        <HvGrid item xs={12} md={6}>
-          <HvGrid container>
-            <HvGrid item xs={12}>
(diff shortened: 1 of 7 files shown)


francisco-guilherme (approved):

zettca (approved):
👌

PR #3866, opened 2023-12-04
perf(QueryBuilder): responsive layout

MEsteves22:
Small refactor of the query builder component to make it responsive. 

Since we don't have any DS specs for this component, let me know if you think the changes regarding responsiveness should be different.


MEsteves22:
The changes on Applitools are expected.

zettca (commented):
Looking good 👍 

Can we get rid of the `HvGrid`s on the add buttons? It's overlapping the "links" and we're not even using the breakpoints

image

Maybe we could do the same for the the AND/OR combinators and the Delete query buttons - they're not leveraging the breakpoints either

MEsteves22:
@zettca Updated: I removed the unnecessary grids from the rule group component. I had to adjust the spacing since removing the grids increased it (the grids were adding a negative margin). 

I also updated the numeric value CSS because I noticed the inputs were not responsive on lower screen widths.

zettca (approved):
LGTM 👍

PR #3868, opened 2023-12-04
fix(OverflowTooltip): paragraph overflow

MEsteves22:
- v4.x workflows fixed: several workflows were pointing to master and not v4.x 
- Backports part of [this fix](https://github.com/lumada-design/hv-uikit-react/pull/3820) to v4: `HvOverflowTooltip` fixed when `paragraphOverflow` is set to true: the tooltip was not showing.
   - I didn't update the line clamp to 3 (instead 2) in this PR since I don't think it's needed. But let me know if you think otherwise.

### DIFF
--- a/.github/workflows/applitools.yml
+++ b/.github/workflows/applitools.yml
@@ -1,4 +1,5 @@
 name: Applitools
+run-name: v4.x Applitools
 on:
   workflow_dispatch:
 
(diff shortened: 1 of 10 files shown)


MEsteves22:
This PR has changes on Applitools for the controls samples but I did not update them. Maybe it's because of some previous changes that were not saved.

zettca (approved):
LGTM 👍

PR #3869, opened 2023-12-04
fix(FilterGroup): fix total number of items on the right panel not showing the correct value when searching

plagoa:


packages/core/src/components/FilterGroup/RightPanel/RightPanel.tsx
```diff
@@ -49,8 +49,11 @@ export const HvFilterGroupRightPanel = ({
   } = useContext(HvFilterGroupContext);
 
   const activeGroupOptions = useMemo(
```
MEsteves22, line 51: I noticed the following when typing "category":

https://github.com/lumada-design/hv-uikit-react/assets/43220251/87ef810b-b4ae-4081-a3f7-53b220dd65d4

I think this should solve the problem:

```
const activeGroupOptions = useMemo(
    () =>
      filterOptions[activeGroup]?.data
        .filter((option) =>
          option.name.toLowerCase().includes(searchStr.toLowerCase())
        )
        .map((option) => option.id) || [],
    [filterOptions, activeGroup, searchStr]
);
```

* I added `toLowerCase()` and removed the `trim()` since if we type "category " only 3 options are shown but the counter has 4.

packages/core/src/components/FilterGroup/RightPanel/RightPanel.tsx
```diff
@@ -49,8 +49,13 @@ export const HvFilterGroupRightPanel = ({
...
-    () => filterOptions[activeGroup]?.data.map((option) => option.id) || [],
-    [filterOptions, activeGroup]
+    () =>
+      filterOptions[activeGroup]?.data
```
MEsteves22, line 53: I just noticed another thing when selecting the "All" when there's a filter: 

https://github.com/lumada-design/hv-uikit-react/assets/43220251/9d89ef67-7ecb-42ec-b2f4-35e0fe4259ca

I think we need to update `handleSelectAll`.

plagoa: @MEsteves22 yeah, that's a weird behavior. I updated the PR, let me know what you think.

PR #3873, opened 2023-12-05
docs(TreeView): use raw code samples

zettca:


MEsteves22 (approved):
LGTM 👌🏻

➡️ We could do the same to the "Main" sample like we have for the [query builder](https://github.com/lumada-design/hv-uikit-react/blob/master/packages/core/src/components/QueryBuilder/stories/QueryBuilder.stories.tsx#L22).

PR #3883, opened 2023-12-07
fix(Drawer): fix incorrect usage of the backdrop

plagoa:
I added a story to have a `Drawer` open on top of some content, namely the `GlobalActions` that might have some styles clash with the one's on the `Drawer`. I opted to leave that be checked by Storybook because it might be important to have that tested in case anything changes.

### DIFF
--- a/packages/core/src/components/Drawer/Drawer.stories.tsx
+++ b/packages/core/src/components/Drawer/Drawer.stories.tsx
@@ -1,5 +1,7 @@
+import { useState } from "react";
 import { css } from "@emotion/css";
 import { Meta, StoryObj } from "@storybook/react";
+import { waitFor, screen, fireEvent } from "@storybook/testing-library";
 import {
   HvButton,
   HvDialogActions,
@@ -8,6 +10,7 @@ import {
   HvDrawer,
   HvDrawerProps,
   HvTypography,
+  HvGlobalActions,
 } from "@hitachivantara/uikit-react-core";
 
 const meta: Meta<typeof HvDrawer> = {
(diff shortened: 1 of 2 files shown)


zettca (commented):
Nice 👌 

There's [something wrong](https://github.com/lumada-design/hv-uikit-react/pull/3883#discussion_r1419171143) with the docs though

MEsteves22 (commented):
Looks good 🙏 

I noticed the following:

- When clicking on the button in the last sample, the page scrolls to the top automatically and the drawer doesn't open. But this happens when we click anywhere on the screen. I don't really understand what's happening but putting everything in one sample helps. For applitools, we can open the drawer programatically like we do with the dialog component. 

- `disableBackdropClick` doesn't seem to be working. Suggestion:

```
const handleOnClose: MuiDrawerProps["onClose"] = (event, reason) => {
    if (reason === "backdropClick" && disableBackdropClick) return;

    onClose?.(event, reason);
  };

  return (
    <MuiDrawer
      (...)
      slotProps={{
        backdrop: {
          (...),
          onClick: (event) => handleOnClose(event, "backdropClick"),
        },
      }}
      onClose={handleOnClose}
      {...others}
    >
      <IconButton
        id={setId(id, "close")}
        className={classes.closeButton}
        variant="secondaryGhost"
        onClick={onClose}
        title={buttonTitle}
      >
        <Close role="none" />
      </IconButton>
      {children}
    </MuiDrawer>
  );
```

plagoa:
@MEsteves22 updated the PR. I removed the second story and fixed the `disableBackdropClick` issue.

MEsteves22 (dismissed):
👌🏻

zettca (approved):

PR #3887, opened 2023-12-11
docs(Dropdown): search added to icons sample

MEsteves22:
- The "With Icons" sample was updated to enable to search a value. This was requested on our UI Kit channel since it isn't clear that the `value` or `searchValue` props are needed for the search to work when `label` is a `React.ReactNode`.
- The samples were divided into separate files to improve the code samples. Because of this, we no longer need the unnecessary `useEffect` in the "With Icons" sample.


MEsteves22:
Applitools is expected to have changes

plagoa (approved):

PR #3889, opened 2023-12-11
chore(Flow): update edges layout

plagoa:
- updated edges type to conform to DS specs
- added snap grid properties to limit position values to ints
  - we can leave it as is but it's a bit weird to have a position object to be `{ x: 90.256345, y: 120.18231 }`. this can be overridden by the user anyway


MEsteves22 (approved):
👌🏻

francisco-guilherme (approved):

PR #3892, opened 2023-12-12
feat(Flow): update default node styles

plagoa:


packages/lab/src/components/Flow/DroppableFlow.tsx
```diff
@@ -278,6 +278,9 @@ export const HvDroppableFlow = ({
           defaultEdgeOptions={defaultEdgeOptions}
           snapGrid={[1, 1]}
           snapToGrid
+          onError={(code, message) => {
+            console.error(message);
```
zettca, line 282: We should wrap this in an `if (import.meta.env.DEV) {}` block, and eslint-ignore it

packages/lab/src/components/Flow/base.ts
```diff
@@ -231,16 +232,17 @@ export const flowStyles = css`
...
     text-align: center;
-    border-width: 1px;
-    border-style: solid;
-    border-color: #1a192b;
-    background-color: white;
+    border: 1px solid ${theme.colors.negative};
+    background-color: ${theme.colors.negative_20};
+  }
+  .react-flow__node-default::before {
+    content: "Invalid node type";
```
zettca, line 244: I think that _Unknown_ instead of _Invalid_ is clearer for the user, as we're saying _why_ it's invalid.

packages/lab/src/components/Flow/DroppableFlow.tsx
plagoa: 👍

packages/lab/src/components/Flow/base.ts
plagoa: Nice, I'll update it.

plagoa:
@zettca updated PR.

zettca (approved):
👍

PR #3896, opened 2023-12-13
docs(Flow): add Invalid Flow sample

plagoa:


packages/lab/src/components/Flow/stories/Invalid/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+        >
+          Load Invalid Flow
+        </HvButton>
+      </HvGlobalActions>
+      <div className={classes.flow}>
+        <ErrorBoundary>
+          <HvFlow
+            nodeTypes={nodeTypes}
+            nodeGroups={nodeGroups}
+            onInit={onInit}
```
zettca, line 90: I guess this is fine as is, but since the default state is an empty flow that we can only make invalid, we could start with the invalid state already, making the sample simpler

My "invalid button" suggestion was to turn a valid flow into an invalid one, ie.
```tsx
// @ts-expect-error invalidate flow by removing node positions
setNodes(nodes => nodes.map({position, ...rest} => rest));
```

packages/lab/src/components/Flow/stories/Invalid/ErrorBoundary.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface Props {
+  children: React.ReactNode;
+}
+
+interface State {
+  hasError: boolean;
+  errorMessage?: string;
+}
+
+export class ErrorBoundary extends React.Component<Props, State> {
```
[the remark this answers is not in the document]
plagoa: Yeah, I thought this would make someone itch. I'll update.

packages/lab/src/components/Flow/stories/Invalid/index.tsx
plagoa: I'll update this.

plagoa:
@zettca updated the PR

packages/lab/src/components/Flow/stories/Flow.stories.tsx
```diff
@@ -160,3 +162,18 @@ export const DynamicHandles: StoryObj<HvFlowProps> = {
   },
   render: () => <DynamicHandlesStory />,
 };
+
+export const InvalidFlow: StoryObj<HvFlowProps> = {
+  parameters: {
+    docs: {
+      description: {
```
MEsteves22, line 169: We could add a description to contextualise the sample

package.json
```diff
@@ -133,5 +133,8 @@
   "workspaces": [
     "packages/*",
     "app/"
-  ]
+  ],
+  "dependencies": {
```
MEsteves22, line 137: We should add this to the list of `devDependencies` like the other packages

packages/lab/src/components/Flow/stories/Flow.stories.tsx
plagoa: Of course, I'll add.

package.json
plagoa: Missed that. I'll update that.

packages/lab/package.json
```diff
@@ -62,6 +62,7 @@
     "@types/react-dom": "^18.2.5",
     "@types/react-grid-layout": "^1.3.5",
     "npm-run-all": "^4.1.5",
+    "react-error-boundary": "^4.0.11",
```
zettca, line 65: This should be in root's `devDependencies`, as it's used for docs/stories and not the "package" itself. It's a bit strange, but it's what we're doing elsewhere

plagoa: Thanks, updated the PR

MEsteves22 (approved):

zettca (approved):

PR #3906, opened 2023-12-15
feat(Tag): add actions

plagoa:
- Applitools errors are expected as the text size in the tags changed from `caption1` (12px) to `caption2` (10px) as specified.

### DIFF
--- a/packages/core/src/Tag/Tag.styles.tsx
+++ b/packages/core/src/Tag/Tag.styles.tsx
@@ -1,19 +1,31 @@
 import { CSSProperties } from "react";
+
 import { theme } from "@hitachivantara/uikit-styles";
 
+import { chipClasses } from "@mui/material/Chip";
+
 import { outlineStyles } from "../utils/focusUtils";
+
 import { createClasses } from "../utils/classes";
 
 export const { staticClasses, useClasses } = createClasses("HvTag", {
-  root: {},
+  root: {
+    color: theme.colors.base_dark,
+
+    [`& .${chipClasses.avatar}`]: {
+      width: 12,
+      height: 12,
+      marginLeft: 2,
+      marginRight: 0,
+    },
+  },
 
   chipRoot: {
-    "&.MuiChip-root": {
+    [`&.${chipClasses.root}`]: {
       height: 16,
       borderRadius: 0,
       maxWidth: 180,
       fontFamily: theme.fontFamily.body,
-
       "&:focus-visible": {
         backgroundColor: theme.alpha("base_light", 0.3),
       },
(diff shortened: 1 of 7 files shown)


MEsteves22 (commented):
👌🏻

Just some minor comments:

plagoa:
@MEsteves22 updated the PR. I added a test to the `Tag` component tests, not the `TagsInput` component like you mentioned, seemed more appropriate.

MEsteves22 (approved):

packages/core/src/components/Tag/Tag.stories.tsx
```diff
@@ -243,3 +245,176 @@ export const TagArray: StoryObj<HvTagProps> = {
     );
   },
 };
+
+export const Selectable: StoryObj<HvTagProps> = {
```
zettca, line 249: Docs-wise, the `Selectable` & `SelectableControlled` samples show pretty much the same code, as the functions are rendered 😞

plagoa:
@zettca updated the PR with the suggested changes.

packages/core/src/Tag/stories/SelectableControlled.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <HvTag
+          label="Asset 1"
+          color="cat1"
+          selectable
+          selected={selectedTags.includes("Asset 1")}
+          onDelete={() => {
+            alert("On Delete Action");
+          }}
+          onClick={handleSelect}
+        />
```
zettca, line 39: We could create an array of `tags` with `label`/`color?`/`bgcolor`, and `map` over them to reduce duplication - (likely) similarly to what apps will be doing

We also have a `TagArray` sample that... doesn't use an array? Shouldn't it? 🤔 


```suggestion
      {tags.map(({label, color, bgColor}) => (
        <HvTag
          label={label}
          color={bgColor}
          selectable
          selected={selectedTags.includes(label)}
          className={css({ color: color ? theme.colors[color] : undefined })}
          onDelete={() => alert("On Delete Action")}
          onClick={(evt, selected) => handleSelect(evt, selected, label)}
        />
```

packages/core/src/Tag/Tag.styles.tsx
```diff
@@ -5,15 +5,23 @@ import { outlineStyles } from "../utils/focusUtils";
 import { createClasses } from "../utils/classes";
 
 export const { staticClasses, useClasses } = createClasses("HvTag", {
-  root: {},
+  root: {
+    color: theme.colors.base_dark,
+
+    "& .MuiChip-avatar": {
```
HQFOX, line 11: I think that it would be better to use the material exported classes object, something like : [& .${MuichipClasses.avatar}]

plagoa:
@zettca addressed the issues. Also removed the `TagsArray` sample as it was useless.
@HQFOX updated the styles to use the MUI classes.

zettca (approved):
👌

PR #3908, opened 2023-12-18
feat(QueryBuilder): custom operator renderer

MEsteves22:
- Value renderers are now configurable per operator:
   - Previously, if a custom renderer was passed for one of the default attribute types (`boolean`, `numeric`, `text`, `textarea`, and `dateandtime`), the renderer was not used. This was fixed since it should be possible to customise these attribute types.
   - As discussed, the `DEFAULT` key can be used in the `renderers` prop to customise the value renderers at a more broad level.  
      - ℹ️ I was not able to update the `type` and `operator` props in the `HvQueryBuilderAttribute` and `HvQueryBuilderQueryOperator` types, respectively, to accept all `string` except `DEFAULT`. I added a `console.error` in DEV mode because of this. ❓**Let me know if you know a solution for this.** 
   - Previously, when defining a custom renderer for an attribute type, the `Empty` and `IsNotEmpty` operators were overridden internally and the custom renderer was not used. This was changed since it should be possible to control these operators. Thus, as discussed, the internal logic for these operators was removed and an `EmptyValue` component was created to render `null` and clear the rule `value`.  
      - ➡️ To avoid introducing breaking changes to those that are currently relying on the previous behaviour when using  `renderers`, a `emptyRenderer` property that defaults to `["Empty", "IsNotEmpty"]` was created. This property defines which operators should use the empty value renderer when selected: to reset the rule value and render an empty component.  ❓**Let me know if you think differently.** 
- Type `HvQueryBuilderChangedQuery` fixed: `id` was not omitted inside `rules`
- `:first-child` replaced by `first-of-type` to remove the errors in the console
- Docs updated for the query builder and a new usage page was added

### DIFF
--- a/packages/core/src/QueryBuilder/QueryBuilder.tsx
+++ b/packages/core/src/QueryBuilder/QueryBuilder.tsx
@@ -53,14 +53,14 @@ export interface HvQueryBuilderProps {
    * Operators that should use the empty value renderer when selected.
    *
    * When one of the listed operators is selected, the rule value is reset and an empty component is rendered.
-   * This property takes priority to the `renderers` property.
+   * This property takes priority over `renderers`.
    *
-   * Defaults to `["Empty", "IsNotEmpty"]`.
+   * @default ["Empty", "IsNotEmpty"]
    * */
   emptyRenderer?: string[];
-  /** Custom renderers for the values. */
+  /** Custom renderers for the rules' value. */
   renderers?: HvQueryBuilderRenderers;
-  /** Whether to opt-out of the confirmation dialogs shown before removing rules and rule groups. Default to `false`. */
+  /** Whether to opt-out of the confirmation dialogs shown before removing rules and rule groups. @default false. */
   disableConfirmation?: boolean;
   /** A Jss Object used to override or extend the styles applied. */
   classes?: HvQueryBuilderClasses;
@@ -75,6 +75,8 @@ export interface HvQueryBuilderProps {
 /**
  * This component allows you to create conditions and group them using logical operators.
  * It outputs a structured set of rules which can be easily parsed to create SQL/NoSQL/whatever queries.
+ *
+ * Take a look at the [usage page](https://lumada-design.github.io/uikit/master/?path=/docs/widgets-query-builder-usage--docs) to learn more about this component.
  */
 export const HvQueryBuilder = (props: HvQueryBuilderProps) => {
   const {
--- a/packages/core/src/QueryBuilder/QueryBuilder.tsx
+++ b/packages/core/src/QueryBuilder/QueryBuilder.tsx
@@ -49,7 +49,16 @@ export interface HvQueryBuilderProps {
   labels?: HvQueryBuilderLabels;
   /** Whether the query builder is in read-only mode. */
   readOnly?: boolean;
-  /** Renderers for custom attribute types. */
+  /**
+   * Operators that should use the empty value renderer when selected.
+   *
+   * When one of the listed operators is selected, the rule value is reset and an empty component is rendered.
+   * This property takes priority to the `renderers` property.
+   *
+   * Defaults to `["Empty", "IsNotEmpty"]`.
+   * */
+  emptyRenderer?: string[];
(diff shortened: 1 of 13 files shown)


packages/core/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -48,9 +49,18 @@ export interface HvQueryBuilderProps {
...
-  /** Renderers for custom attribute types. */
+  /**
+   * Operators that should use the empty value renderer when selected.
+   *
+   * When one of the listed operators is selected, the rule value is reset and an empty component is rendered.
+   * This property takes priority over `renderers`.
+   *
+   * @default ["Empty", "IsNotEmpty"]
+   * */
+  emptyRenderer?: string[];
```
zettca, line 60: (Just a small note - I don't have a better suggestion, this is a good compromise)

In one hand, this adds an extra layer of complexity when dealing with/learning about QueryBuilder renderers, and will likely mean we'll be maintaining this in the next major versions.
But but on the other hand, it allows opting-out of the weird hard-coded `"Empty"` and `"IsNotEmpty"` values.

We also can't make nullish renderers the `EmptyValue` because that'd be a breaking change (as our default is `TextValue`)

zettca (approved):
Great work 👍 

(Given our current restrictions) the API & the docs also look great 💯 
From the extensive docs, I get the feeling that the API is a bit too complex - but given our current restrictions there's not much we can do

Now for the annoying question: Can we add a unit test on the custom attribute / attribute+value renders? 😬

MEsteves22:
> Great work 👍
> 
> (Given our current restrictions) the API & the docs also look great 💯 From the extensive docs, I get the feeling that the API is quite complex - but given our current restrictions there's not much we can do
> 
> Now for the annoying question: Can we add a unit test on the custom attribute / attribute+value renders? 😬

Yes, I'll add a unit test 🙏

packages/core/src/QueryBuilder/QueryBuilder.tsx
MEsteves22: Yes, I was also on the fence about adding or not this property but, in the end, I decided to add it mainly because of the breaking changes it would introduce not having it: users that are already using `renderers` may be expecting the "magic" behaviour we have with `Empty` and `IsNotEmpty` 😅

packages/core/src/QueryBuilder/Rule/Value/Value.tsx
```diff
@@ -19,13 +21,56 @@ export const Value = ({
...
+        renderers[attrType][operator]) ||
+      (typeof renderers[attrType] === "object" &&
+        renderers[attrType][defaultRendererKey]) ||
+      (typeof renderers[defaultRendererKey] === "function" &&
+        renderers[defaultRendererKey]) ||
+      (typeof renderers[defaultRendererKey] === "object" &&
+        operator &&
+        renderers[defaultRendererKey][operator]) ||
+      (typeof renderers[defaultRendererKey] === "object" &&
+        renderers[defaultRendererKey][defaultRendererKey]) ||
```
zettca, line 57: There's 2 big similar "chains" here - checking for `renderers[attrType]` and then `renderers[defaultRendererKey]`. Maybe a refactor like so would make it easier to follow

```suggestion
      getRenderer(renderers[attrType], operator) ||
      getRenderer(renderers[defaultRendererKey], operator) ||
      undefined;
```

zettca:
```tsx
const getRenderer = (renderer: HvQueryBuilderRenderers, operator: string) =>
  // 1. Custom renderer
  (typeof renderer === "function" && renderer) ||
  // 2. Custom operator renderer
  (typeof renderer === "object" && operator && renderer[operator]) ||
  // 3. Custom DEFAULT renderer
  (typeof renderer === "object" && renderer[defaultRendererKey]);
```

MEsteves22: Good call, I'll add this 🙏

plagoa (approved):

PR #3910, opened 2023-12-18
feat(AvatarGroup): add avatar group component

plagoa:


packages/core/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+import { HvAvatar } from "../Avatar/Avatar";
+
+export { staticClasses as avatarGroupClasses };
+
+export type HvAvatarGroupClasses = ExtractNames<typeof useClasses>;
+
+export interface HvAvatarGroupProps extends HvBaseProps {
+  /** Inline styles to be applied to the root element. */
+  style?: CSSProperties;
```
MEsteves22, line 23: I believe the `style` property is already included in the `HvBaseProps` type so we shouldn’t have to list this property again.

packages/core/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+    default:
+      return spacing === "compact" ? 30 : 18;
+  }
+};
+
+/**
+ * The AvatarGroup component is used to group multiple avatars.
+ */
+export const HvAvatarGroup = forwardRef<any, HvAvatarGroupProps>(
+  (props, ref) => {
```
MEsteves22, line 70: We could use a less generic type than `any` for the `ref`, like `HTMLDivElement`

packages/core/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    const totalChildren = Children.count(children);
+    const zIndexMultiplier = toBack ? -1 : 1;
+    const willOverflow = totalChildren > maxVisible;
+
+    return (
+      <div
+        className={cx(
+          classes.root,
+          css({ flexDirection: direction }),
```
MEsteves22, line 96: We could create a classes for each `direction` instead of having the code in the component file

packages/core/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+};
+
+/**
+ * The AvatarGroup component is used to group multiple avatars.
+ */
+export const HvAvatarGroup = forwardRef<any, HvAvatarGroupProps>(
+  (props, ref) => {
+    const {
+      className,
+      style,
```
MEsteves22, line 73: `style` is extracted but I can’t find where it’s used in the code

packages/core/src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+          css({ flexDirection: direction }),
+          className
+        )}
+        ref={ref}
+        {...others}
+      >
+        {Children.map(children, (child: any, index: number) => {
+          if (index < maxVisible) {
+            return cloneElement(child, {
+              style: {
```
MEsteves22, line 105: Do you think it’s possible to leverage a bit more `classes` in the code to enable users to more easily override the styles if necessary? Also, I believe using `style` for our styles instead of `className` makes it more difficult to override the styles externally.

MEsteves22 (commented):
Looks good 👌🏼

I just have some minor comments:

packages/core/src/components/AvatarGroup/AvatarGroup.tsx
plagoa: These styles are dynamic and need to be calculated on the fly. I needed to use the `style` prop to ensure the `zIndex` was preserved and the avatars were stacked properly.

MEsteves22: Okay, makes sense!

MEsteves22 (approved):
LGTM 👌🏻

We just need to fix the linting, and I think the last two samples could benefit from using the raw format we already use in other samples because some code is missing.

plagoa:
> LGTM 👌🏻
> 
> We just need to fix the linting, and I think the last two samples could benefit from using the raw format we already use in other samples because some code is missing from the code snippets.

Updated.

HQFOX (approved):

PR #3911, opened 2023-12-18
chore: app shell example updated to v1.x

MEsteves22:
- App shell example updated to v1.x to work with the latest changes
   - Migration guide: https://github.com/lumada-design/hv-app-shell/blob/main/docs/migrationFromBeta.md


examples/uikit-app-shell/vite.config.ts
```diff
@@ -17,6 +17,9 @@ export default defineConfig(({ mode }) => ({
       },
     }),
     tsconfigPaths(),
-    HvAppShellVitePlugin({ mode }),
+    HvAppShellVitePlugin({
+      mode,
+      modules: ["src/pages/Home", "src/pages/Page2"],
```
zettca, line 22: Since this is a simple/minimal example, we could leverage the `autoMenu` and `autoViewsAndRoutes` features from App Shell - simplifying this and the config file

```suggestion
      autoMenu: true,
      autoViewsAndRoutes: true,
```

MEsteves22: Yup, I'll use these options 🙏

examples/uikit-app-shell/src/pages/Home/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./Home";
```
zettca, line 1: Doesn't App Shell worth with `autoViewsAndRoutes` and `src/pages/*.tsx`? 🤔

MEsteves22: I wasn't able to make it work with `src/pages/*.tsx` but I can try again in case I did something wrong the first time around and ask the App Shell team about this

MEsteves22: I tried again but it still not working. I asked the App Shell about this, we'll see!

MEsteves22:
Reopening since v1.x is now available

zettca (commented):
Can we bump `npm i vite@5 typescript@5`? It seems like App Shell depends on it indirectly:

```
npm WARN Could not resolve dependency:
npm WARN peer vite@"^5.0.0" from vite-plugin-static-copy@1.0.0
```

zettca (approved):

PR #3915, opened 2023-12-20
fix(QueryBuilder): buttons not disabled in read only and tests updated

MEsteves22:
- New query builder tests added [as requested](https://github.com/lumada-design/hv-uikit-react/pull/3908#pullrequestreview-1789361870)
- Query builder tests updated
- 2 bugs fixed:
   - In read only mode, some buttons were not disabled in the query builder
   - The `disabled` prop was extracted and not passed down to the typography component

### DIFF
--- a/packages/core/src/QueryBuilder/RuleGroup/RuleGroup.tsx
+++ b/packages/core/src/QueryBuilder/RuleGroup/RuleGroup.tsx
@@ -7,7 +7,6 @@ import { HvMultiButton } from "../../MultiButton";
 import { HvTypography } from "../../Typography";
 import { IconButton } from "../../utils/IconButton";
 import { ExtractNames } from "../../utils/classes";
-
 import { Rule } from "../Rule";
 import { useQueryBuilderContext } from "../Context";
 import { useClasses } from "../QueryBuilder.styles";
@@ -199,6 +198,7 @@ export const RuleGroup = ({
                   dispatchAction({ type: "add-rule", id });
                 }}
                 className={classes.createConditionButton}
+                disabled={readOnly}
               >
                 {`${labels.empty?.createCondition}`}
               </HvTypography>
@@ -212,6 +212,7 @@ export const RuleGroup = ({
                       dispatchAction({ type: "add-group", id });
                     }}
                     className={classes.createGroupButton}
+                    disabled={readOnly}
                   >
                     {`${labels.empty?.createGroup}`}
                   </HvTypography>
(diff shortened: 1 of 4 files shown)


packages/core/src/QueryBuilder/RuleGroup/RuleGroup.tsx
```diff
@@ -198,7 +197,10 @@ export const RuleGroup = ({
                 onClick={() => {
                   dispatchAction({ type: "add-rule", id });
                 }}
-                className={classes.createConditionButton}
+                className={cx(classes.createConditionButton, {
+                  [classes.createButtonDisabled]: readOnly,
+                })}
```
zettca, line 202: Can we use the `:disabled` selector instead of adding new classes?

packages/core/src/Typography/Typography.tsx
```diff
@@ -92,14 +92,15 @@ export const HvTypography: <C extends React.ElementType = "p">(
...
       link = false,
-      disabled = false,
       noWrap = false,
       paragraph = false,
       ...others
     } = useDefaultProps("HvTypography", props);
     const { classes, cx } = useClasses(classesProp);
     const { activeTheme } = useTheme();
 
+    const { disabled } = props;
```
zettca, line 102: If we want to forward this to the component, we should forward it directly (ie. `disabled={disabled}`) - destructuring in multiple places makes it harder to follow what is/isn't forwarded

packages/core/src/QueryBuilder/RuleGroup/RuleGroup.tsx
MEsteves22: I did that but the tests were breaking. For some reason, the `pointerEvents: "none"` style was applied even if the button was not disabled. I'll try again though.

packages/core/src/Typography/Typography.tsx
MEsteves22: I did that but the same from above was happening. But I'll try again to see if I missing something.

packages/core/src/QueryBuilder/RuleGroup/RuleGroup.tsx
MEsteves22: Updated ✅

packages/core/src/Typography/Typography.tsx
MEsteves22: Updated ✅

zettca (approved):
LGTM 👍

francisco-guilherme (approved):

PR #3920, opened 2024-01-02
test: fix calendar

MEsteves22:
- The calendar tests started breaking with the new year: this was fixed and should no longer break when the year changes.


zettca (approved):
Isn't this an intended easter egg? 😝

PR #3923, opened 2024-01-02
feat(useId): add react useid to useuniqueid hook

HQFOX:


packages/core/src/Controls/LeftControl/LeftControl.tsx
```diff
@@ -41,6 +43,8 @@ export const HvLeftControl = ({
 }: HvLeftControlProps) => {
   const { classes, cx } = useClasses(classesProp);
 
+  const id = useUniqueId(idProp, "hvleftcontrol");
```
zettca, line 46: We don't need this for a11y

packages/core/src/Slider/Slider.tsx
```diff
@@ -632,7 +632,7 @@ export const HvSlider = forwardRef<SliderRef, HvSliderProps>((props, ref) => {
             <HvLabel
               id={setId(elementId, "label")}
               className={classes.label}
-              htmlFor={setId(elementId, "slider")}
+              htmlFor={elementId}
```
zettca, line 635: This doesn't seem right - we're be labelling the parent of the label. And `elementId` isn't being passed to an element

packages/core/src/hooks/useUniqueId.ts
```diff
@@ -1,6 +1,25 @@
...
+  idPrefix?: string
+): string => {
+  const [id, setId] = React.useState<string | undefined>(useReactId());
+
+  React.useMemo(
+    () => {
+      if (!id) setId(uniqueId(idPrefix));
+    }, // loadash unique id can be removed when react 17 support is dropped
+    [id, idPrefix]
+  );
```
zettca, line 22: - `useMemo` means we'll have a different (hence unstable) id in the server-side, right?
  - this should be an effect
  - we're not saving the memo value anywhere
- we shouldn't need the `id` dependency

- can't we get rid of the the `lodash` dependency here? Radix's counter solution looks nice and simple

Basically, can't this just be the same as Radix's example?

HQFOX: I left `loadash` just to avoid any breaking changes but it should be ok. I'm updating it now.

zettca (approved):
LGTM 👍

francisco-guilherme (approved):

PR #3936, opened 2024-01-05
feat(Flow): add Slider parameter renderer

plagoa:
- Added the Slider renderer. 
- Updated the docs.

### DIFF
--- a/packages/lab/src/Flow/Node/Parameters/Slider.tsx
+++ b/packages/lab/src/Flow/Node/Parameters/Slider.tsx
@@ -1,0 +1,60 @@
+import { useState } from "react";
+
+import { css } from "@emotion/css";
+
+import { useReactFlow } from "reactflow";
+
+import { HvSlider, HvSliderProps } from "@hitachivantara/uikit-react-core";
+
+import { HvFlowNodeSliderParam } from "../../types";
+
+interface SliderProps {
+  nodeId: string;
+  param: Omit<HvFlowNodeSliderParam, "type">;
+  data: any;
+}
+
+const classes = {
+  labelContainer: css({
+    marginRight: 0,
+    marginLeft: 0,
+  }),
+  sliderBase: css({
+    padding: 0,
+  }),
+};
+
+const Slider = ({ nodeId, param, data }: SliderProps) => {
+  const { id } = param;
+
+  const reactFlowInstance = useReactFlow();
+
+  const [value, setValue] = useState(data[id]);
+
+  const onSliderChange: HvSliderProps["onChange"] = (val) => {
+    const nodes = reactFlowInstance.getNodes();
+    const newNodes = nodes.map((node) => {
+      if (node.id === nodeId) {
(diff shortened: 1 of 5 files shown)


packages/lab/src/Flow/types/flow.ts
```diff
@@ -95,7 +98,18 @@ export interface HvFlowNodeSelectParam extends HvFlowNodeSharedParam {
...
 }
 
-export type HvFlowNodeParam = HvFlowNodeSelectParam | HvFlowNodeTextParam;
+export interface HvFlowNodeSliderParam extends HvFlowNodeSharedParam {
+  type: "slider";
+  minPointValue?: HvSliderProps["minPointValue"];
+  maxPointValue?: HvSliderProps["maxPointValue"];
+  markStep?: HvSliderProps["markStep"];
+  divisionQuantity?: HvSliderProps["divisionQuantity"];
+}
```
zettca, line 107: Why stop at these 4 props? What if the user needs (for instance, `formatMark`)? We're already forwarding any props to the `HvSlider`, I think we should allow in on the types too
(we might need to Omit some props)

```suggestion
export interface HvFlowNodeSliderParam extends HvFlowNodeSharedParam,
  Omit<HvSliderProps, keyof HvFlowNodeSharedParam>{
  type: "slider";
}
```

packages/lab/src/Flow/Node/Parameters/Slider.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const Slider = ({ nodeId, param, data }: SliderProps) => {
+  const { id } = param;
+
+  const reactFlowInstance = useReactFlow();
+
+  const [value, setValue] = useState(data[id]);
+
+  const onSliderchange: HvSliderProps["onChange"] = (val) => {
```
zettca, line 34: We shouldn't need to control the value of the element internally, we could use `defaultValues={data[id]}`

Also, typo on the capitalisation of `onSliderChange`

```suggestion
  const onSliderChange: HvSliderProps["onChange"] = (val) => {
```

packages/lab/src/Flow/stories/Usage.stories.mdx
```diff
@@ -243,7 +243,18 @@ interface HvFlowNodeSelectParam {
...
 }
 
-type HvFlowNodeParam = HvFlowNodeSelectParam | HvFlowNodeTextParam;
+export interface HvFlowNodeSliderParam {
+  type: "slider";
+  minPointValue: number;
+  maxPointValue: number;
+  markStep: number;
+  divisionQuantity: number;
+}
```
zettca, line 252: We can simplify this is we point to the Slider props. Also, the `id` and `label` keys are present in the other params but not this one

Ideally, we wouldn't have to duplicate the types

francisco-guilherme (approved):

plagoa:
@zettca updated PR.

zettca (approved):

Issue #3852, opened 2023-11-27
HvFileUploader acceptedFiles prop


fookoo:
### Which version of NEXT UI Kit are you opening this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

According to [documentation](https://lumada-design.github.io/uikit/master/?path=/docs/widgets-file-uploader--docs)

acceptedFiles is suppose to describe what file extensions are acceptable.

so i assume that code:

```
        <HvFileIploader
          fileList={fileList}
          maxFileSize={MAX_FILE_UPLOAD_SIZE}
          acceptedFiles={['scd']}
          onFilesAdded={(files) => {
            setFileList(files)
            console.log(files.at(0)?.type)
          }}
        />
```

allows me to upload any *.scd file

but it gives me type error when i tried to upload scd file

`console.log(files.at(0)?.type)`
 gives me empty string
 
 
![image](https://github.com/lumada-design/hv-uikit-react/assets/7860124/8544749d-1152-40c5-9d87-5148c571c5ef)

### Expected behavior 🤔

to be able to specify extension 

### Your Environment 🌎

Device: Desktop
OS: Windows 11
Browser: Chrome
Version 5.36.1

MEsteves22:
Hello @fookoo 👋

In this case you'll have to use `acceptedFiles={[".scd"]}` (the dot is missing).

smietanka:
Hello @MEsteves22 
Unfortunetly I had the same problem. I used `.scd` file extension as you mentioned. 
Using this solution (with dot) I started to see 
![image](https://github.com/lumada-design/hv-uikit-react/assets/17428659/1a1e553b-2fd0-4d02-b144-c1610d4f96ae)
and selected by default `scd` extension inside of default browser file uploader. Unfortunetly error still occures. Even this I am still able to upload file with different extension
![image](https://github.com/lumada-design/hv-uikit-react/assets/17428659/bde6a10c-f10c-4ece-8209-f529e529acca)

MEsteves22:
Hello 👋

Can you let us know if you are able to upload an `scd` file using [this example](https://stackblitz.com/edit/github-ixonro-zenoa5?file=src%2FApp.tsx,src%2Fmain.tsx), please? On my hand it seems to be working:

[...]

PR #3942, opened 2024-01-08
fix(FileUploader): validate function ESM import

zettca:
`attr-accept` ESM build is producing CJS code, leading to runtime issues (see [this sandbox](https://stackblitz.com/edit/github-zugymp?file=src%2FApp.tsx%3AL9))

Furthermore, ~`attr-accept`'s implementation~our custom validation doesn't seem to be equivalent to the native one - see [this issue](https://github.com/lumada-design/hv-uikit-react/issues/3852)

This PR:
- Copies `attr-accept`'s code to our codebase
  - we should evaluate removing this in favour of HTML's accept and allowing a custom validation
- Changes docs to align validation with native APIs
- Some minor refactors

### DIFF
--- a/packages/core/src/FileUploader/DropZone/DropZone.tsx
+++ b/packages/core/src/FileUploader/DropZone/DropZone.tsx
@@ -2,7 +2,7 @@ import React, { useRef, useState } from "react";
 
 import uniqueId from "lodash/uniqueId";
 
-import accept from "attr-accept";
+import validateAccept from "attr-accept";
 
 import { Doc } from "@hitachivantara/uikit-react-icons";
 
@@ -78,7 +78,7 @@ export interface HvDropZoneProps {
   /**
    * Files extensions accepted for upload.
    */
-  acceptedFiles: string[];
+  accept?: React.InputHTMLAttributes<HTMLInputElement>["accept"];
   /**
    * Max upload size
    * */
@@ -106,7 +106,7 @@ export const HvDropZone = (props: HvDropZoneProps) => {
     id: idProp,
     classes: classesProp,
     labels,
-    acceptedFiles,
+    accept,
     maxFileSize,
     inputProps,
     hideLabels,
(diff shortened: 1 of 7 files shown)


packages/core/src/FileUploader/DropZone/DropZone.tsx
```diff
@@ -118,31 +138,32 @@ export const HvDropZone = (props: HvDropZoneProps) => {
...
+    setDragState(true);
   };
 
   const onChangeHandler = (filesList: FileList) => {
     const filesToProcess = Object.values(filesList);
 
     const newFiles = filesToProcess.map((file) => {
-      const newFile: HvFileData = file;
+      const newFile: HvFileData = {
+        ...file,
```
MEsteves22, line 161: It seems like we can't spread the `file` to copy it, otherwise we get this:

Screenshot 2024-01-08 at 17 53 45

We could do something like: 

```tsx
const newFile: HvFileData = file;
newFile.id = uniqueId("uploaded-file-data-");
```

MEsteves22 (commented):
Very nice, good catch 👌🏻 I just left some comments.

packages/core/src/FileUploader/DropZone/DropZone.tsx
zettca: Nice catch - we can't copy files with spread 🤦

I ended up still cloning the file, so that we're not mutating the input (and issues with test)

packages/core/src/FileUploader/DropZone/DropZone.tsx
```diff
@@ -118,31 +138,32 @@ export const HvDropZone = (props: HvDropZoneProps) => {
...
+      };
 
       const isSizeAllowed = file.size <= maxFileSize;
-      const isFileAccepted =
-        !acceptedFiles.length ||
-        acceptedFiles.indexOf(file.type.split("/")[1]) > -1 ||
-        acceptedFiles.some((acceptExtension) =>
-          accept({ name: file.name, type: file.type }, acceptExtension)
-        );
+      const isFileAccepted = !accept || validateAccept(file, accept);
```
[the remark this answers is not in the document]
zettca: That's right!
That's actually the piece of code that's accepting the `{["jpg", "jpeg", "png"]}` format, not `attr-accept` - I've added it back.

MEsteves22 (approved):

PR #3949, opened 2024-01-10
chore(Flow): small refactors & type fix

MEsteves22:
- The node parameters were refactored to simplify the code: there was no need for `reactFlowInstance.getNodes()`
- `HvDashboardNode` updates:
   - `Partial` added to `labels` to not make it mandatory to fill all the labels
   - `title` and `description` removed from `labels` since they weren't used
   - `children` only rendered when defined to avoid adding unnecessary space
- The `HvFlowNodeSharedParam` type was simplified in the docs


zettca (approved):
Nice 👌

PR #3958, opened 2024-01-12
fix(FilerGroup): error when all option is selected & wrong counter

MEsteves22:
This PR fixes the following two bugs:
- When the filter group is empty and the “All” option is selected an error is thrown and the option can’t be selected. It’s only possible to select “All” after selecting another option: 

https://github.com/lumada-design/hv-uikit-react/assets/43220251/2e66556c-aba2-4cd2-be42-d2395d2340d6

- When fixing this bug I noticed the counter was not working as expected:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/84981b12-51ef-4645-8c6d-40af6fc1860c


MEsteves22:
ℹ️ After these changes are accepted, I'll back port these fixes to v3 and v4 since the first bug was introduced a few weeks ago when we merged a PR.

HQFOX (approved):

zettca (approved):
Nice 👍

PR #3959, opened 2024-01-12
feat(icons): add enter icon from ds 5.9

HQFOX:


packages/icons/assets/Enter.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="10 11 13 11">
+    <path d="M22.43 11h-6.42v1h5v5h-9.13l2.72-2.72-.71-.71L10 17.46l3.89 3.89.71-.71-2.65-2.65H22V11h.43Zm-11.02 6.46h.01v.01h-.01v-.01Z" fill="#414141"/>
+</svg>
```
zettca, line 3: This should be formatted with `svgo` (`npm run optimize`) instead of `prettier`, otherwise we'll get a diff any time we run `optimize`

zettca (approved):

PR #3972, opened 2024-01-16
fix(TimePicker): allow empty controlled value with null

zettca:
Allow passing `null` to `TimePicker`'s controlled state `value` to represent an "empty" controlled value


MEsteves22 (commented):
🙏

MEsteves22 (approved):

HQFOX (approved):

Issue #3975, opened 2024-01-18
Issues in Dropdown Column Renderer


Quernest:
### Which version of NEXT UI Kit are you opening this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

I identified two issues with the [Dropdown Column Renderer](https://lumada-design.github.io/uikit/master/?path=/docs/visualizations-table-table-renderers-dropdown-column-renderer--docs) in tables:

1. When there are 2-3 rows, the dropdown is incorrectly displayed, appearing within the table and getting cropped.
2. When no row data is provided for the column, the page crashes with the following error:

```text
TypeError: Cannot read properties of undefined (reading 'length')
at Cell (renderers.tsx:226:27)
at renderWithHooks (react-dom.development.js:16305:18)
```

### Expected behavior 🤔

1. The dropdown should display appropriately, even when there are 2-3 rows, without getting cropped.
2. The page should handle scenarios where no row data is provided for the column without crashing.

### Steps to reproduce 🕹

**For the Dropdown Display Issue:**
1. Create a table with 2-3 rows.
2. Add a Dropdown Column Renderer to the table.
3. Observe the display of the dropdown in the specified condition.

**Recording:**

https://github.com/lumada-design/hv-uikit-react/assets/21086792/51f20aae-1b81-44d2-97f6-925ef6695155

**For the Page Crash Issue:**
1. Ensure no row data is provided for the column.
2. Load or refresh the page.
3. Observe the error and resulting crash.

For example:
```tsx
<Table
  title="Data"
  columns={[
    {
      Header: 'Column Name',
      accessor: 'columnName',
    },
    hvDropdownColumn<any, string>(
      {
        Header: 'Input Type',
        accessor: 'inputType',
        id: 'inputType',
      },
      undefined,
      'Select Input Type...',
      'Select Input Type...',
      () => console.log('select me'),
    ),
    hvDropdownColumn<any, string>(
      {
        Header: 'Severity',
        accessor: 'severity',
        id: 'severity-header',
      },
      undefined,
      'Select severity...',
      'Select severity...',
      () => console.log('select me'),
    ),
  ]}
  data={[
    {
      columnName: 'State Code',
      // inputType: [], <- IS MISSED
      severity: getDropdownOptions(
        ['Critical', 'Major', 'Average', 'Minor'],
[...]

PR #3978, opened 2024-01-19
fix(Table): dropdown renderer crashing & cut-off

MEsteves22:
Addresses https://github.com/lumada-design/hv-uikit-react/issues/3975

- The table was crashing when the dropdown renderer didn't have any values for the dropdown
- The dropdown list was cut-off when there wasn't many rows in the table: `disablePortal` was removed to fix this. 
   - Let me know if you think there's any problem removing `disablePortal` and it should be opt-out instead for backward compatibility purposes 

After:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/7bd65816-21e3-4014-9335-0a34a604db42

### DIFF
--- a/packages/core/src/Table/stories/TableRenderers/TableRenderers.tsx
+++ b/packages/core/src/Table/stories/TableRenderers/TableRenderers.tsx
@@ -27,6 +27,7 @@ import {
   hvProgressColumn,
   theme,
 } from "@hitachivantara/uikit-react-core";
+import { fireEvent, screen, waitFor, within } from "@storybook/testing-library";
 
 import { makeRenderersData, NewRendererEntry } from "../storiesUtils";
 
@@ -1881,6 +1882,8 @@ const DropdownColumnRenderer = () => {
 
   const [data, setData] = useState(initialData);
 
+  // initialData[0].severity = undefined;
+
   const columns = useMemo(() => {
     return [
       hvDropdownColumn<NewRendererEntry, string>(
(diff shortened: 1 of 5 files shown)


MEsteves22:
@VGervasio will be adding tests for these bugs 🙏

plagoa (approved):

francisco-guilherme (approved):

zettca (approved):

packages/core/src/Table/stories/TableRenderers/TableRenderers.tsx
```diff
@@ -1985,6 +1990,23 @@ const DropdownColumnRenderer = () => {
...
+        // reduce the number of visible rows
+        fireEvent.click(
+          screen.getByRole("combobox", { name: "Select how many to display" })
+        );
+        fireEvent.click(screen.getByRole("option", { name: "5" }));
+
+        const tableElement = screen.getByRole("table", {
+          name: "Severity",
+        });
+        fireEvent.click(within(tableElement).getByText("Major"));
```
zettca, line 2005: We're doing unnecessary work here (indirectly testing the Pagination); we could select the first "Major" with `getAllByText` without needing to "hide" the others

```suggestion
        const [firstMajor] = screen.getAllByText("Major");
        fireEvent.click(firstMajor);
```

VGervasio: We need to do it because the issue only appears with few rows on the table. However the way to do it needs to be improved.
After talking with @MEsteves22 and with her help we decided we will start showing the table with only 5 elements to save the need for this interaction.

MEsteves22: The test was updated and the pagination starts with 5 ✅

MEsteves22:
The tests look good 👌🏻

Applitools is expected to break since a new test was added. The PR is ready to merge for me 🚀

PR #3979, opened 2024-01-19
feat: update handles layout

plagoa:


packages/lab/src/Flow/Node/BaseNode.tsx
```diff
@@ -231,21 +231,37 @@ export const HvFlowBaseNode = ({
...
+                      style: {
+                        backgroundColor: theme.colors.secondary_60,
+                        width: 8,
+                        height: 8,
+                      },
+                    })}
+                  />
+                  <HvTypography>{input.label}</HvTypography>
+                  {input.isMandatory &&
+                    !isInputConnected(id, "target", input.id!, inputEdges) && (
```
MEsteves22, line 259: I think we should use the `isConnected` calculated above to avoid repeating code and calculating again (same for the outputs below)

packages/lab/src/Flow/Node/BaseNode.tsx
```diff
@@ -231,21 +231,37 @@ export const HvFlowBaseNode = ({
...
-              </div>
-            ))}
+            {inputs?.map((input, idx) => {
+              const handleId = input.id ?? idx.toString();
+              const isConnected = isInputConnected(
+                id,
+                "target",
+                handleId,
+                inputEdges
+              );
```
MEsteves22, line 241: We actually no longer need to calculate the `handleId` and we can always use `input.id` since we generate ids for the inputs and outputs that don't have one defined. Thus, we can do: (same for the outputs below)

```tsx
const isConnected = isInputConnected(id, "target", input.id!, inputEdges)
```

packages/lab/src/Flow/Node/BaseNode.tsx
```diff
@@ -231,21 +231,37 @@ export const HvFlowBaseNode = ({
...
+                    type="target"
+                    isConnectableStart={false}
+                    id={input.id}
+                    position={Position.Left}
+                    {...(isConnected && {
+                      style: {
+                        backgroundColor: theme.colors.secondary_60,
+                        width: 8,
+                        height: 8,
+                      },
```
MEsteves22, line 254: What do you think about creating a class for this to have all the styling in the same place? (same for the outputs below)

MEsteves22 (commented):
Looks good 🙏

I just left some comments and noticed the following behaviour:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/8c6cabbd-d976-4565-b95b-398e07df4090

MEsteves22 (approved):
🙏

francisco-guilherme (approved):

PR #3983, opened 2024-01-22
feat: apply root element styles via class for easier overrides

nantunes:


packages/styles/src/CssBaseline.ts
```diff
@@ -76,6 +76,16 @@ const baseline = {
...
   },
+
+  ".uikit-root-element": {
+    backgroundColor: "var(--uikit-colors-backgroundColor)",
+    accentColor: "var(--uikit-colors-secondary)",
+    color: "var(--uikit-colors-secondary)",
+    fontSize: "var(--uikit-typography-body-fontSize)",
+    fontWeight: "var(--uikit-typography-body-fontWeight)",
+    lineHeight: "var(--uikit-typography-body-lineHeight)",
+    letterSpacing: "var(--uikit-typography-body-letterSpacing)",
```
zettca, line 87: These CSS var names are internal. We should use the `theme` object instead (shorter, type-safe, future-proof).
```suggestion
    backgroundColor: theme.colors.backgroundColor,
    accentColor: theme.colors.secondary,
    color: theme.colors.secondary,
    fontSize: theme.typography.body.fontSize,
    fontWeight: theme.typography.body.fontWeight,
    lineHeight: theme.typography.body.lineHeight,
    letterSpacing: theme.typography.body.letterSpacing,
```

packages/core/src/utils/theme.ts
```diff
@@ -37,18 +25,9 @@ export const setElementAttrs = (
...
-    setElementStyle(element, {
-      colorScheme,
-      backgroundColor: theme.colors.backgroundColor,
-      accentColor: theme.colors.secondary,
-      color: theme.colors.secondary,
-      fontSize: theme.typography.body.fontSize,
-      fontWeight: theme.typography.body.fontWeight,
-      lineHeight: theme.typography.body.lineHeight,
-      letterSpacing: theme.typography.body.letterSpacing,
-      fontFamily: theme.fontFamily.body,
```
zettca, line 50: looks like we lost `fontFamily` in the refactor; is this intentional?

nantunes: No, it wasn't.

MEsteves22 (approved):

zettca (approved):

PR #3984, opened 2024-01-22
fix: ensure the charts change when the theme changes

plagoa:


packages/viz/src/BaseChart/BaseChart.tsx
```diff
@@ -29,7 +29,7 @@ export const HvBaseChart = forwardRef<ReactECharts, HvBaseChartProps>(
   (props, ref) => {
     const { option, width, height } = props;
 
-    const { theme } = useVizTheme();
+    const { theme, activeTheme } = useVizTheme();
```
zettca, line 32: Why do we need this? Isn't `activeTheme` the same value as the one provided by `useTheme`? We could just use it instead
```suggestion
    const { theme } = useVizTheme();
    const { activeTheme } = useTheme();
```

packages/viz/src/BaseChart/BaseChart.tsx
```diff
@@ -60,7 +60,13 @@ export const HvBaseChart = forwardRef<ReactECharts, HvBaseChartProps>(
...
         replaceMerge: ["xAxis", "yAxis", "series", "dataset"],
       });
-    }, [theme, option]);
+
+      return () => {
+        // Dispose the instance of the chart when the component unmounts. This ensures that
+        // the chart theme will update when the theme changes.
+        instance.dispose();
+      };
+    }, [theme, option, activeTheme]);
```
zettca, line 69: Do we want `activeTheme` to affect the rest of the behaviour? Couldn't this be a separate `useEffect`?

MEsteves22 (approved):

PR #3985, opened 2024-01-22
chore(Flow): missing labels added

MEsteves22:
- Some values were hardcore in the flow component which makes it impossible to internationalise these values. New labels were added to fix this.
- The default node actions were wrapped by a tooltip to add the label (I opted to use a tooltip to have a visual label instead of only using `aria-label`)


zettca (approved):
👌

PR #3989, opened 2024-01-23
feat(hvDropdownColumn): add dropdown props

HQFOX:
I added this as a solution to [this](https://hv-eng.atlassian.net/jira/software/c/projects/HVUIKIT/boards/267?selectedIssue=HVUIKIT-7016) as  way for the user to be able to change how the dropdown behaves.


MEsteves22 (approved):
Nice, enabling users to override the `dropdownProps` seems every useful 🙏

plagoa (approved):

PR #3993, opened 2024-01-25
feat(hvDatePicker): add time selection to date-picker

fookoo:
I have added option to display HvTimeSelector inside HvDatePicker and choose time both for start and end date

### DIFF
--- a/packages/core/src/Calendar/Calendar.tsx
+++ b/packages/core/src/Calendar/Calendar.tsx
@@ -37,6 +37,10 @@ export interface HvCalendarProps {
    */
   value?: DateRangeProp | Date;
   /**
+   * Controls the visibility of time selector
+   */
+  showTime?: boolean;
+  /**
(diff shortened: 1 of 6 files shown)


zettca:
Hey @fookoo, thank you for your contribution 💪 

Can you share with us a bit more about your particular use-case?
Do you need a "date-time picker" for your application, or could you use separate `HvDatePicker` and `HvTimePicker` components?

I'm asking because there is no mention on this "date-time" behaviour in the Design System specification.
Furthermore, we would probably want this to be a new (eg. `HvDateTimePicker`) component (similarly to the HTML spec and other libraries), as `HvDatePicker` and `HvTimePicker` are two distinct form elements and shouldn't be nested inside one another; and not to impact current `HvDatePicker` users with the changes

francisco-guilherme:
Hi @fookoo we'll close this PR for now.

PR #3996, opened 2024-01-26
fix(DropdownMenu): add variant prop to be inline with other components

HQFOX:
It doesn't make much sense that this component has a prop categories while others have variants and the values are the same.

category was marked as deprecated and not remove to not cause breaking changes

### DIFF
--- a/packages/core/src/DropDownMenu/DropDownMenu.tsx
+++ b/packages/core/src/DropDownMenu/DropDownMenu.tsx
@@ -55,8 +55,13 @@ export interface HvDropDownMenuProps
   expanded?: boolean;
   /** When uncontrolled, defines the initial expanded state. */
   defaultExpanded?: boolean;
-  /** The variant to be used in the header. */
+  /**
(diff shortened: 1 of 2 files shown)


zettca (approved):

packages/core/src/DropDownMenu/DropDownMenu.tsx
```diff
@@ -80,6 +85,7 @@ export const HvDropDownMenu = (props: HvDropDownMenuProps) => {
     expanded,
     defaultExpanded = false,
     category = "secondaryGhost",
+    variant = "secondaryGhost",
```
MEsteves22, line 88: If the user doesn't use the new `variant` and it's still using `category`, the default value for the `variant` property will always override the value set for `category`. I believe not setting a default value for `variant` should fix this for now.

MEsteves22 (changes requested):

MEsteves22 (approved):

PR #3997, opened 2024-01-28
test: playwright flow test scenarios

VGervasio:


.github/workflows/tests-playwright.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install Playwright
+        run: npx playwright install --with-deps
+
+      - name: Run Playwright tests
```
MEsteves22, line 32: If I'm not mistaken, the `playwright` workflow will use the docs built and published in the `pr` workflow (when run in a PR). However, when the docs are built in the `pr` workflow, it takes some minutes before the docs are actually available in the url. Did you have any problems with that here?

MEsteves22 (commented):
Great stuff 👌🏻 I just left some comments:

.github/workflows/tests-playwright.yml
```diff
@@ -0,0 +1,35 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        PROJECT: ["chromium", "firefox", "webkit"]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install dependencies
```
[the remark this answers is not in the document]
VGervasio: I will try it.

playwright.config.ts
```diff
@@ -0,0 +1,96 @@
...
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    headless: true,
+    baseURL: process.env.CI ? process.env.URL : "http://localhost:6006",
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: "on-first-retry",
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
```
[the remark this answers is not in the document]
VGervasio: Instead of the "generic" chromium I think we can have Chrome and Edge. Or we can even have all 3. I will try it.

.github/workflows/tests-playwright.yml
```diff
@@ -0,0 +1,35 @@
...
+    name: Playwright Tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        PROJECT: ["chromium", "firefox", "webkit"]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

.github/workflows/tests-playwright.yml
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

playwright.config.ts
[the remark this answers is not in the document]
VGervasio: Changed in new commit. For now all 3 are included. We can see if we should remove the "generic" chromium.

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+
+    expect(nodeInCanvasBox?.x).toBeGreaterThanOrEqual(dragDestinationBox!.x);
+    expect(nodeInCanvasBox?.x).toBeLessThanOrEqual(dragDestinationBox!.x + 20);
+    expect(nodeInCanvasBox?.y).toBeGreaterThanOrEqual(
+      dragDestinationBox!.y + 20
+    );
+    expect(nodeInCanvasBox?.y).toBeLessThanOrEqual(dragDestinationBox!.y + 40);
+  });
+
+  test("should be able to copy a flow", async ({ page }) => {
```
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+      .hover();
+    await page.getByRole("button", { name: "Duplicate" }).hover();
+    await expect(
+      page.getByRole("tooltip", { name: "Duplicate" })
+    ).toBeVisible();
+    await page.getByRole("button", { name: "Duplicate" }).click();
+    expect(await createdNode.all()).toHaveLength(2);
+  });
+
+  test("should be able to delete a flow", async ({ page }) => {
```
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+    await flowCanvasLocator.getByRole("button", { name: "Delete" }).hover();
+    await expect(page.getByRole("tooltip", { name: "Delete" })).toBeVisible();
+    await flowCanvasLocator.getByRole("button", { name: "Delete" }).click();
+    await expect(createdNode).not.toBeVisible();
+    await expect(flowCanvasLocator.getByText("Empty Flow")).toBeVisible();
+  });
+});
+
+test.describe("Connections", () => {
+  test("should be able to connect two flows", async ({ page }) => {
```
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+    await expect(
+      flowCanvasLocator.getByRole("button", { name: "Edge" })
+    ).toBeVisible();
+    await deleteConnection("Edge");
+    await expect(
+      flowCanvasLocator.getByRole("button", { name: "Edge" })
+    ).not.toBeVisible();
+  });
+
+  test("should not be able to connect incompatible flows", async ({ page }) => {
```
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+    inputBox = await inputNode.boundingBox();
+    lineChartBox = await flowCanvasLocator
+      .getByRole("button", { name: "Line Chart" })
+      .boundingBox();
+    expect(Math.floor(inputBox!.y - lineChartBox!.y)).toEqual(
+      Math.floor(inputLineChartDiff)
+    );
+  });
+
+  test("should not be able to duplicate/delete flows after interactive button is locked", async ({
```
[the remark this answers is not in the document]
VGervasio: Changed in new commit.

.github/workflows/tests-playwright.yml
MEsteves22: Okay, great 🙏 I agree, let's see how this will behave over time.

MEsteves22 (approved):

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+
+async function dragToPosition3(node: Locator) {
+  await node.dragTo(flowCanvasLocator, { targetPosition: { x: 620, y: 80 } });
+}
+
+async function connectNodes(
+  sourceNode: string,
+  sourceConnection: string,
+  destinyNode: string,
+  destinyConnection
```
zettca, line 38: "Destiny" means else; did you mean "destination"?
Either way, I think we should use `target`, as it's what we're using in the Flow APIs

```suggestion
  targetNode: string,
  targetConnection: string,
```

packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+  await referencePage.mouse.move(
+    initialPointX,
+    initialPointY - verticalMovement
+  );
+  await referencePage.mouse.up();
+}
+
+test.beforeEach(async ({ page }) => {
+  referencePage = page;
+  flowCanvasLocator = page.locator("id=hvFlow:r1:");
```
zettca, line 79: This selector is brittle, as this is a generated `id` and there's no guarantees it won't change.
We could use the `HvFlow-root` class or the `react-flow` class (in the child element) as these are stable

playwright.config.ts
```diff
@@ -0,0 +1,114 @@
...
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  /* webServer: {
+     command: 'npm run doc',
```
zettca, line 110: Isn't something like this faster than the 4+ minute wait for the the build+deploy of the storybook+app+nextra?

```suggestion
     command: 'npm run build:doc && npx -y http-server dist -p 6006',
```

.github/workflows/pr.yml
```diff
@@ -28,3 +28,10 @@ jobs:
       ORG_GHPAGES_DEPLOY_KEY: ${{ secrets.ORG_GHPAGES_DEPLOY_KEY }}
     with:
       publish-folder: pr-${{ github.event.number }}
+
+  playwright:
+    name: Playwright
+    needs: artifacts
```
zettca, line 34: Not a fan of depending/waiting on the build+publish of all these artefacts. We don't really need a publication to test this.
This also means that `tests-playwright.yml` depends on an active deployment, and can't run isolated on a branch's code

zettca (commented):
Nice stuff 🎭 👌

Some small comments:

playwright.config.ts
VGervasio: It is faster, but since these tests are more realistic I was thinking it may be a good idea to run them against "built code". Additionally since it finishes before the Aplitools checks the downside of taking more time is minimized.
But we can discuss it better. I might be missing something.

.github/workflows/pr.yml
VGervasio: Regarding the first paragraph I understand your point, see my previous reply.
Regarding the second paragraph is it possible to have branch's code without an active deployment?

playwright.config.ts
zettca: They're equally "realistic", they're the "production" artifacts.
The difference is that one is served by `http-server`, the other is served by GitHub Pages' server

.github/workflows/pr.yml
zettca: Yes, by running the `http-server` in the workflow run (previous reply)

packages/lab/src/Flow/Flow.spec.ts
VGervasio: Changed in new commit.

playwright.config.ts
VGervasio: Changed in new commit.

.github/workflows/pr.yml
VGervasio: Changed in new commit.

zettca (approved):
🚀

PR #3999, opened 2024-01-29
chore(Flow): useReactFlow export, small refactor, and docs updates

MEsteves22:
- Docs updated to leverage `useFlowNodeUtils` (only missing in one story)
- Clean-up: unnecessary console logs removed
- The nodes' hooks were standardised since some hooks required the node `id` as prop but not `useFlowNodeUtils`. The node `id` is now optional for all hooks. If not passed, the hooks will use React Flow's `useNodeId` to get the `id` of the node the hook is being used in. 
- The `useReactFlow` hook was re-exported as `useFlowInstance` to avoid installing `reactflow` only to use this hook. I also added the `HvFlowInstance` type. In this sense, I removed all imports from `reactflow` in our samples.
- Usage docs updated
- The node handle style was fixed when already connected:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/7077eb14-5bd0-4c6d-960f-ac85b96dcd01

### DIFF
--- a/apps/app/src/pages/Flow/Flow.tsx
+++ b/apps/app/src/pages/Flow/Flow.tsx
@@ -1,5 +1,4 @@
 import { Suspense, useEffect, useMemo, useState } from "react";
-import { Edge, ReactFlowInstance, Node } from "reactflow";
 import {
   HvButton,
   HvGlobalActions,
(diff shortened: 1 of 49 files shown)


plagoa (approved):

zettca (approved):
Nice 👍

PR #4002, opened 2024-01-30
feat(TableSection): add TableSection component

plagoa:
- Added the `HvTableSection` component
- Updated the Table Section samples and documentation
- Updated the `HvSection` component to only show the header section if indeed there's a title


packages/core/src/Table/stories/TableSamples/PropsTableSection.tsx
```diff
@@ -0,0 +1,117 @@
...
+  HvTableCell,
+  HvTypography,
+  HvActionsGeneric,
+  useHvPagination,
+  useHvRowSelection,
+  useHvBulkActions,
+  useHvData,
+} from "@hitachivantara/uikit-react-core";
+
+import { HvTableSection } from "packages/core/src/TableSection";
```
MEsteves22, line 19: Just for consistency, we should import `HvTableSection` from `@hitachivantara/uikit-react-core` in the stories.

packages/core/src/TableSection/TableSection.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    const { classes } = useClasses(classesProp);
+    const elementId = useUniqueId(id, "hvTableSection");
+
+    return (
+      <HvSection
+        ref={ref}
+        id={elementId}
+        className={className}
+        classes={{
```
MEsteves22, line 41: If I'm not mistaken, all our classes will be overridden if the user provides a `classes` object inside `sectionProps`. To avoid this we could simply do:

```tsx
export type HvTableSectionProps = HvSectionProps
```

and then use `{...others}` instead of `{...sectionProps}`.


I also noticed that sometimes the header can have [a shadow underneath](https://designsystem.hitachivantara.com/6c705d900/p/29efb2-table/b/80c89a/i/269215643). We could have a prop for this:

```tsx
export interface HvTableSectionProps extends HvSectionProps {
   raisedHeader?: boolean; // name example, couldn't find a better one at the moment
}
```

MEsteves22 (commented):
Really nice 👌🏻 Just left some comments and I found the following regarding the styling:


https://github.com/lumada-design/hv-uikit-react/assets/43220251/5f6764da-db13-41fa-a1e3-a16ca8700de6


Screenshot 2024-01-31 at 09 46 11


Screenshot 2024-01-31 at 09 47 14

packages/core/src/TableSection/TableSection.tsx
```diff
@@ -0,0 +1,56 @@
...
+          root: cx(classes.section, sectionClasses?.root),
+          header: cx(classes.sectionHeader, sectionClasses?.header),
+          actions: cx(classes.sectionActions, sectionClasses?.actions),
+          content: cx(classes.content, sectionClasses?.content),
```
zettca, line 48: Having these 2 "layers" of  `useClasses` is confusion. We need a better way to address this 😞 

Maybe `HvTableSection` should be a `styled(HvTableSection)` - we're doing nothing besides overriding styles

zettca (commented):
Looking good 👌 

Some notes:

packages/core/src/Section/Section.tsx
```diff
@@ -77,25 +80,31 @@ export const HvSection = forwardRef<HTMLDivElement, HvSectionProps>(
...
-              aria-label={isOpen ? "Collapse" : "Expand"}
-              {...expandButtonProps}
-            >
-              {isOpen ? <Up /> : <Down />}
-            </HvButton>
-          )}
-          {title}
-          <div className={classes.actions}>{actions}</div>
-        </div>
+        {(title || actions || expandable) && (
```
zettca, line 83: Can we add a samples of a `title`'less `HvSection` - it's really common in app mocks, and we'd cover regressions with Applitools

zettca (approved):
LGTM 👌

packages/core/src/Section/Section.stories.tsx
```diff
@@ -159,6 +159,35 @@ export const WithActions: StoryObj<HvSectionProps> = {
   },
 };
 
+export const NoHeader: StoryObj<HvSectionProps> = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "This sample showcases as example where multiple sections are used together.",
```
MEsteves22, line 167: Wrong description 🕵🏻‍♀️

packages/core/src/TableSection/TableSection.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export { staticClasses as tableSectionClasses };
+
+export type HvTableSectionClasses = ExtractNames<typeof useClasses>;
+
+export interface HvTableSectionProps extends HvSectionProps {}
+
+/**
+ * The `TableSection` component is a wrapper for the `Section` component that applies
+ * specific stylings for tables that follow the latest DS specifcations.
```
MEsteves22, line 17:
```suggestion
 * specific stylings for tables that follow the latest DS specifications.
```

MEsteves22 (approved):
Looks great 👌🏻

Not merging since the PR needs to be rebased. I also found these small lapses:

PR #4005, opened 2024-02-01
fix: disabled button inside tooltip

MEsteves22:
Addresses https://github.com/lumada-design/hv-uikit-react/issues/4013

- We have an internal component called `IconButton` which uses a tooltip around a button. When the button is disabled, the error shown above is thrown. To solve this, a new [`focusableWhenDisabled`](https://mui.com/base-ui/react-button/#focus-on-disabled-buttons) property was added to the `HvButton` as proposed in the thread below. When using this property, the button is disabled only through  `aria-disabled` and not through `disabled` in order to be accessible by screen readers and the tooltip to show on hover. 
- Tests updated and new tests created

### DIFF
--- a/packages/core/src/Button/Button.stories.tsx
+++ b/packages/core/src/Button/Button.stories.tsx
@@ -1,4 +1,4 @@
-import { StoryFn, StoryObj } from "@storybook/react";
+import { StoryObj } from "@storybook/react";
 import { css } from "@emotion/css";
 import {
   Delete,
@@ -40,8 +40,21 @@ export const Main: StoryObj<HvButtonProps> = {
   },
 };
 
-export const Variants: StoryFn<HvButtonProps> = () => {
-  return (
+export const Variants: StoryObj<HvButtonProps> = {
+  decorators: [
+    (Story) => (
+      <HvBox
+        sx={{
+          display: "grid",
+          gap: 20,
+          gridTemplateColumns: "repeat(3, 140px)",
+        }}
+      >
+        {Story()}
+      </HvBox>
+    ),
+  ],
+  render: () => (
     <>
       <HvButton variant="primary">Primary</HvButton>
       <HvButton variant="primarySubtle">Primary Subtle</HvButton>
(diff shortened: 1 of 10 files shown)


packages/core/src/utils/IconButton.tsx
```diff
@@ -11,9 +11,20 @@ export interface IconButtonProps extends Omit<HvButtonProps, "icon" | "title"> {
...
-  </HvTooltip>
-);
+}: IconButtonProps) => {
+  const button = (
+    <HvButton icon disabled={disabled} aria-label={title} {...others} />
+  );
+
+  if (disabled) {
+    return button;
+  }
```
zettca, line 23: I think we should "restore" the disabled button + tooltip behaviour.
Previously we were doing something like:
```tsx
<HvTooltip title="label">
  <div> // <-- focusable, but gets the aria-label
    <HvButton disabled>
```
but we get `aria-label` on the wrong element like this


I think the way to approach this is to just not use `<HvButton disabled={disabled}>`
We could use `<HvButton aria-disabled>` and use this prop to disable the button only visually


What do you think?

zettca (commented):
Nice 👍
I had noticed this already and been thinking about a way to solve this issue, since `disabled` buttons aren't focusable and aren't present the A11y tree...

packages/core/src/utils/IconButton.tsx
zettca: It would not look great if we need to toggle the "focusable" though. but this probably shouldn't be the case
```tsx
{...(focusable
  ? {"aria-disabled": true}
  : {disabled: true} 
)}
```

We could instead do like MUI and add a `focusableWhenDisabled` prop that toggles this `disabled`/`aria-disabled` and `tabIndex`

packages/core/src/Pagination/Pagination.tsx
```diff
@@ -241,7 +241,7 @@ export const HvPagination = (props: HvPaginationProps) => {
           className={classes.iconContainer}
           disabled={!canPrevious}
           onClick={() => changePage(0)}
-          title={labels?.firstPage || labels?.paginationFirstPageTitle}
+          title={labels?.firstPage || labels?.paginationFirstPageTitle || ""}
```
zettca, line 244: strange that `useLabels` isn't safeguarding from the `undefined` 🤔

zettca: fix is here https://github.com/lumada-design/hv-uikit-react/pull/4007

packages/core/src/utils/IconButton.tsx
MEsteves22: Oh right, the way I did it the tooltip doesn't appear when the button is disabled 😅 Good idea, I'll look into `focusableWhenDisabled`!

MEsteves22: Updated ✅

MEsteves22:
It was kind of tricky to have this working perfectly, namely due to the `pointerEvents` style we have for the button when disabled. I was not able to remove it since it's needed for when the button is a link, `a`. Without it, the link opens when clicked even if it's disabled. 
However, using `pointerEvents` was making it impossible for the tooltip to appear when the button was hovered. To solve this, I had to override the button's style. 

Let me know what you think 🙏 Tests were added and/or fixed.

zettca:
> I was not able to remove it since it's needed for when the button is a link, `a`. Without it, the link opens when clicked even if it's disabled. 

Maybe we could stop supporting the disabled links use-case. A11y-wise, links with `href` [shouldn't be disabled](https://w3c.github.io/html-aria/#el-a):
> It is NOT RECOMMENDED to use aria-disabled="true" on an a element with an href attribute.
> If a link needs to be programmatically communicated as "disabled", [remove the href attribute](https://w3c.github.io/html-aria/#example-communicate-a-disabled-link-with-aria).

zettca (approved):
Nice 💯

francisco-guilherme (approved):

PR #4010, opened 2024-02-02
docs: review DropDownMenu & Panel

zettca:
- review opened/redudant DropDownMenu samples
  - `iconSelectedColor` made no sense as there's no selection state, but something similar is needed for SB not to enter an infinite-loop 😞 
  - merge disabled & icons samples
    - they don't add much separately, and this way we visually test disabled icons
- review Panel samples
  - use DS's padding instead of `20` override (Applitools changes are expected)

### DIFF
--- a/packages/core/src/DropDownMenu/DropDownMenu.stories.tsx
+++ b/packages/core/src/DropDownMenu/DropDownMenu.stories.tsx
@@ -1,26 +1,20 @@
 import { useState } from "react";
 import { Calendar, Plane, User } from "@hitachivantara/uikit-react-icons";
 import { Meta, StoryObj } from "@storybook/react";
+import { fireEvent, screen, waitFor } from "@storybook/testing-library";
 import {
   HvButton,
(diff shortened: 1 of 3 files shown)


packages/core/src/DropDownMenu/DropDownMenu.stories.tsx
```diff
@@ -33,74 +27,47 @@ export const Main: StoryObj<HvDropDownMenuProps> = {
...
-};
-
-export const DisabledItems: StoryObj<HvDropDownMenuProps> = {
-  parameters: {
-    docs: {
-      description: {
-        story: "DropDownMenu with disabled items.",
+    parameters: {
+      eyes: {
+        runBefore() {
```
MEsteves22, line 56: Looking at Applitools, the dropdown menu doesn't seem to have opened for some reason. Maybe doing `waitFor` the `menu` could solve this.

MEsteves22 (approved):

PR #4012, opened 2024-02-02
refactor(BaseDropdown): break component into reusable pieces

zettca:
Break into the components:
- `DropdownButton`
  - the default button
  - ~refactored to use `HvButton`~ reverted (address this in DS button update?)
- `DropdownContainer`
- `DropdownPanel`
  - `HvPanel` container with DS3 extensions

Improve the DropdownButton to:
- animate the arrow icon
- inherit the icon colors from the text
  - similarly to what `HvButton` does

### DIFF
--- a/packages/core/src/BaseDropdown/DropdownPanel.tsx
+++ b/packages/core/src/BaseDropdown/DropdownPanel.tsx
@@ -1,0 +1,89 @@
+import { forwardRef } from "react";
+import { theme } from "@hitachivantara/uikit-styles";
+
+import { HvPanel, HvPanelProps } from "../Panel";
+import { createClasses, ExtractNames } from "../utils/classes";
+
+export const { staticClasses, useClasses } = createClasses("HvDropdownPanel", {
+  panel: {
+    border: `1px solid ${theme.colors.secondary}`,
+    // borderRadius: 0,
+    marginTop: -1,
+    marginBottom: -1,
+  },
+  panelOpenedUp: {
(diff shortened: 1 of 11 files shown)


packages/core/src/BaseDropdown/DropdownPanel.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * HvBaseDropdown container HvPanel that includes DS3 "extensions"
+ *
+ * TODO:
+ * - make default HvBaseDropdown inherit `HvPanel`'s padding
+ * - refactor extensions into ::before and ::after to remove DS3-only DOM elements
+ */
+export const HvDropdownPanel = forwardRef<HTMLDivElement, HvDropdownPanelProps>(
```
MEsteves22, line 48: I noticed the following in DS3:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/9f5e2d27-8042-4ca5-8656-b7a129585faf

Maybe this has something to do with the `extensionWidth` that was removed in the base dropdown component. 

I also noticed a visual change in the extension regarding the box shadow. Looking at what we previously had, the box shadow seemed less pronounced.

MEsteves22 (commented):
Very nice 🙏 I just noticed the following:

zettca:
This refactor ended up being too large and affecting more than it should have. Taking into consideration that users might be using `HvBaseDropdown` with custom styles and behaviours, this is best kept as-is for now.

I'll follow up with some of the smaller things

PR #4014, opened 2024-02-05
fix(BaseDropdown): popper types

zettca:
`popperProps` are incorrectly inheriting from MUI instead of `react-popper`. keys such as `style` are not present

- replace MUI with react-popper types
- fix misuse of `popperProps`
- inherit types in `HvDropdown` (from `HvBaseDropdown` and `HvFormElement`)

### DIFF
--- a/packages/core/src/Table/stories/TableRenderers/DropdownColumnRenderer.tsx
+++ b/packages/core/src/Table/stories/TableRenderers/DropdownColumnRenderer.tsx
@@ -46,11 +46,6 @@ export const DropdownColumnRenderer = () => {
             return newVal;
           });
           setData(newData);
-        },
-        {
-          popperProps: {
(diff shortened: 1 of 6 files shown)


packages/core/src/Table/stories/TableRenderers/DropdownColumnRenderer.tsx
```diff
@@ -46,11 +46,6 @@ export const DropdownColumnRenderer = () => {
             return newVal;
           });
           setData(newData);
-        },
-        {
-          popperProps: {
-            disablePortal: true,
-          },
```
zettca, line 53: This seems to be equivalent to what we had previously.
Passing `disablePortal: true` to the Dropdown causes other issues:

image

MEsteves22 (approved):

PR #4016, opened 2024-02-06
refactor: fix `forwardRef` type inference

zettca:
- fix `forwardRef` issues in our `forwardRef` + polymorphic/generic components that make us lose TSDoc definitions
  - add `fixedForwardRef` to fix generics (fix based on [this article](https://www.totaltypescript.com/forwardref-with-generic-components#the-problem-with-forwardref))


zettca:
will address this later

plagoa (approved):

MEsteves22 (approved):
👌🏻

PR #4024, opened 2024-02-08
chore(deps): bump minor versions

zettca:


plagoa (approved):

packages/core/src/Carousel/Carousel.tsx
```diff
@@ -7,7 +7,7 @@ import React, {
   useState,
 } from "react";
 
-import useCarousel, { EmblaOptionsType } from "embla-carousel-react";
```
zettca, line 10: this type was removed 😬

PR #4027, opened 2024-02-08
feat(Select): add `HvSelect` component

zettca:
Add `HvSelect` component (and `HvOption`/`HvOptionGroup`)
- uses MUI's `useSelect` behind the scenes (handles multi-select/focus/navigation/search)
- uses `HvButton` for the "header component"

### DIFF
--- a/packages/core/src/Select/Option.tsx
+++ b/packages/core/src/Select/Option.tsx
@@ -1,0 +1,72 @@
+import { useRef } from "react";
+import { useOption } from "@mui/base/useOption";
+import { OptionOwnProps } from "@mui/base/Option";
+import { useForkRef } from "@mui/material/utils";
+
+import { HvListItem, HvListItemProps } from "../ListContainer";
+import { useDefaultProps } from "../hooks/useDefaultProps";
+import { fixedForwardRef } from "../types/generic";
+import { ExtractNames, createClasses } from "../utils/classes";
+import { outlineStyles } from "../utils/focusUtils";
+
+const { staticClasses, useClasses } = createClasses("HvOption", {
+  root: {},
+  highlighted: {
+    ...outlineStyles,
+  },
+});
+
+export { staticClasses as optionClasses };
+
+export type HvOptionClasses = ExtractNames<typeof useClasses>;
+
+export interface HvOptionProps<OptionValue extends {}>
+  extends Omit<HvListItemProps, "value" | "disabled">,
+    Pick<OptionOwnProps<OptionValue>, "disabled" | "label" | "value"> {
+  classes?: HvOptionClasses;
+}
+
+export const HvOption = fixedForwardRef(function HvOption<
+  OptionValue extends {}
+>(props: HvOptionProps<OptionValue>, ref: React.Ref<HTMLLIElement>) {
+  const {
+    classes: classesProp,
+    className,
+    disabled = false,
+    label,
+    value,
+    children,
+    ...others
+  } = useDefaultProps("HvOption", props);
+  const { classes, cx } = useClasses(classesProp);
+
+  const optionRef = useRef<HTMLElement>(null);
+  const rootRef = useForkRef(optionRef, ref);
+
+  const computedLabel =
+    label ??
+    (typeof children === "string"
+      ? children
+      : optionRef.current?.textContent?.trim());
+
+  const { getRootProps, selected, highlighted } = useOption({
+    disabled,
+    label: computedLabel,
+    rootRef,
+    value,
+  });
+
+  return (
+    <HvListItem
+      ref={ref}
+      selected={selected}
+      className={cx(classes.root, className, {
+        [classes.highlighted]: highlighted,
+      })}
+      {...getRootProps()}
+      {...others}
+    >
+      {children}
+    </HvListItem>
+  );
+});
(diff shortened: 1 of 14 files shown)


nantunes (approved):

packages/core/src/Select/Select.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+export const Main: StoryObj<HvSelectProps<{}, false>> = {
+  args: {
+    multiple: false,
+  },
+  argTypes: {},
+  decorators: [decorator],
+  render: (args) => {
+    return (
+      <HvSelect
+        required
```
MEsteves22, line 38: When required, the border is not red if I click on the dropdown and close it without selecting anything. The error message doesn't appear also. Regarding the border, we have this for the other form elements. Was this intentional? I noticed `disabledBorder` was added to the warning component.

packages/core/src/Select/Select.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+          <HvOption value="pt">Portugal</HvOption>
+          <HvOption value="pl">Poland</HvOption>
+          <HvOption value="sp">Spain</HvOption>
+        </HvOptionGroup>
+      </HvSelect>
+    );
+  },
+};
+
+export const Variants: StoryObj<HvSelectProps<{}, false>> = {
```
MEsteves22, line 61: For some reason, I can't close the dropdown by clicking outside in this sample. I can only close it by clicking on the dropdown again.

packages/core/src/Select/Select.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+          onChange={(evt, val) => setSelection(val)}
+          onOpenChange={setOpen}
+        >
+          {options.map(({ value, label, flag }) => (
+            <HvOption key={value} value={value} label={label}>
+              {`${flag} ${label}`}
+            </HvOption>
+          ))}
+        </HvSelect>
+        <HvButton variant="secondarySubtle" onClick={() => setOpen((o) => !o)}>
```
MEsteves22, line 169: There's a weird behaviour here: if a click on the button to close the dropdown, it closes and opens again.

packages/core/src/Select/Option.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export type HvOptionClasses = ExtractNames<typeof useClasses>;
+
+export interface HvOptionProps<OptionValue extends {}>
+  extends Omit<HvListItemProps, "value" | "disabled">,
+    Pick<OptionOwnProps<OptionValue>, "disabled" | "label" | "value"> {
+  classes?: HvOptionClasses;
+}
+
+export const HvOption = fixedForwardRef(function HvOption<
```
MEsteves22, line 29: I was wondering if it wouldn't be better to use `HvSelectOption` instead of `HvOption` to keep the more generic `HvOption` name free in case we need it in the future. Same thing for the `HvOptionGroup`. Not sure about this one, it was just a thought.

packages/core/src/Select/SelectButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+        [classes.disabled]: disabled,
+        [classes.readOnly]: readOnly,
+      })}
+      classes={{ endIcon: classes.arrowContainer }}
+      endIcon={endIcon}
+      {...others}
+    >
+      <div className={classes.selection}>
+        {children && typeof children === "string" ? (
+          <div className={cx(classes.placeholder)}>{children}</div>
```
MEsteves22, line 110: `cx` not needed

packages/core/src/Select/SelectButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+    color: "inherit",
+    flex: 1,
+    textAlign: "start",
+
+    overflow: "clip visible",
+    textOverflow: "ellipsis",
+    whiteSpace: "nowrap",
+  },
+  placeholder: {},
+  arrowContainer: {
```
MEsteves22, line 44: In DS3 the dropdown icon is almost going out of the dropdown:

Screenshot 2024-02-16 at 09 20 34

MEsteves22 (commented):
Really nice 👌🏻

packages/core/src/Select/Option.tsx
zettca: I thought of that too. I ended up going with `HvOption` because it's shorter and it maps _directly_ to the HTML element `<option>`. Similarly for `HvOptionGroup` and `<optgroup>` (maybe we could call it `HvOptGroup` instead) 🤔

But I don't have a strong opinion on this.

packages/core/src/Select/SelectButton.tsx
zettca: Changed styles 👍

packages/core/src/Select/Select.stories.tsx
zettca: I aligned the behaviour with the native Select. If there are no items/`children`, it shouldn't open
But I've also improved the sample to have an item to better display the variants

packages/core/src/Select/Select.stories.tsx
zettca: The sample is both controlled and uncontrolled, that's why. I'll remove the button, but keep it controlled.
I'll also move this sample to a separate file to `?raw` load the source code

MEsteves22 (approved):

PR #4031, opened 2024-02-14
docs(Table): table filter sample added

MEsteves22:
Table filter sample added:
- As discussed, the filter group was used instead of opening a dialog with a new filter.

### DIFF
--- a/packages/core/src/Table/stories/TableSection/TableFilter.tsx
+++ b/packages/core/src/Table/stories/TableSection/TableFilter.tsx
@@ -1,0 +1,329 @@
+import { useMemo, useRef, useState } from "react";
+import { css, keyframes } from "@emotion/css";
+import {
+  HvButton,
+  HvCheckBox,
+  HvFilterGroup,
+  HvFilterGroupProps,
+  HvFilterGroupValue,
+  HvInput,
+  HvPagination,
+  HvTable,
+  HvTableBody,
+  HvTableCell,
+  HvTableColumnConfig,
+  HvTableContainer,
+  HvTableHead,
+  HvTableHeader,
+  HvTableRow,
+  HvTableSection,
+  HvTypography,
+  baseDropdownClasses,
+  theme,
+  useHvData,
+  useHvFilters,
+  useHvGlobalFilter,
+  useHvPagination,
+} from "@hitachivantara/uikit-react-core";
+import { Add, Close, Filters } from "@hitachivantara/uikit-react-icons";
+
+import { AssetEvent, getColumns, makeData, range } from "../storiesUtils";
+
+const EmptyRow = () => (
+  <HvTableRow>
+    <HvTableCell colSpan={100} />
+  </HvTableRow>
+);
+
+const slide = keyframes`
+  0% { 
+    max-height: 0;
+  }
+  100% { 
+    max-height: 300px;
+  }
+`;
+
+const classes = {
+  filtersContainer: css({
+    display: "flex",
+    width: "100%",
+    backgroundColor: theme.colors.warning_20,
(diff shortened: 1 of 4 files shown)


packages/core/src/Table/stories/TableSection/TableFilter.tsx
```diff
@@ -0,0 +1,338 @@
...
+      { id: "low", name: "Low" },
+    ],
+  },
+];
+
+export const TableFilter = () => {
+  const filterRef = useRef<HTMLDivElement>(null);
+
+  const [selectedFilters, setSelectedFilters] = useState<HvFilterGroupValue>();
+  const [checked, setChecked] = useState(false);
```
zettca, line 131: I was half-expecting this to be turned on by default.
Using `HvFilterGroup` with an "AND" filtering logic is a bit strange to me, as selecting more than one checkbox (in a group) would always yield no results 🤔

packages/core/src/Table/stories/TableSection/TableSection.stories.mdx
```diff
@@ -67,3 +68,19 @@ the section content.
 <Canvas>
   <Story name="Editable Table" story={EditableStory} />
 </Canvas>
+
+### Table Filter
```
zettca, line 72: One one hand, I would expect a sample that's _all about_ filtering to be under `useHvFilters`... on the other hand it does contain an `HvTableSection`... 🤷

zettca (approved):
LGTM 👍 

Just some small (optional) notes:

packages/core/src/Table/stories/TableSection/TableFilter.tsx
```diff
@@ -0,0 +1,338 @@
...
+      }));
+    }
+
+    return cols;
+  }, [checked]);
+
+  const processedFilters = useMemo(
+    () =>
+      selectedFilters
+        ?.map(
```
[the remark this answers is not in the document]
MEsteves22: I'll update to use it

packages/core/src/Table/stories/TableSection/TableFilter.tsx
```diff
@@ -0,0 +1,338 @@
...
+              onChange={(e, v) => setGlobalFilter?.(v)}
+            />
+            <HvFilterGroup
+              ref={filterRef}
+              filters={filters}
+              disablePortal={false}
+              value={selectedFilters}
+              onChange={(event, value) => {
+                setSelectedFilters(value);
+                handleFilters(value);
```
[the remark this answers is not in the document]
MEsteves22: True, I'll used it inside the handler directly

packages/core/src/Table/stories/TableSection/TableFilter.tsx
MEsteves22: I used the AND behaviour has the default because that's the default react table behaviour when using `setAllFilters`. But I agree that it's a weird behaviour, that's why I added the OR one. I can change the sample and default to OR.

packages/core/src/Table/stories/TableSection/TableSection.stories.mdx
MEsteves22: I understand. I can add a link in the `useHvFilters` page to let the users know we have this sample. It just seems a bit weird (to me) to have this sample with a completely different styling inside `useHvFilters`.

MEsteves22:
I'll be updating this PR with the comments 🕣

MEsteves22:
Updated ✅

plagoa (approved):

plagoa:
Looks very nice! 🚀

PR #4020, opened 2024-02-07
docs: add Drag and Drop samples

plagoa:

PR #4033, opened 2024-02-16
docs(storybook): move to .mdx files

zettca:
Moves MDX docs from `.stories.mdx` to the standard `.mdx`

- Storybook support for `.stories.mdx` files is deprecated and dropped in Storybook 8
- We're declaring stories in non `.stories.tsx` files which is a hack and isn't supported
  - We use `ComponentStories.tsx` instead of `Component.stories.tsx`

This was part of the cause for Applitools tests to fail in some PRs (eg. https://github.com/lumada-design/hv-uikit-react/pull/4020)

Also includes some small changes in `docs/*` samples:
- Add height to viz samples
- Merge theme `component` override props
- Remove the `styled` with dynamic props sample as [it's not a good practice](https://emotion.sh/docs/best-practices#use-the-style-prop-for-dynamic-styles)

Please rebase, don't squash the commits or we'll lose the docs history. 
Review the commits separately - it's much easier


docs/guides/provider/Provider.mdx
```diff
@@ -0,0 +1,34 @@
...
+are used to specify the active theme and color mode. For more information, please read the [**theming**](./?path=/docs/guides-theming--docs#theming) documentation.
+
+If several `HvProvider`'s are used, you'll need to create different root elements, set the `rootElementId` property for each one of the providers, and scope the
+theme styles to the root element with the `cssTheme` property.
+Otherwise, the styling will not work properly. Please, find an example below.
+
+**UI Kit components will not work properly if the `HvProvider` is not configured correctly** since they will not be able to access the properties of the active theme.
+
+<Canvas of={ProviderStories.Main} />
+<ArgsTable of={ProviderStories.Main} />
```
MEsteves22, line 34: The args table is not working:

```suggestion
<ArgsTable of={ProviderStories.Main} />
<ArgsTable of={HvProvider} />
```

MEsteves22 (approved):
Looks good 👌🏻

I just noticed the following:

PR #4036, opened 2024-02-19
fix(Pagination): fix page size dropdown styles

plagoa:
Applitools errors are expected.

### DIFF
--- a/packages/core/src/Table/stories/TableSection/CompleteTableSection.tsx
+++ b/packages/core/src/Table/stories/TableSection/CompleteTableSection.tsx
@@ -132,7 +132,10 @@ export const CompleteTableSection = () => {
         </HvTable>
       </HvTableContainer>
       {page?.length ? (
-        <HvPagination {...getHvPaginationProps?.()} />
+        <HvPagination
+          {...getHvPaginationProps?.()}
+          labels={{ pageSizePrev: "", pageSizeEntryName: `of ${data.length}` }}
+        />
       ) : undefined}
     </HvTableSection>
   );
--- a/packages/core/src/Table/stories/TableSection/TableEditable.tsx
+++ b/packages/core/src/Table/stories/TableSection/TableEditable.tsx
@@ -129,6 +129,7 @@ interface TableProps<T extends Data> {
   loading?: boolean;
   pageCount: number;
   columns: HvTableColumnConfig<T, string>[];
+  totalRecords?: number;
   onRowRestore?: (id: string) => Promise<void>;
   onRowDelete?: (id: string) => Promise<void>;
   onRowAdd?: (row: Partial<T>) => Promise<void>;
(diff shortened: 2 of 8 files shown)


packages/core/src/Table/stories/TableSection/CompleteTableSection.tsx
```diff
@@ -32,8 +32,9 @@ const EmptyRow = () => (
 );
 
 export const CompleteTableSection = () => {
+  const totalRows = 64;
   const columns = useMemo(() => getColumns(), []);
-  const [data, setData] = useState(makeData(64));
+  const [data, setData] = useState(makeData(totalRows));
```
zettca, line 37: We should us `data.length`, not `totalRows`. Rows could be added or removed
```suggestion
  const [data, setData] = useState(makeData(64));
```

packages/core/src/TableSection/TableSection.styles.tsx
```diff
@@ -85,6 +85,12 @@ export const { staticClasses, useClasses } = createClasses("HvTableSection", {
     [`& .${paginationClasses.pageSizeOptions}`]: {
       left: theme.space.sm,
     },
+    [`& .${paginationClasses.pageSizeHeader}`]: {
+      border: "none",
```
zettca, line 89: Shouldn't we also remove the border of the popup content? This looks strange...
image

packages/core/src/Table/stories/TableSection/CompleteTableSection.tsx
```diff
@@ -132,7 +133,10 @@ export const CompleteTableSection = () => {
         </HvTable>
       </HvTableContainer>
       {page?.length ? (
-        <HvPagination {...getHvPaginationProps?.()} />
+        <HvPagination
```
MEsteves22, line 136: I think we should add this new pagination to all samples in the table section page

packages/core/src/Table/stories/TableSection/TableEditable.tsx
```diff
@@ -585,7 +585,13 @@ const Table = <T extends Data>({
         </HvTableContainer>
       </LoadingContainer>
       {page.length > 0 ? (
-        <HvPagination {...getHvPaginationProps?.()} />
+        <HvPagination
+          {...getHvPaginationProps?.()}
+          labels={{
+            pageSizePrev: "",
+            pageSizeEntryName: `of ${(data && data.length * pageCount) || 0}`,
```
MEsteves22, line 592: Sometimes the page count is wrong (for example if you go to the last page, the page count is 900)

```suggestion
            pageSizeEntryName: `of ${(totalRecords}`,
```

You can get the `totalRecords` from `useServerData`. You'll  need to pass it to the `Table` component as a prop.

MEsteves22 (approved):
🚀

I just have this last comment before merging since the page count is wrong for the editable table:

packages/core/src/Table/stories/TableSection/TableEditable.tsx
plagoa: Got it, forgot about that. Updated.

HQFOX (approved):

PR #4042, opened 2024-02-21
feat(SplitButton): add split button variant to `MultiButton` component

HQFOX:
- A style was removed from the `HvMultiButton` and added to `ds3` theme where the selected button would be slightly bigger then the others to match the ds5 specs. Applitools is expected to break because of it.
- Button sizes was added to `HvMultiButton`
- `HvMultiButton` variants style were added. The prop was already there but the styles didn't match ( there was no ds specification so I added the styles that made sense looking at the button variants specs)
- changed the `type` `css` selectors with classes for compatibility  with the `HvDropdownMenu` component (e.g. replaced `first-of-type` with the `firstButton` class)
- Added the `sizes` prop from the `HvButton` to the `HvDropdownMenu` component to create different size split buttons
- Added diferent icon sizes styles to `HvButton` according ds5 to match iconless button height

### DIFF
--- a/packages/core/src/Button/Button.tsx
+++ b/packages/core/src/Button/Button.tsx
@@ -106,6 +106,7 @@ export const HvButton = fixedForwardRef(function HvButton<
     focusableWhenDisabled,
     onClick: onClickProp,
     onMouseDown: onMouseDownProp,
+    selected,
     ...others
   } = useDefaultProps("HvButton", props);
   const { classes, css, cx } = useClasses(classesProp);
@@ -149,6 +150,7 @@ export const HvButton = fixedForwardRef(function HvButton<
         tabIndex: focusableWhenDisabled ? 0 : -1,
         "aria-disabled": true,
       })}
+      {...(selected && { "aria-pressed": selected })}
       {...others}
     >
       {startIcon && <span className={classes.startIcon}>{startIcon}</span>}
--- a/packages/core/src/Button/Button.tsx
+++ b/packages/core/src/Button/Button.tsx
@@ -12,6 +12,7 @@ import {
   getOverrideColors,
   getRadiusStyles,
   getSizeStyles,
+  getIconSizeStyles,
   useClasses,
 } from "./Button.styles";
 import { HvButtonRadius, HvButtonSize, HvButtonVariant } from "./types";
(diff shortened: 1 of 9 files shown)


francisco-guilherme (approved):

packages/core/src/Button/Button.tsx
```diff
@@ -130,13 +132,14 @@ export const HvButton = fixedForwardRef(function HvButton<
...
         classes[variant],
-        size && css(getSizeStyles(size)),
+        size && !icon && css(getSizeStyles(size)),
         radius && css(getRadiusStyles(radius)),
         overrideIconColors && css(getOverrideColors()),
         {
           [classes.icon]: icon,
           [classes.disabled]: disabled,
         },
+        size && icon && css(getIconSizeStyles(size)),
```
zettca, line 142: Is this needed?
The purpose of the `icon` prop is so that the Button's styles are unset and the icon (`children`) can decide the size of the button. The `size` styles come before `classes.icon` because of that.
If we need a larger icon button, we should increase the icon size
Am I missing something? 🤔

packages/core/src/MultiButton/MultiButton.stories.tsx
```diff
@@ -408,3 +417,223 @@ export const VerticalOrientation: StoryObj<HvMultiButtonProps> = {
...
+    ];
+
+    const [selectedOption, setSelectedOption] = useState(options[0]);
+
+    return (
+      <HvSimpleGrid cols={3} spacing="sm">
+        <HvMultiButton variant="primary" style={{ width: "232px" }} split>
+          <HvButton>{selectedOption.label}</HvButton>
+          <HvDropDownMenu
+            dataList={options}
```
zettca, line 443: We should probably do a `keepOpened={false}` - keeping it open is strange and I'd say unexpected

zettca (commented):
The sample looks pretty simple, which is great 👍 

The styles of the opened button look off, we need to review this:
image

packages/core/src/Button/Button.tsx
HQFOX: I made this change so that it's possible for a icon button to have different sizes and still have the 'squarish' look like in the [specs](https://designsystem.hitachivantara.com/6c705d900/p/73880a-button/b/50959e) even when the icon is smaller. It was also useful for integrating an icon button like the dropmenu with different sizes in a `Multibutton` and have them matching the height. But let me know if you see a better way to do this

HQFOX:
> The sample looks pretty simple, which is great 👍
> 
> The styles of the opened button look off, we need to review this: image

@zettca do you mean the panel border? maybe i can talk to someone from DS or open another issue to review this

francisco-guilherme (approved):
Lets merge as it is and improve it once we get feedback from DS

PR #4051, opened 2024-03-04
feat(flow): improve flow groups handling

francisco-guilherme:


packages/lab/src/Flow/stories/Usage.mdx
```diff
@@ -132,17 +132,17 @@ The parameters and the list of inputs and outputs are passed as a property of th
 Find an example of a custom node below.
```
MEsteves22, line 132: I think we also need to review the "Sidebar" and "Custom nodes and node types" sections since they are not up to date with these changes.

packages/lab/src/Flow/stories/Usage.mdx
```diff
@@ -132,17 +132,17 @@ The parameters and the list of inputs and outputs are passed as a property of th
...
-  HvFlowNodeTypeMeta,
-} from "@hitachivantara/uikit-react-lab";
+import { HvFlowNode, HvFlowNodeFC } from "@hitachivantara/uikit-react-lab";
 
 export const CustomNode: HvFlowNodeFC = (props) => {
   return (
     <HvFlowNode
       description="CustomNode description"
       expanded
+      defaults={{
```
MEsteves22, line 142: `defaults` is not a prop of `HvFlowNode`:

```suggestion
      group="someNodeGroup"
      title="CustomNode"
```

packages/lab/src/Flow/Sidebar/Sidebar.tsx
```diff
@@ -40,6 +39,8 @@ export interface HvFlowSidebarProps
   title?: string;
   /** Sidebar description. */
   description?: string;
+  /** Flatten sidebar items */
+  flatten?: boolean;
```
MEsteves22, line 43: `defaultGroupProps` should be removed for the `HvFlowSidebarProps` since it's no longer needed

packages/lab/src/Flow/types/flow.ts
```diff
@@ -8,57 +8,39 @@ import { HvColorAny } from "@hitachivantara/uikit-styles";
...
+  | HvFlowNodeComponentClass
+  | HvFlowNodeFC<NodeData>;
+
+export type HvFlowNodeTypes<NodeData = any> = Record<
+  string,
+  HvFlowNodeComponentType<NodeData>
+>;
+
+export type HvFlowGroupItem = {
+  type: string;
```
MEsteves22, line 26: It took me quite some type to understand that this was the `nodeType`. Should we change the name to `nodeType`?  Maybe it's only a matter of adding some docs or a me problem 😅

packages/lab/src/Flow/Sidebar/Sidebar.tsx
```diff
@@ -190,40 +173,35 @@ export const HvFlowSidebar = ({
...
-                    itemProps={{
-                      "aria-roledescription": labels?.itemAriaRoleDescription,
-                    }}
-                    {...obj[1]}
+          <ul id={groupsElementId} className={classes.groupsContainer}>
+            {Object.entries(groups).map((obj) => {
+              if (flatten) {
+                return obj[1].items?.map((item) => (
+                  <HvFlowDraggableSidebarGroupItem
+                    key={item.label}
```
MEsteves22, line 181: Should we add an `id` prop to the items to avoid using the label as `id` and `key`? Maybe we make the items as `Record<string, HvFlowGroupItem>` instead of `HvFlowGroupItem[]` for simplicity and to be sure that the `id`'s are not repeated.

I'm also suggesting this change because we could add a prop to the `HvFlowNode` like `itemId` to get the `subtitle` directly from the node group's `items` (instead of providing it twice: once in the node groups definition and another time in the node).

MEsteves22 (commented):
Really great stuff 👌🏻💪

I left some comments and suggestions:

MEsteves22:
Updated with review as requested ✅

MEsteves22 (approved):

plagoa (approved):

packages/lab/src/Flow/types/flow.ts
```diff
@@ -8,57 +8,40 @@ import { HvColorAny } from "@hitachivantara/uikit-styles";
...
+  data?: unknown;
+}
 
 /** Node groups */
 export interface HvFlowNodeGroup {
   label: string;
   description?: string;
   color?: HvColorAny;
   icon?: React.ReactNode;
+  items?: Record<string, HvFlowGroupItem>;
```
zettca, line 38: why not
```suggestion
  items?: HvFlowGroupItem[];
```

aren't the object's `key` and `.id` the same?

packages/lab/src/Flow/types/flow.ts
```diff
@@ -8,57 +8,40 @@ import { HvColorAny } from "@hitachivantara/uikit-styles";
...
+export type HvFlowNodeComponentType<NodeData = any> =
+  | HvFlowNodeComponentClass
+  | HvFlowNodeFC<NodeData>;
+
+export type HvFlowNodeTypes<NodeData = any> = Record<
+  string,
+  HvFlowNodeComponentType<NodeData>
+>;
+
+export interface HvFlowGroupItem {
```
zettca, line 25: Shouldn't we be able to configure other Node properties? eg. subtitle, params, etc?

packages/lab/src/Flow/types/flow.ts
MEsteves22: I added the key to have a unique id for the item

packages/lab/src/Flow/Node/Node.tsx
```diff
@@ -50,12 +43,16 @@ export interface HvFlowNodeProps<T = any> extends HvFlowBaseNodeProps<T> {
   actionsIconOnly?: HvActionsGenericProps["iconOnly"];
   /** Node maximum number of actions visible. */
   maxVisibleActions?: HvActionsGenericProps["maxVisibleActions"];
+  /** Node subtitle */
+  subtitle?: string;
+  /** Node group ID */
+  group?: string;
+  /** Node group item ID */
+  groupItem?: string;
```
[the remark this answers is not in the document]
MEsteves22: Agree 👍

packages/lab/src/Flow/types/flow.ts
MEsteves22: We already get the subtitle from the item (line 95, `Node.tsx`):

```
const nodeSubtitle = (groupItem && nodeGroup?.items?.[groupItem]?.label) || subtitle;
```

As for adding `params`, is there an advantage to adding it to the item structure that I'm not seeing? If not, I think we should leave the node's configuration at the node level. We have `data` because we need to add it to the node when we are dropping the node on the canvas.

zettca:
moving to draft as we're still ironing out some issues with dynamic nodes

packages/lab/src/Flow/types/flow.ts
zettca: What if nodes different "node configurations" need different `params`?
DA actually was doing this at the config level (though avoidable)

zettca: Label/Subtitle is addressed through `data.customNodeLabel`

apps/app/src/pages/Flow/config.tsx
```diff
@@ -1,53 +1,124 @@
...
   kpi: Kpi,
   donutChart: DonutChart,
   table: Table,
 } satisfies HvFlowProps["nodeTypes"];
 
-export type NodeType = keyof typeof baseNodeTypes;
+export type NodeType = keyof typeof nodeTypes;
 
 /** Initial Flow */
 export const nodes = [
```
MEsteves22, line 124: When we click on "Preview", the charts don't appear. We need to update line 160 in this file to `endpoint: "steelwheels1"` since the id changed.
Also, when we look at the nodes, only the node "Dataset" has a "subtitle". We should add the `nodeLabel` property to the other nodes too.

packages/lab/src/Flow/Node/Node.tsx
```diff
@@ -53,12 +45,16 @@ export interface HvFlowNodeProps<T = any> extends HvFlowBaseNodeProps<T> {
   actionsIconOnly?: HvActionsGenericProps["iconOnly"];
   /** Node maximum number of actions visible. */
   maxVisibleActions?: HvActionsGenericProps["maxVisibleActions"];
+  /** Node subtitle - this is typically the node "name" */
+  subtitle?: string;
+  /** Node group ID */
+  groupId?: string;
+  /** Node ID to match with the `nodeGroups`'s item */
```
MEsteves22, line 52: I don't know if I'm missing something, but the `nodeId` doesn't seem to be used anywhere. We should remove it.

packages/lab/src/Flow/stories/Visualizations/Filter.tsx
```diff
@@ -8,19 +8,17 @@ import {
...
 
-import type { NodeGroup } from ".";
 import { NodeData } from "./data";
 
 function filterDataByCountries(data, countriesToFilter: string[]) {
   return data.filter((item) => countriesToFilter.includes(item.country));
 }
 
-export const Filter: HvFlowNodeFC<NodeGroup> = (props) => {
+export const Filter: HvFlowNodeFC = (props) => {
```
MEsteves22, line 21: We need to update line 35 for this node by spreading the previous data because when we check an option, the node title disappears.

MEsteves22 (approved):
👌🏻🚀

packages/lab/src/Flow/Node/Node.tsx
zettca: it's not 👍

MEsteves22 (approved):

PR #4057, opened 2024-03-06
refactor: remove lodash & bump to ES2022 syntax

zettca:
- Bump the syntax to the ES2022 baseline
- Remove usage of `lodash` in samples, so that users don't need to install it
- Inline `range` (Array(n)) utility in table samples not to depend on external utils
- Remove usage of `lodash` in components, to reduce the bundle size
  - even the simplest lodash utilities (eg. [range](https://bundlejs.com/?q=lodash%2Frange&treeshake=%5B*%5D&text=%22export+%7Bfn%7D%22&config=%7B%22analysis%22%3Atrue%7D)) require lots of other internal utilities, bloating the final bundles of even the simplest components (eg. `HvLoading`)

### DIFF
--- a/apps/app/src/generator/GeneratorContext.tsx
+++ b/apps/app/src/generator/GeneratorContext.tsx
@@ -12,8 +12,11 @@ import {
   DeepPartial,
   HvTheme,
 } from "@hitachivantara/uikit-react-core";
-import { HvBaseTheme, HvThemeStructure } from "@hitachivantara/uikit-styles";
-import merge from "lodash/merge";
+import {
+  HvBaseTheme,
+  HvThemeStructure,
+  mergeTheme,
+} from "@hitachivantara/uikit-styles";
 
 import { themeDiff } from "./utils";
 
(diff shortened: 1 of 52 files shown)


apps/app/src/generator/GeneratorContext.tsx
```diff
@@ -86,14 +89,13 @@ const GeneratorProvider = ({ children }: { children: React.ReactNode }) => {
         if (!isBaseChange) {
           if (!isCodeEdit) {
             const diff = themeDiff(prev, changes);
-            return merge({}, prev, diff);
+            return mergeTheme(prev, diff);
```
MEsteves22, line 92: This doesn't seem to be working as expected: when I try to update a color and then another, the first one is removed from the custom theme

MEsteves22 (commented):
👌🏻

I just noticed the following:

apps/app/src/generator/GeneratorContext.tsx
zettca: `mergeTheme` was squashing the previous sub-object; I've added a fix

MEsteves22 (approved):

francisco-guilherme (approved):

PR #4064, opened 2024-03-08
feat: new button variants and sizes & captionLabel typography added

MEsteves22:
- Button aligned with DS: `disabled` style, `ds3` fixes, and `size` style
   - ⚠️ `xs` and `xl` are not DS sizes for the button. I kept them to not introduce breaking changes. 
   - ⚠️ `theme.sizes` was not used since the sizes don't match with the ones needed for the button. I did not change the sizes to not introduce breaking changes.
- Context colors added for the button: new variants were added for warning, negative, and positive
- `captionLabel` typography added since it was needed for the button `sm` size

ℹ️ Applitools changes expected


packages/core/src/Button/Button.styles.ts
```diff
@@ -46,8 +40,8 @@ export const { staticClasses, useClasses } = createClasses("HvButton", {
   focusVisible: {},
   disabled: {
     color: theme.colors.secondary_60,
-    borderColor: theme.colors.atmo4,
-    backgroundColor: theme.colors.atmo3,
+    border: "none",
+    backgroundColor: "transparent",
```
zettca, line 55: If disabled buttons always have a transparent `backgroundColor`, we should be able to increase the specificity here, instead in the `root` and most variant styles, right?

packages/core/src/Button/Button.styles.ts
```diff
@@ -67,27 +72,70 @@ export const { staticClasses, useClasses } = createClasses("HvButton", {
...
+    "&:not($disabled)": {
+      "&:hover, &:focus-visible": {
+        backgroundColor: theme.colors.positive_80,
+      },
     },
   },
+  positiveSubtle: {
+    color: theme.colors.positive,
+    ...subtle,
+  },
```
zettca, line 93: There is lots of duplication here (that we can remove) - and users will get all of this in their bundles whether they use these new variants of not...

My suggestions are:
- Keep the disabled "overrides" in the disabled class (increase `backgroundColor` specificity)
- Have classes for the button types (eg. `subtle`/`ghost`/`contained`?) and apply the styles there
- Do the "coloring" (`color` + `backgroundColor`) with the `getX` utilities like we're doing for the sizes

zettca (commented):
Nice work 👌 

I think we should review the styles/classes duplication. The user shouldn't get a significantly larger `HvButton` bundle for classes they're unlikely to use

MEsteves22:
Updated with comments ✅

packages/core/src/Button/Button.styles.ts
```diff
@@ -45,82 +38,105 @@ export const { staticClasses, useClasses } = createClasses("HvButton", {
...
+    borderColor: "transparent",
+    backgroundColor: "transparent",
+    "&:hover": {
+      backgroundColor: "transparent",
+      borderColor: "transparent",
+    },
+    "&:focus-visible": {
+      backgroundColor: "transparent",
+      borderColor: "transparent",
+    },
```
zettca, line 52: Can't we merge these too? 👀 

```suggestion
    "&:hover, &:focus-visible": {
      backgroundColor: "transparent",
      borderColor: "transparent",
    },
```

zettca (approved):
Nice 💯

packages/core/src/Button/Button.styles.ts
MEsteves22: Of course, I'll update the styles 🙏

MEsteves22: Done ✅

packages/core/src/Button/Button.stories.tsx
```diff
@@ -141,52 +192,34 @@ export const Icons: StoryObj<HvButtonProps> = {
     <>
       <div>
         <HvButton icon aria-label="Play" variant="primaryGhost">
-          <Play iconSize="M" />
+          <Play />
```
zettca, line 195: I just noticed... these were the only samples testing that `HvButton` "behaves well" with larger content. The remaining samples are all "sm" icons and/or normal-sized text.
Maybe there's value in keeping something like this, what do you think? 🤔

MEsteves22: I remove this because DS specs always use icons with 32px but I'll add the icon sizes back to test how the button behaves

MEsteves22: ✅

PR #4065, opened 2024-03-11
feat(Skeleton): add Skeleton component

plagoa:
- Add the new component `HvSkeleton`
- Update the `AssetInventory` template to use this new component

### DIFF
--- a/packages/core/src/Skeleton/Skeleton.stories.tsx
+++ b/packages/core/src/Skeleton/Skeleton.stories.tsx
@@ -1,0 +1,226 @@
+import { useState } from "react";
+import { Meta, StoryObj } from "@storybook/react";
+import {
+  HvActionBar,
+  HvActionsGeneric,
+  HvAvatar,
+  HvCard,
+  HvCardContent,
+  HvCardHeader,
+  HvCardMedia,
+  HvCheckBox,
+  HvRadio,
+  HvRadioGroup,
+  HvSkeleton,
+  HvSkeletonProps,
+  HvSwitch,
+  HvTypography,
+  theme,
+} from "@hitachivantara/uikit-react-core";
+
+import { CustomSkeleton } from "./CustomSkeleton";
+import CustomSkeletonRaw from "./CustomSkeleton?raw";
+
+const meta: Meta<typeof HvSkeleton> = {
+  title: "Components/Skeleton",
+  component: HvSkeleton,
+};
+export default meta;
+
+export const Main: StoryObj<HvSkeletonProps> = {
+  args: {
+    width: 150,
+    hidden: false,
+  },
+  argTypes: {
+    classes: { control: { disable: true } },
+  },
+  render: ({ hidden, ...args }) => {
+    return (
+      <HvTypography>
+        {!hidden ? <HvSkeleton {...args} /> : "This is my text"}
+      </HvTypography>
+    );
+  },
+};
+
+export const Variants: StoryObj<HvSkeletonProps> = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "The `HvSkeleton` component can have different variants to represent different shapes and sizes.",
+      },
+    },
+  },
+  render: () => {
+    return (
+      <div
+        style={{
+          display: "flex",
+          flexDirection: "column",
+          gap: theme.space.sm,
+        }}
+      >
+        <HvSkeleton variant="square" width={100} animation="wave">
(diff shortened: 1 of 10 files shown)


MEsteves22 (commented):
👌🏻

packages/core/src/Skeleton/Skeleton.stories.tsx
```diff
@@ -0,0 +1,299 @@
...
+      description: {
+        story:
+          "The `HvSkeleton` component can be customized with a custom animation if none of the default animations fit your needs.",
+      },
+      source: {
+        code: `
+const myAnimation = keyframes\`
+  0%, 100% {
+    box-shadow: inset 0px 0px 60px 0px rgba(0,0,0,0.50);
+  }
```
zettca, line 245: Why? Let's use the `?raw` import instead!

zettca (changes requested):

templates/AssetInventory/CardView/CardView.tsx
```diff
@@ -10,17 +10,23 @@ import {
...
   instance: HvTableInstance<AssetInventoryEntry, string>;
+  loading?: boolean;
 }
 
-export const CardView = ({ id, instance }: CarViewProps) => {
+export const CardView = ({ id, instance, loading }: CarViewProps) => {
   const selectedCardsIds = instance.selectedFlatRows.map((r) => r.id);
+  const mode = "wave";
+
+  console.log(instance.page);
```
MEsteves22, line 28: `console.log`

templates/AssetInventory/CardView/CardView.tsx
```diff
@@ -44,7 +50,13 @@ export const CardView = ({ id, instance }: CarViewProps) => {
             statusColor={row?.original?.statusColor}
             icon={getStatusIcon(row?.original?.statusColor)}
           >
-            <HvCardHeader title={row?.values?.name} />
+            <HvCardHeader
```
MEsteves22, line 53: We already have the `LoadingSkeleton` component so we don't need these changes in card view to use `HvSkeleton`

templates/AssetInventory/index.tsx
```diff
@@ -124,33 +124,36 @@ const AssetInventory = () => {
...
-                "aria-controls": `${idsToControl.cards} ${idsToControl.list}`,
-              },
-            }}
-          />
-          <HvRightControl
-            values={rightControlValues}
-            sortProps={{
-              "aria-label": "Sort by",
+    <div className={classes.root}>
+      <HvControls
```
MEsteves22, line 128: I think we should disable the controls while it's loading to prevent the users from switching to the table view. Otherwise we get card skeletons when the view is a table.

MEsteves22 (approved):
LGTM 🙏

I just noticed the following:

templates/AssetInventory/index.tsx
plagoa: There's really no way to disable it without making changes to the `HvControls` component, which I don't think fall into the context of the work. I opted to hide the view switcher using the available prop on the component, which I think achieves the same goal.

MEsteves22: Makes sense to me 🙏

MEsteves22:
I just remembered we are missing some tests for this new component 🙈

plagoa:
> I just remembered we are missing some tests for this new component 🙈

Just added the tests :)

MEsteves22 (approved):

packages/core/src/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,74 @@
...
+      )}
+      style={{
+        width,
+        height,
+        ...(backgroundImage && {
+          backgroundImage: `url(${backgroundImage})`,
+          backgroundSize: "contain",
+          backgroundPosition: "center",
+          backgroundRepeat: "no-repeat",
+        }),
```
zettca, line 67: The user could easily configure this on their side
```tsx
background: `url(${img}) center/contain no-repeat`
```

packages/core/src/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }}
```
zettca, line 68: We need to spread user-defined styles too
```suggestion
      ...style
      }}
```

zettca (approved):
Looks great, nice work 💪

PR #4068, opened 2024-03-12
feat: allow run custom validations on Wizard before going to next page. DIG-1348

rubenbarroshv:
In the Digital Advisor, we need to use the Wizard but we need to do some HTTP requests to validate that the user can go to the next page. We added a callback function to the WizardActions to allow for that.

Note: This will only work when we use the Wizard with separate components, since we don't have the Context to change the page without it.

I also made a change to allow us to define the action labels as a ReactNode so that we can put a Loading while we are performing the HTTP request validation.

### DIFF
--- a/packages/lab/src/Wizard/WizardActions/WizardActions.tsx
+++ b/packages/lab/src/Wizard/WizardActions/WizardActions.tsx
@@ -26,15 +26,15 @@ export interface HvWizardActionsProps extends HvBaseProps {
   /** An object containing all the labels for the wizard actions component. */
   labels?: {
     /** Cancel button label. */
-    cancel?: string;
+    cancel?: string | React.ReactNode;
     /** Skip button label. */
-    skip?: string;
+    skip?: string | React.ReactNode;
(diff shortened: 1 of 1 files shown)


packages/lab/src/Wizard/WizardActions/WizardActions.tsx
```diff
@@ -26,22 +26,24 @@ export interface HvWizardActionsProps extends HvBaseProps {
...
+    submit?: string | React.ReactNode;
   };
   /** Whether the submit button is disabled. */
   loading?: boolean;
   /** Enables the skip button. */
   skippable?: boolean;
   /** A Jss Object used to override or extend the styles applied to the empty state Wizard. */
   classes?: HvWizardActionsClasses;
+  /** Function executed instead of default go to next page */
+  handleBeforeNext?: () => void;
```
zettca, line 46: Event handlers should be called `on<Event>`, but let's keep the `handle*` for consistency 👍

zettca (approved):

PR #4073, opened 2024-03-14
chore: remove generated id prefixes

zettca:
remove prefixes from the `useUniqueId`'s generated ids as:
- they make the ids needlessly longer (`hvbasedropdown-:ba:` instead of `:ba:`)
- can lead to confusion when testing as the `hv<ComponentName>` makes it seem like a stable id
- ~remove `setUid` (unused for a long time now); we should use `useUniqueId` instead~

### DIFF
--- a/packages/core/src/hooks/useUniqueId.ts
+++ b/packages/core/src/hooks/useUniqueId.ts
@@ -9,7 +9,13 @@ let count = 0;
 
 export const useUniqueId = (
   deterministicId?: string,
-  idPrefix?: string
+  /**
+   * @deprecated
+   * Users should pick between a fully deterministic or fully generated id
+   * @example
+   * useUniqueId(setId(idPrefix, deterministicId))
+   * */
+  idPrefix?: string,
 ): string => {
   const [id, setId] = React.useState<string | undefined>(useReactId());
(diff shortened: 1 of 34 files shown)


packages/core/src/hooks/useUniqueId.ts
```diff
@@ -7,16 +7,13 @@ import * as React from "react"; // this can be optimized when react 17 support i
 const useReactId = (React as any)["useId".toString()] || (() => undefined);
 let count = 0;
 
-export const useUniqueId = (
-  deterministicId?: string,
-  idPrefix?: string
-): string => {
+export const useUniqueId = (deterministicId?: string) => {
```
MEsteves22, line 10: Since we are exporting this, I don't think we should remove this parameter.

MEsteves22 (approved):
I just have the following concern:

packages/core/src/hooks/useUniqueId.ts
zettca: You're right - doesn't hurt to keep it. I've changed to keep `setUid` and deprecated it instead

HQFOX (approved):

PR #4076, opened 2024-03-14
feat(LoadingContainer): add component

zettca:
Add `HvLoadingContainer` component
- A container component that wraps any element with an `HvLoading` and a background with opacity
- Adjusts to parent's size with just CSS (no libraries)

This component or a variant of it is very frequently used whenever there's backend-data


MEsteves22 (approved):
👌🏻🧹

francisco-guilherme (approved):

Issue #4060, opened 2024-03-07
HvQueryBuilder: onChange event is getting triggered unnecessarily And not able to edit without Ids.


vkumarakella:
### Which UI Kit version is this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

When we are navigating from page A to page B for the first time, the Query builder is expanding with saved details.
But, when we do back and forth of the navigation, the query is triggering `onChange` event and hence QueryBuilder is in `closed` state.

And, without `ids` associated with rule query, we are not able to `edit the fields` in Query builder.
Means: Every object in the rule is being associated with id so as to edit. Please make the ids as optional or remove that dependency to edit the rule query.

### Expected behavior 🤔

`onChange` event should only trigger, when user is trying to edit the fields in query builder. Not on navigating between routes.

Rule Query `should be editable` without having `nested ids` associated in the rule query object.

### Steps to reproduce 🕹

Sandbox link: 

Steps to reproduce:

1. Open Query Builder with saved values.
2. Refresh once or twice, and see that onChange event is getting triggered.
3. Have no ids for the rule Query object, it is not being edited.
4.

### Context 🔦

It is affecting the deverbals.

### Your Environment 🌎

Build tool: (e.g. Vite 4, Webpack 5)
Device & OS: (e.g. Windows 11 laptop, iPhone 13 iOS15)
Browser: (e.g. Chrome 112, Safari 15.4)
Other details (e.g. 4K resolution, ...)

MEsteves22:
Hello 👋

We tackled theses issues in the latest release, [@hitachivantara/uikit-react-core@5.60.0](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.60.0). Please, find an example [here](https://lumada-design.github.io/uikit/master/?path=/docs/widgets-query-builder--docs#controlled) under the "Controlled" sample.  
Let us know we you need anything else 🙏

PR #4082, opened 2024-03-15
feat(QueryBuilder): controlled & optional ids

MEsteves22:
Addresses https://github.com/lumada-design/hv-uikit-react/issues/4060

- The query builder can now be controlled and uncontrolled:
   - `query` deprecated in favour of `defaultValue`
   - `value` added to enable controlled mode
- Ids are now optional to use inside the query: 
   - When not provided the component generates ids automatically.
   - To avoid breaking changes, the ids provided by the user and the automatically generated ones are removed in uncontrolled mode `onChange`. However, in controlled mode, only the automatically generated ids are removed. The ones provided by the user are kept since it doesn't make much sense to remove them internally. 
- Tests and stories updated


zettca (approved):
Nice work 💪 

`onChange` Will likely come back to haunt us until we remove it from the `useEffect` 👻

francisco-guilherme (approved):

PR #4084, opened 2024-03-18
fix: table settings story a11y issues

plagoa:
- Set the Table Settings popper to be open by default
- Set the appropriate `list` and `listitem` roles
- Applitools errors expected

### DIFF
--- a/packages/core/src/Table/stories/TableSection/TableSettings.tsx
+++ b/packages/core/src/Table/stories/TableSection/TableSettings.tsx
@@ -139,6 +139,7 @@ const ColumnItem = ({ item }: { item: Column }) => {
       }}
       {...attributes}
       {...listeners}
+      role="listitem"
     >
       {item.title}
       <Drag />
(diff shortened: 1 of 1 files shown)


VGervasio (approved):

packages/core/src/Table/stories/TableSection/TableSettings.tsx
```diff
@@ -421,7 +422,7 @@ const SettingsDialog = ({
 
 export const TableSettings = () => {
   const tableColumns = useMemo(() => getColumns(), []);
-  const [popperOpen, setPopperOpen] = useState(false);
```
MEsteves22, line 424: Shouldn't we open the popper only for applitools tests?

VGervasio: Don't we need it open to run the a11y checks also? (Not sure if we can also make it open for those checks)

MEsteves22: Right, I forgot about that. Maybe the `play` property can be used for that: https://storybook.js.org/docs/writing-stories/play-function.

zettca: Agreed. We should about having popper/panels opened by default in our samples.
The samples are docs-first - it's strange for the user to see them opened by default, and these "poppers" typically (this one doesn't seem too bad) hyjack the page focus/scroll

MEsteves22 (approved):

PR #4086, opened 2024-03-18
fix: css vars deduplication

plagoa:
Prevent common CSS vars to be duplicated through all theme/mode pairs. 
- `theme / mode` contains only color related CSS vars
- `theme` contains the remaining theme level CSS vars (radii, spaces, sizes, etc).


HQFOX (approved):

zettca (approved):
Nice 💯

PR #4095, opened 2024-03-21
feat: add useHvRowState hook

HQFOX:
The hook was need to create a better Editable Table sample, I separated this from that pr for easier review.
Some broken links were fixed, and some tan stack were replaced with an unofficial link because the official was removed. This might not be a great solution, tell me if you think of something better.

### DIFF
--- a/packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
+++ b/packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
@@ -10,7 +10,7 @@ import * as TableHooksStories from "./TableHooks.stories";
 
 Rows can have a state.
 
-Check [React Table's `useRowState` documentation](https://react-table.tanstack.com/docs/api/useRowSelect) for configuration details.
+Check [React Table's `useRowState` documentation](https://react-table-v7-docs.netlify.app/docs/api/useRowSelect) for configuration details.
 
 Press the edit button to see the state of the row changing.
 
--- a/packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
+++ b/packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
@@ -10,7 +10,7 @@ import * as TableHooksStories from "./TableHooks.stories";
 
 Rows can have a state.
 
-Check [React Table's `useRowState` documentation](https://react-table-v7-docs.netlify.app/docs/api/useRowSelect) for configuration details.
+Check [React Table's `useRowState` documentation](https://react-table-v7-docs.netlify.app/docs/api/useRowState) for configuration details.
 
 Press the edit button to see the state of the row changing.
 
--- a/packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
+++ b/packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
@@ -1,0 +1,17 @@
+import { Meta, Canvas, Story } from "@storybook/addon-docs";
+import * as TableHooksStories from "./TableHooks.stories";
+
+<Meta
+  of={TableHooksStories}
+  title="Visualizations/Table/Table Hooks/useHvRowState"
+/>
+
+# useHvRowState
+
+Rows can have a state.
+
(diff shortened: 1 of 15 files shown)


packages/core/src/Table/stories/TableHooks/UseHvRowState.mdx
```diff
@@ -0,0 +1,17 @@
...
+<Meta
+  of={TableHooksStories}
+  title="Visualizations/Table/Table Hooks/useHvRowState"
+/>
+
+# useHvRowState
+
+Rows can have a state.
+
+Check [React Table's `useRowState` documentation](https://react-table-v7-docs.netlify.app/docs/api/useRowSelect) for configuration details.
```
zettca, line 13: URL is wrong
```suggestion
Check [React Table's `useRowState` documentation](https://react-table-v7-docs.netlify.app/docs/api/useRowState) for configuration details.
```

packages/core/src/Table/stories/TableHooks/UseHvRowState.tsx
```diff
@@ -0,0 +1,93 @@
...
+      ...getColumns(),
+      {
+        id: "edit",
+        Cell: (props) => {
+          const { row, setRowState } = props;
+          return (
+            <HvIconButton
+              title={row.state.isEditing ? "Close" : "Edit"}
+              variant="secondaryGhost"
+              aria-label="edit button"
```
zettca, line 34: Title already does this
```suggestion
```

zettca (approved):
Nice work 👌

There's a typo in the link 👀

packages/core/src/Table/stories/TableHooks/UseHvRowState.tsx
```diff
@@ -0,0 +1,93 @@
...
+    useHvData<AssetEvent, string>({ columns, data }, useHvRowState);
+
+  return (
+    <HvTableContainer>
+      <HvTable {...getTableProps()}>
+        <HvTableHead>
+          {headerGroups.map((headerGroup) => (
+            <HvTableRow {...headerGroup.getHeaderGroupProps()}>
+              {headerGroup.headers.map((col) => (
+                <HvTableHeader {...col.getHeaderProps()}>
```
zettca, line 61: we're gonna need this
```suggestion
                <HvTableHeader {...col.getHeaderProps()} aria-hidden={col.variant === "actions" ? true : undefined}>
```

packages/core/src/Table/stories/TableHooks/UseHvRowState.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  <HvTableCell
+                    {...cell.getCellProps()}
+                    style={row.state?.isEditing ? { maxWidth: 70 } : undefined}
+                  >
+                    {row.state?.isEditing && cell.column.id !== "edit" ? (
+                      <EditableCell value={cell.value} />
+                    ) : (
+                      cell.render("Cell")
+                    )}
+                  </HvTableCell>
```
zettca, line 83: This should be in the `columns`' `Cell:`, not in the table rendering.
Because of this, this sample should have its own `columns` array instead of using `getColumns`.

packages/core/src/Table/stories/TableHooks/UseHvRowState.tsx
```diff
@@ -0,0 +1,123 @@
...
+  HvInput,
+  HvTableColumnConfig,
+  HvIconButton,
+} from "@hitachivantara/uikit-react-core";
+
+import { Close, Edit } from "@hitachivantara/uikit-react-icons";
+
+import { makeData, AssetEvent } from "../storiesUtils";
+
+const EditableCell = ({ value }) => <HvInput defaultValue={value} />;
```
zettca, line 21: Let's hoist the toggle to this component (reasons above). Also, we should type the arguments (`noImplicitAny`)

```suggestion
const EditableCell = ({ value, row }: HvCellProps<AssetEvent, string>) => {
  return row.state?.isEditing ? <HvInput value={value} /> : value;
};
```

zettca (approved):

PR #4097, opened 2024-03-21
chore: configure Chromatic & review stories

MEsteves22:
- Chromatic configured: 
   - CI/GitHub Workflow
   - All stories are disabled by default and snapshotted stories are enabled at the story level
- Snapshotted stories reviewed: 
   - 🎉 **133 stories total**: feel free to suggest changes to reduce/increase the number of stories
   - ℹ️ **Since I'm using `play` to add interactions to the stories, we now have a11y issues to solve. Something to tackle in another PR. `play` is something we should start using in our stories to catch more a11y errors.**
- Fix: Stories inside "Tests/*" folder are hidden in the production build
- ℹ️ **Chromatic will only run in the nightly and in PRs if the `visual` tag is added.** This is temporary until we no longer depend on the free version. In the nightly, the workflow will continue if Chromatic finishes with errors. Applitools use the same stories as Chromatic and all stories are also disabled by default.


packages/core/src/DatePicker/DatePicker.stories.tsx
```diff
@@ -303,13 +300,13 @@ export const Controlled: StoryObj<HvDatePickerProps> = {
 
 export const WithSelectionList: StoryObj<HvDatePickerProps> = {
   parameters: {
-    eyes: {
-      runBefore() {
-        fireEvent.click(screen.getByRole("combobox"));
-
-        return waitFor(() => screen.getByRole("button", { name: "Apply" }));
```
VGervasio, line 310: Isn't a similar logic to this wait needed to make sure the image is taken when the date picker is open?
This comment is applicable in other places.

packages/core/src/Kpi/Kpi.stories.tsx
```diff
@@ -64,6 +59,10 @@ export const Main: StoryObj<HvKpiProps> = {
     },
     classes: { control: { disable: true } },
   },
+  parameters: {
+    // Enables Chromatic snapshot
+    chromatic: { disableSnapshot: false },
```
VGervasio, line 64: Isn't a delay needed given the previous waitBeforeCapture?

packages/core/src/Table/stories/TableSection/TableSection.stories.tsx
```diff
@@ -56,5 +68,10 @@ export const TableSettingsStory: StoryObj = {
   parameters: {
     docs: { source: { code: TableSettingsRaw } },
   },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const button = canvas.getByRole("button", { name: /settings/i });
+    await userEvent.click(button);
```
VGervasio, line 74: Do we need to add a check to make sure the panel is open when the accessibility tests run?

packages/core/src/DatePicker/DatePicker.stories.tsx
VGervasio: Looking at the image taken by Chromatic it does not seem needed:
image

packages/core/src/Kpi/Kpi.stories.tsx
VGervasio: Looking at the image taken by Chromatic it does not seem needed:
image

packages/core/src/Table/stories/TableSection/TableSection.stories.tsx
VGervasio: It seems this code runs every time storybook is deployed which means it will also run when the accessibility tests run.

zettca (approved):
Awesome stuff 💪

PR #4098, opened 2024-03-22
docs(templates): re-structure for App Shell routing

zettca:
Restructure templates:
- Remove sub-directories to avoid "collisions" with the App Shell auto-routing configuration
  - templates are simple enough so that they can stay in one directory IMO
- Co-locate classes (and rename `styles` > `classes`) with components ([according to the best practices](https://emotion.sh/docs/best-practices))
  - reduces number of files from `46` to `24`
- Improve Form template (use Native Form example from our Forms Guide)


francisco-guilherme (approved):

francisco-guilherme (approved):
💪

PR #4106, opened 2024-03-27
fix(icons): misc

zettca:
- Fix `Temperature` icon colors
- Remove old _unused_ code (`replaceColorsWithTheme`, `extractSize` dims, etc)
- Add `forwardRef` and `memo` to icons
- Reduce iconGenerator output further, now just a `createHvIcon` with necessary arguments

image


plagoa (approved):

MEsteves22:
I rerun the PR workflow because Playwright was failing, and we now have new errors in Applitools which are not expected. I was looking at the code and I don't really get why the icons are red  🤔

zettca:
> I rerun the PR workflow because Playwright was failing, and we now have new errors in Applitools which are not expected. I was looking at the code and I don't really get why the icons are red now 🤔

I rebased and couldn't reproduce it locally. Pushed the rebase and the issue seems to have gone away 😅

MEsteves22 (approved):

Issue #4109, opened 2024-04-01
HvQueryBuilder - Error in Value & (defaultValue) while changing the attributes list at run time.


vkumarakella:
### Which UI Kit version is this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

We observed that when we are dynamically changing attributes based on object type, query builder is giving error (in Value component).

And so, can you please change the code from **attributes[attribute].type**  to **attributes[attribute]?.type** and let us know.
![query-error](https://github.com/lumada-design/hv-uikit-react/assets/129264654/b005bbd1-d090-461b-9856-371f2c457475)

https://github.com/lumada-design/hv-uikit-react/assets/129264654/6f9c9c6c-a1d7-4200-b8bf-25ba9f25a7df

### Expected behavior 🤔

The Component should not throw error and the Attribute list should get updated dynamically based on object type.

### Steps to reproduce 🕹

Sandbox link: 

Steps to reproduce:

1.
2.
3.
4.

### Context 🔦

Affecting the deliverables and release.

### Your Environment 🌎

Build tool: (e.g. Vite 4, Webpack 5)
Device & OS: (e.g. Windows 11 laptop, iPhone 13 iOS15)
Browser: (e.g. Chrome 112, Safari 15.4)
Other details (e.g. 4K resolution, ...)

MEsteves22:
Hello 👋

Thanks for opening this issue. We'll take a look 🙏

MEsteves22:
Fixed in [@hitachivantara/uikit-react-core@5.62.1](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.62.1)

PR #4110, opened 2024-04-01
fix(QueryBuilder): value renderer

MEsteves22:
Addresses https://github.com/lumada-design/hv-uikit-react/issues/4109

### DIFF
--- a/packages/core/src/QueryBuilder/Rule/Value/Value.tsx
+++ b/packages/core/src/QueryBuilder/Rule/Value/Value.tsx
@@ -36,8 +36,7 @@ export const Value = ({
   const { attributes, initialTouched, renderers, emptyRenderer } =
     useQueryBuilderContext();
 
-  const attrType =
-    attribute && attributes ? attributes[attribute].type : undefined;
+  const attrType = attributes?.[attribute]?.type;
 
   // Empty value renderer
   if (emptyRenderer?.find((op) => op === operator)) {


packages/core/src/QueryBuilder/Rule/Value/Value.tsx
```diff
@@ -37,7 +37,7 @@ export const Value = ({
     useQueryBuilderContext();
 
   const attrType =
-    attribute && attributes ? attributes[attribute].type : undefined;
+    attribute && attributes ? attributes[attribute]?.type : undefined;
```
zettca, line 40: According to our types, `attribute` is always a string. Even if it were `undefined`, `({})[undefined]` wouldn't blow up - it just returns `undefined`.
```suggestion
    attributes?.[attribute]?.type;
```

I'm just suggesting this because (besides being simpler) the [operator precedence of the ternary](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table) isn't obvious to be, and I had to check it 😅 
```tsx
// which one is it?
attribute && (attributes ? attributes[attribute]?.type : undefined);
(attribute && attributes) ? attributes[attribute]?.type : undefined;
```

zettca (approved):

packages/core/src/QueryBuilder/Rule/Value/Value.tsx
MEsteves22: I'll update this

PR #4112, opened 2024-04-01
docs(Dropdown): add base dropdown documentation

HQFOX:
I'm not sure if "Custom Dropdown" the best title for this sample or if I should just leave it has "HvBaseDropdown". I also used this story to add a sample with TreeView with search, but it may be hard to find, tell me if you think its better to separate.

### DIFF
--- a/packages/core/src/Dropdown/stories/CustomDropdown.tsx
+++ b/packages/core/src/Dropdown/stories/CustomDropdown.tsx
@@ -1,0 +1,148 @@
+import { forwardRef, useState } from "react";
+import { css } from "@emotion/css";
+import {
+  HvIconButton,
+  HvInput,
+  HvSimpleGrid,
+  HvTreeItem,
+  HvTreeItemProps,
+  HvTreeView,
+  theme,
+} from "@hitachivantara/uikit-react-core";
+import { Doc, Folders, Plant } from "@hitachivantara/uikit-react-icons";
+
+import { HvBaseDropdown } from "../../BaseDropdown";
+import { HvPanel } from "../../Panel";
+
+type TreeData = { id: string; label: string; children?: TreeData[] };
+
+const treeDataObject = {
+  id: "user",
+  label: "User",
+  children: [
+    {
+      id: "Applications",
+      label: "Applications",
+      children: [
+        { id: "Code", label: "Code.app" },
+        { id: "Chrome", label: "Chrome.app" },
+        { id: "Firefox", label: "Firefox.app" },
+      ],
+    },
+    {
+      id: "Documents",
+      label: "Documents",
+      children: [{ id: "secret", label: "secret.txt" }],
+    },
+    {
+      id: "git",
+      label: "git",
+      children: [
+        {
+          id: "uikit-react",
+          label: "uikit-react",
+          children: [{ id: "uikit-pkg", label: "package.json" }],
+        },
+        {
+          id: "app-shell",
+          label: "app-shell",
+          children: [{ id: "as-pkg", label: "package.json" }],
+        },
+      ],
+    },
+  ],
+} satisfies TreeData;
+
+const SimpleTreeItem = forwardRef<HTMLLIElement, HvTreeItemProps>(
+  (props, ref) => {
+    const { children, nodeId, label, ...others } = props;
+    const Icon = children ? Folders : Doc;
+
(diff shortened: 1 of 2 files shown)


packages/core/src/Dropdown/stories/CustomDropdown.tsx
```diff
@@ -0,0 +1,149 @@
...
+  HvInput,
+  HvSimpleGrid,
+  HvTreeItem,
+  HvTreeItemProps,
+  HvTreeView,
+  theme,
+} from "@hitachivantara/uikit-react-core";
+import { Doc, Folders, Plant } from "@hitachivantara/uikit-react-icons";
+
+import { HvBaseDropdown } from "../../BaseDropdown";
```
MEsteves22, line 14: Import from `@hitachivantara/uikit-react-core` should be used 🔍

packages/core/src/Dropdown/stories/CustomDropdown.tsx
```diff
@@ -0,0 +1,149 @@
...
+    {children?.map(renderItem)}
+  </SimpleTreeItem>
+);
+
+export const CustomDropdown = () => {
+  const [results, setResults] = useState<TreeData | undefined>(treeDataObject);
+
+  const search = (data: TreeData, searchTerm: string): TreeData | undefined => {
+    let newChildren: TreeData[] = [];
+    if (data.label.includes(searchTerm)) {
```
MEsteves22, line 91: If I research "firefox" for example, the "Firefox" item is not shown because of the capitalised letter. We could add `toLowerCase`:

```suggestion
    if (data.label.toLowerCase().includes(searchTerm.toLowerCase())) {
```

packages/core/src/Dropdown/stories/CustomDropdown.tsx
```diff
@@ -0,0 +1,149 @@
...
+);
+
+/** Render tree view items */
+const renderItem = ({ id, label, children }: TreeData) => (
+  <SimpleTreeItem key={id} nodeId={id} label={label}>
+    {children?.map(renderItem)}
+  </SimpleTreeItem>
+);
+
+export const CustomDropdown = () => {
```
MEsteves22, line 86: In the first example, when we select an item from the tree view, the dropdown input is not updated. I think it would be useful to add an example where we showcase this typical dropdown behaviour. 

Regarding the second example with the icon button, I think it's very useful to have it since this is something that is starting to appear more frequently: for example with the "Table Settings" story. However, we could improve it by using a pattern similar to [this one](https://hv-eng.atlassian.net/browse/DIG-1229) but simplified: the button opens a panel with two tree view dropdown for example.  

What do you think?

MEsteves22 (approved):
👌🏻

packages/core/src/Dropdown/stories/Dropdown.stories.tsx
```diff
@@ -30,6 +37,16 @@ const widthDecorator: Decorator = (Story) => (
 export default {
   title: "Components/Dropdown",
   component: HvDropdown,
+  // @ts-expect-error https://github.com/storybookjs/storybook/issues/20782
+  subcomponents: { HvBaseDropdown },
+  parameters: {
```
MEsteves22, line 42: Can we remove this? It's not needed. We already have the `play` functions on the stories that need interactions.

packages/core/src/Dropdown/stories/Dropdown.stories.tsx
```diff
@@ -213,3 +230,20 @@ export const Virtualized: StoryObj<HvDropdownProps> = {
...
+    docs: {
+      description: {
+        story:
+          "You can also use the `HvBaseDropdown` component to create your own dropdown components. This component can be leveraged to create similar dropdown components to the ones we provide, or you can use the `component` prop to add it to any component of your choosing.",
+      },
+      source: {
+        code: CustomDropdownRaw,
+      },
+    },
+    eyes: { include: false },
```
MEsteves22, line 245: We can also remove this. It's no longer needed. All stories are disabled in Applitools by default now.

MEsteves22 (approved):
✅

I just left these comments (some stuff to remove):

plagoa (approved):

zettca (approved):

PR #4114, opened 2024-04-02
fix(a11y): fix issues

zettca:
- fix a11y issues found by new tool
- rename `aria-prohibited-attr` > `aria-allowed-attr`
- merge KPI Selectable stories & use `?raw` loader
- refactor `ScrollTo` to better address the tooltip + label issues
  - move the tooltip logic to the `*ScrollListItem.tsx` & leverage `HvOverflowTooltip`
  - move the "bullet" icon to `*ScrollListItem.tsx` to not include it in the tooltip
    - always use a rounded div, instead of div + icon - simplifying the logic & making it easier to customise (no need for the `activeTheme` tokens anymore)
  - small fixes in spacing (using DS5 units instead of DS3)
- refactor `HvList` to use `HvOverflowTooltip`
  - fixes tooltip a11y issues in components using it

### DIFF
--- a/packages/core/src/ScrollTo/Horizontal/HorizontalScrollListItem/HorizontalScrollListItem.styles.tsx
+++ b/packages/core/src/ScrollTo/Horizontal/HorizontalScrollListItem/HorizontalScrollListItem.styles.tsx
@@ -7,51 +7,44 @@ const name = "HvHorizontalScrollListItem";
 
 export const { staticClasses, useClasses } = createClasses(name, {
   root: {
-    padding: "10px 0",
+    padding: theme.spacing("xs", 0),
+    maxWidth: 120,
   },
   button: {
     display: "flex",
+    flexDirection: "column",
     justifyContent: "center",
     alignItems: "center",
-    height: "48px",
-    cursor: "pointer",
-    borderBottom: "none",
-    "&:hover": {
-      backgroundColor: "transparent",
-
-      "& $notSelected": {
-        height: "6px",
-        width: "6px",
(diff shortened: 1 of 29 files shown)


packages/core/src/ScrollTo/Horizontal/HorizontalScrollListItem/HorizontalScrollListItem.styles.tsx
```diff
@@ -7,51 +7,44 @@ const name = "HvHorizontalScrollListItem";
...
-    },
-    "&:focus": {
-      outline: "none",
-    },
     "&:focus-visible": {
       ...outlineStyles,
     },
   },
   text: {
+    padding: theme.spacing("xs", "xs", "0"),
```
MEsteves22, line 23: The tooltip is always shown even if the label isn't truncated. Changing this to margin fixes the issue.

MEsteves22 (commented):
👌🏻

I just have the following comment and I accepted the Chromatic changes:

packages/core/src/ScrollTo/Horizontal/HorizontalScrollListItem/HorizontalScrollListItem.styles.tsx
zettca: That's right, I changed it to `margin` because of this, but reverted it somehow 😅 

https://github.com/lumada-design/hv-uikit-react/commit/f0a38981b799885622e616cca9dcdb7270065391

plagoa (approved):

PR #4115, opened 2024-04-02
fix: base radio and base checkbox style

MEsteves22:
- The wrong border color was used for the base checkbox component when disabled and checked
- The wrong background color was used for the base radio component


MEsteves22:
Applitools did not detect the changes for the radio and checkbox component. However, some errors were detected but they are due to some instability. I'll try to fix this by adding expect statements in storybook `play` as discussed.

francisco-guilherme (approved):

PR #4118, opened 2024-04-03
docs: add Kanban Board template

plagoa:


MEsteves22 (approved):
👌🏻

Note: When a column has no card, its width decreases. Maybe we should divide the column containers into equal widths.

plagoa:
> Note: When a column has no card, its width decreases. Maybe we should divide the column containers into equal widths.

I'll check it, thanks :)

plagoa:
@MEsteves22 the column width issue should be fixed now.

francisco-guilherme (approved):

PR #4119, opened 2024-04-03
chore: assertions added to storybook play

MEsteves22:
- Assertions were added to storybook play functions to try to fix the unstable stories (example [here](https://eyes.applitools.com/app/test-results/00000251690157190775/?accountId=NWMfIF0XfUeh3g9JqZjRrg~~) with the tree view component)

ℹ️ I initially added the Interactions addon to storybook to have a [UI tool for interactions](https://storybook.js.org/docs/writing-stories/play-function#setup-the-interactions-addon) but I had to remove it since Applitools was not able to run when used. Once we no longer depend on Applitools, we can add it back.

### DIFF
--- a/docs/tests/Tags.stories.tsx
+++ b/docs/tests/Tags.stories.tsx
@@ -1,4 +1,5 @@
 import { css } from "@emotion/css";
+import { expect } from "@storybook/jest";
 import { StoryObj } from "@storybook/react";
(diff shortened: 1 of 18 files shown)


MEsteves22:
⚠️ The applitools change are not related with this PR but with instability with the icons.

francisco-guilherme (approved):

PR #4123, opened 2024-04-04
fix: icon palette mutation

MEsteves22:
Problem:

https://github.com/lumada-design/hv-uikit-react/assets/43220251/d83ee55a-c36e-4ff7-b830-2b1c3e3bc6d8

- This is why we were having several problems in Applitools


MEsteves22:
🟢 Applitools on the first try 🎉

francisco-guilherme (approved):

zettca (approved):

PR #4124, opened 2024-04-04
fix(Drawer): remove onclose trigger

HQFOX:
It was reported in ui-kit channel that `onClose` was being triggered twice when clicking the backdrop. This code seems to me to be redundant and removing it fixed it.


zettca (commented):
Can we add a `.toHaveBeenCalledTimes(1)` assertion to the close test, to ensure this is working as expected? 👀

zettca (approved):

MEsteves22 (approved):

PR #4130, opened 2024-04-05
feat: onEvents added to visualisations & sample with tooltip for truncated labels added

MEsteves22:
- `onEvents` added for the visualisations to trigger events when needed:
   - ℹ️ I had to put `dispose` from `HvBaseChart` in another `useEffect` because having it in the same effect was preventing us from using `onEvents` (the callback was not triggered). 
-  Sample that triggers a tooltip when hovering a truncated y axis label added (based [on](https://github.com/apache/echarts/pull/16315))


zettca (approved):
🚀

PR #4143, opened 2024-04-11
feat: add Treemap viz

plagoa:


packages/viz/src/Treemap/Treemap.tsx
```diff
@@ -0,0 +1,98 @@
...
+      height,
+      tooltip,
+      onOptionChange,
+      ...others
+    } = props;
+
+    const { colors } = useTheme();
+    const chartTooltip = useTooltip({
+      ...tooltip,
+      type: "single",
```
MEsteves22, line 58: The `type` defined by the user (inside `tooltip`) will be overridden. If there's no `multiple` tooltip type for the treetop, we should remove it.

MEsteves22 (commented):
LGTM👌🏻I just have the following comments:

packages/viz/src/Treemap/Treemap.tsx
```diff
@@ -0,0 +1,83 @@
...
+import ReactECharts from "echarts-for-react/lib/core";
+import { TreemapChart } from "echarts/charts";
+import { TooltipComponent } from "echarts/components";
+import * as echarts from "echarts/core";
+
+import { HvBaseChart } from "../BaseChart";
+import { HvChartTooltipClasses, useOption, useTooltip } from "../hooks";
+import { HvChartTooltip } from "../types";
+import { HvChartCommonProps } from "../types/common";
+import { HvTreemapData } from "../types/generic";
```
zettca, line 11: This `HvTreemapData` look specific to this chart and not generic

zettca (approved):
LGTM 👍

packages/viz/src/Treemap/Treemap.tsx
plagoa: You're right, I'll update that.

MEsteves22 (approved):

PR #4144, opened 2024-04-12
feat(Node): add inline edit

HQFOX:
Expected applitools error

### DIFF
--- a/packages/lab/src/Flow/Flow.spec.ts
+++ b/packages/lab/src/Flow/Flow.spec.ts
@@ -98,6 +98,7 @@ test.describe("Node", () => {
     const dragDestinationBox = await flowCanvasLocator.boundingBox();
     const nodeInCanvasBox = await flowCanvasLocator
       .getByRole("button", { name: "Asset" })
+      .first()
       .boundingBox();
 
     /* expect(nodeInCanvasBox?.x).toEqual(dragDestinationBox!.x + 10);
@@ -117,19 +118,22 @@ test.describe("Node", () => {
     const node = await selectNode("ML Model", "ML Model Prediction", true);
     await dragToPosition1(node);
 
-    const createdNode = flowCanvasLocator.getByRole("button", {
-      name: "ML Model Prediction",
-    });
+    const createdNode = flowCanvasLocator
+      .getByRole("button", {
+        name: "ML Model Prediction",
+      })
+      .first();
     expect(await createdNode.all()).toHaveLength(1);
     await flowCanvasLocator
       .getByRole("button", { name: "ML Model Prediction" })
+      .first()
       .hover();
     await page.getByRole("button", { name: "Duplicate" }).hover();
     await expect(
       page.getByRole("tooltip", { name: "Duplicate" }),
     ).toBeVisible();
     await page.getByRole("button", { name: "Duplicate" }).click();
-    expect(await createdNode.all()).toHaveLength(2);
+    expect(await createdNode.all()).toHaveLength(1);
   });
 
   test("should be able to delete a node", async ({ page }) => {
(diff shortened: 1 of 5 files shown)


packages/lab/src/Flow/Flow.spec.ts
```diff
@@ -117,19 +118,22 @@ test.describe("Node", () => {
...
       .getByRole("button", { name: "ML Model Prediction" })
+      .first()
       .hover();
     await page.getByRole("button", { name: "Duplicate" }).hover();
     await expect(
       page.getByRole("tooltip", { name: "Duplicate" }),
     ).toBeVisible();
     await page.getByRole("button", { name: "Duplicate" }).click();
-    expect(await createdNode.all()).toHaveLength(2);
+    expect(await createdNode.all()).toHaveLength(1);
```
MEsteves22, line 136: The length should still be 2 here since we are testing if the node gets duplicated. If we need to update this to 1 then there's probably something that needs fixing.

packages/lab/src/Flow/Node/Node.tsx
```diff
@@ -144,11 +160,28 @@ export const HvFlowNode = ({
...
+            (disableInlineEdit ? (
               <HvTypography>{subtitle}</HvTypography>
-            </div>
-          )}
+            ) : (
+              <HvInlineEditor
+                defaultValue={subtitle}
+                showIcon
+                classes={{
+                  root: css({
```
MEsteves22, line 171: I think we should put this CSS inside the style file we have for consistency.

packages/lab/src/Flow/Node/Node.tsx
```diff
@@ -144,11 +160,28 @@ export const HvFlowNode = ({
     >
       {(subtitle || actions) && (
         <div className={classes.subtitleContainer}>
```
MEsteves22, line 162: Can we update the padding of the `subtitleContainer` to `padding: theme.space.xs` when `disableInlineEdit` is `true` to remove the unneeded extract left padding?

packages/lab/src/Flow/Node/Node.tsx
```diff
@@ -62,6 +65,8 @@ export interface HvFlowNodeProps<T = any> extends HvFlowBaseNodeProps<T> {
   labels?: HvFlowBaseNodeProps["labels"] & Partial<typeof DEFAULT_LABELS>;
   /** A Jss Object used to override or extend the styles applied to the component. */
   classes?: HvFlowNodeClasses;
+  /** Remove the ability to customize the label of the Node */
+  disableInlineEdit?: boolean;
 }
 
 export const HvFlowNode = ({
```
MEsteves22, line 72: We need to run Chromatic by adding the `visual` tag to the PR as there will be changes 🙏

packages/lab/src/Flow/Node/Node.tsx
```diff
@@ -144,11 +160,28 @@ export const HvFlowNode = ({
...
+                    display: "inline-flex",
+                    flexGrow: 1,
+                  }),
+                  button: css({ justifyContent: "space-between" }),
+                  inputRoot: css({
+                    width: inlineEditorWidth,
+                  }),
+                }}
+                onBlur={(evt, value) =>
+                  setNodeData((prev) => ({ ...prev, label: value }))
```
MEsteves22, line 181: I was thinking if it wouldn't be better to use a different name than `label` since it seems generic enough that some users may already be using it. 
Maybe `customNodeLabel` or other.

MEsteves22 (commented):
🙏

MEsteves22 (approved):

zettca (approved):

Issue #4148, opened 2024-04-16
HvQueryBuilder - Error while adding translation for labels


vkumarakella:
### Which UI Kit version is this bug for?

v5.x

### Latest version

- [X] I have tested the latest version

### No Hitachi confidential content

- [X] I'm not including any Hitachi confidential content

### Current behavior 😯

Under **HvQueryBuilder** When I am trying to add translations for labels, the page is throwing error.
Note: Observed that we have console error under `<RuleGroup/>` component.

Please find the screenshots for reference:
![queryTError](https://github.com/lumada-design/hv-uikit-react/assets/129264654/3b2c8ce8-f544-4cd8-bfd6-c33de6e63512)

![ruleGroup](https://github.com/lumada-design/hv-uikit-react/assets/129264654/9d5084cb-4d2b-4568-8ebd-f39abc8d02f8)

### Expected behavior 🤔

The Page should not throw 500 error.
and All the Query Builder labels should be allowed for translation to other language scripts.

### Steps to reproduce 🕹

Sandbox link: 

Steps to reproduce:

1.
2.
3.
4.

### Your Environment 🌎

Build tool: (e.g. Vite 4, Webpack 5)
Device & OS: (e.g. Windows 11 laptop, iPhone 13 iOS15)
Browser: (e.g. Chrome 112, Safari 15.4)
Other details (e.g. 4K resolution, ...)

MEsteves22:
This was fixed in the latest release: [@hitachivantara/uikit-react-core@5.64.1](https://github.com/lumada-design/hv-uikit-react/releases/tag/%40hitachivantara%2Fuikit-react-core%405.64.1). Thanks!

PR #4149, opened 2024-04-16
fix(QueryBuilder): override labels

MEsteves22:
Addresses https://github.com/lumada-design/hv-uikit-react/issues/4148

### DIFF
--- a/packages/core/src/QueryBuilder/QueryBuilder.tsx
+++ b/packages/core/src/QueryBuilder/QueryBuilder.tsx
@@ -2,6 +2,8 @@ import { useEffect, useMemo, useReducer, useRef, useState } from "react";
 
 import { useControlled } from "../hooks/useControlled";
 import { useDefaultProps } from "../hooks/useDefaultProps";
+import { useLabels } from "../hooks/useLabels";
+import { DeepPartial } from "../types/generic";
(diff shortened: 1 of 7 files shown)


packages/core/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -52,7 +54,7 @@ export interface HvQueryBuilderProps {
   /** Max depth of nested query groups. */
   maxDepth?: number;
   /** Object containing all the labels. */
-  labels?: HvQueryBuilderLabels;
+  labels?: DeepPartial<HvQueryBuilderLabels>;
```
zettca, line 57: Note to self for v6: flatten these `labels` so we (and our users) don't have to deal with the nesting complexity
The `rule.value` should probably be replaced with a more streamlined way to configure or override "renderers"  too

zettca (approved):

packages/core/src/QueryBuilder/QueryBuilder.tsx
MEsteves22: True, I'll add a comment to not forget this for the next major

plagoa (approved):

PR #4166, opened 2024-04-29
feat: add Boxplot viz

plagoa:


packages/viz/src/hooks/useBoxplotData.tsx
```diff
@@ -0,0 +1,52 @@
...
+    let tableData: ColumnTable;
+    if (data instanceof internal.ColumnTable) {
+      tableData = data;
+    } else if (Array.isArray(data)) {
+      tableData = from(data);
+    } else {
+      tableData = table(data);
+    }
+
+    const groupByField = groupBy
```
MEsteves22, line 29: If we can only use one value for the `groupBy` prop, then we should update `HvBoxplotProps` and only accept a `string`.

packages/viz/src/Boxplot/Boxplot.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+export interface HvBoxplotProps
+  extends HvAxisChartCommonProps,
+    HvChartCommonProps {
+  /** The name of the heatmap */
+  name?: string;
+  /**  Columns to measure on the chart. */
+  measures: Arrayable<HvBoxplotMeasure>;
+  /** The tooltip options. */
+  tooltip?: Omit<HvChartTooltip, "type">;
```
MEsteves22, line 35: Is `valueFormatter` and `titleFormatter` used? If not, we should omit these props I think.

packages/viz/src/types/measures.ts
```diff
@@ -53,6 +53,8 @@ export interface DonutFullMeasure extends BaseMeasure {}
 
 export interface ConfusionMatrixMeasure extends BaseMeasure {}
 
+export interface BoxplotMeasure extends Omit<BaseMeasure, "agg">, AxisMeasure {}
```
MEsteves22, line 56: Is `valueFormatter` used? If not, I think we should omit it.

MEsteves22 (commented):
👌🏻

plagoa:
@MEsteves22 made some changes based on your review.

packages/viz/src/Boxplot/Boxplot.tsx
```diff
@@ -0,0 +1,134 @@
...
+export type HvBoxplotClasses = ExtractNames<typeof useClasses>;
+
+export interface HvBoxplotProps
+  extends HvAxisChartCommonProps,
+    Omit<HvChartCommonProps, "groupBy"> {
+  /** The name of the heatmap */
+  name?: string;
+  /**  Columns to measure on the chart. */
+  measures: Arrayable<HvBoxplotMeasure>;
+  /** */
```
MEsteves22, line 34: Missing description

packages/viz/src/Boxplot/useBoxplotData.tsx
```diff
@@ -0,0 +1,52 @@
+import { useMemo } from "react";
+import { from, internal, table } from "arquero";
+import type ColumnTable from "arquero/dist/types/table/column-table";
+
+import { HvChartData } from "../types";
+import { HvChartCommonProps } from "../types/common";
+
+interface HvDataHookProps {
+  data: HvChartData;
+  groupBy?: HvChartCommonProps["groupBy"];
```
MEsteves22, line 10: This is a `string` now

packages/viz/src/Boxplot/useBoxplotData.tsx
```diff
@@ -0,0 +1,52 @@
...
+    let tableData: ColumnTable;
+    if (data instanceof internal.ColumnTable) {
+      tableData = data;
+    } else if (Array.isArray(data)) {
+      tableData = from(data);
+    } else {
+      tableData = table(data);
+    }
+
+    const groupByField = groupBy
```
MEsteves22, line 29: We no longer need this condition as it will be a `string`

MEsteves22 (approved):
🚀 

Just found the following:

HQFOX (approved):

PR #4167, opened 2024-04-30
feat: add filtering capabilities to the visualizations

plagoa:


packages/viz/src/BarChart/stories/BarChart.stories.tsx
```diff
@@ -494,3 +497,110 @@ export const CustomEchartsOptions: StoryObj<HvBarChartProps> = {
...
+        label: "UK",
+        value: 2700,
+      },
+      {
+        label: "Japan",
+        value: 5800,
+      },
+    ];
+
+    useEffect(() => {
```
MEsteves22, line 546: I think we can remove this `useEffect` and only have one state, `filters`. Then, inside `handleSliderChange` and `handleCountryChange` we could do:

```
// handleCountryChange
setFilters((prevState) => ([
        ...prevState,
         {
           field: "Country",
           operation: "contains",
           value: country,
         }
       ]))

// handleSliderChange
setFilters((prevState) => ([
        ...prevState,
         {
           field: "Sales",
           operation: "between",
           value: [sales[0], sales[1]],
         },
       ]))
```

Less re-renders and states to manage.

packages/viz/src/BarChart/stories/BarChart.stories.tsx
```diff
@@ -494,3 +497,110 @@ export const CustomEchartsOptions: StoryObj<HvBarChartProps> = {
...
+          <HvSlider
+            minPointValue={0}
+            maxPointValue={7000}
+            values={sales}
+            onChange={handleSliderChange}
+            hideInput
+            label="Sales"
+            style={{ width: 400 }}
+          />
+          <HvInput
```
MEsteves22, line 588: The input width changes when we type something 👀 Can we set the width? Or maybe a dropdown (`HvSelect`) with multiple selection would be better for the sample: if I type "india" instead of "India" nothing shows because of the capitalised letter.

packages/viz/src/BarChart/BarChart.tsx
```diff
@@ -70,6 +70,7 @@ export const HvBarChart = forwardRef<ReactECharts, HvBarChartProps>(
       groupBy,
       splitBy,
       sortBy,
+      filters,
```
MEsteves22, line 73: We need to do the same to the other charts that are extending `HvChartCommonProps`. Otherwise the prop is available but doesn't do nothing.

packages/viz/src/types/filter.ts
```diff
@@ -0,0 +1,15 @@
+export type HvChartFilterOperations =
```
MEsteves22, line 1: `HvChartFilterOperation` (conventionally. we shouldn't pluralize)

packages/viz/src/BarChart/stories/BarChart.stories.tsx
zettca: This is [derived state](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state) - we don't need state

```tsx
const filters = [
        {
          field: "Country",
          operation: "contains",
          value: country,
        },
        {
          field: "Sales",
          operation: "between",
          value: [sales[0], sales[1]],
        },
      ];
```

MEsteves22 (approved):

PR #4184, opened 2024-05-21
chore: enable `react/`jsx-key rule & apply fixes

zettca:
Spreading key is discouraged and there's a React warning since `v18.3` https://github.com/facebook/react/pull/25697
[`react/jsx-key`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-key.md#rule-details) rule covers this

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,6 +24,7 @@
   "rules": {
     "arrow-body-style": "off",
     "prefer-arrow-callback": "off",
+    "react/jsx-key": "error",
     "react/prop-types": "off",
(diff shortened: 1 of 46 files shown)


zettca:
> This wasn't caught by ESLint but [it seems like we need to apply these changes to the global action component](https://hitachivantara-eng.slack.com/archives/CFY74GK6G/p1716363801069459?thread_ts=1716277751.499289&cid=CFY74GK6G).

I've added a couple more fixes (searched for `key:` in objects)

MEsteves22 (approved):

PR #4206, opened 2024-06-19
feat: add useBaseNode hook and demo nodes

HQFOX:
A useBaseNode hook (in lack of a better name) was created with the logic parts of `BaseNode` and `Node` leaving the visual implementation to the components that use it. 
To exemplify this some nodes were added and are present in the `Base Hook` story in the documentation:

A simpler `WebttleNode` that uses the `useBaseNode` hook and has a different visual (similar to the webttle use case).

A `DANode` that uses the `useBaseNode` hook instead of the `BaseNode` and `Node` components but keeps the same api.

A `AssetDA` and `MLModelDA` that use the `DANode` component to implement the `Asset` and `MLModel` nodes.

A `AssetWorkbench`  a node that uses the `useBaseNode` to implement a `Asset` node.

`AssetDA` and `AssetWorkbench` were created to compare the different approaches of using the hook. With `AssetDA` we still have a big and cumbersome api and end up with two different complex components `AssetDA` and `DANode` but they cover most use cases of that project. Meanwhile `AssetWorkbench` directly uses the hook is much more "readable" and easy to maintain.

### DIFF
--- a/packages/lab/src/Flow/hooks/useNode.tsx
+++ b/packages/lab/src/Flow/hooks/useNode.tsx
@@ -1,4 +1,10 @@
-import { isValidElement, useCallback, useMemo, useState } from "react";
+import {
+  isValidElement,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import { NodeToolbarProps } from "reactflow";
 import { uid } from "uid";
 import { useLabels } from "@hitachivantara/uikit-react-core";
@@ -27,7 +33,7 @@ const DEFAULT_LABELS = {
   duplicateActionLabel: "Duplicate",
 };
 
-export interface HvUseNodeProps {
+export interface HvUseNodeParams {
   id: string;
   /** Node group ID */
   groupId?: string;
@@ -51,7 +57,7 @@ export interface HvUseNodeProps {
   nodeToolbarProps?: NodeToolbarProps;
 }
 
-export function useHvNode(props: HvUseNodeProps) {
+export function useHvNode(props: HvUseNodeParams) {
   const {
     id,
     title: titleProp,
@@ -67,13 +73,11 @@ export function useHvNode(props: HvUseNodeProps) {
   } = props;
 
   const { registerNode, unregisterNode } = useNodeMetaRegistry();
-
   const labels = useLabels(DEFAULT_LABELS, labelsProps);
-
   const inputs = useMemo(() => identifyHandles(inputsProp), [inputsProp]);
-
+  const inputEdges = useFlowNodeInputEdges();
   const outputs = useMemo(() => identifyHandles(outputsProp), [outputsProp]);
-
+  const outputEdges = useFlowNodeOutputEdges();
   const { nodeGroups } = useFlowContext();
 
   const node = useFlowNode();
@@ -105,7 +109,7 @@ export function useHvNode(props: HvUseNodeProps) {
     [nodeToolbarProps, showActions],
   );
 
-  const nodeActions = useMemo(
+  const nodeActions = useMemo<HvFlowNodeAction[]>(
     () =>
       nodeActionsProp || [
         { id: "delete", label: labels?.deleteActionLabel, icon: <Delete /> },
@@ -118,7 +122,7 @@ export function useHvNode(props: HvUseNodeProps) {
     [labels?.deleteActionLabel, labels?.duplicateActionLabel, nodeActionsProp],
   );
 
-  useMemo(() => {
+  useEffect(() => {
     registerNode(id, {
       label: title || "",
       inputs,
@@ -126,9 +130,6 @@ export function useHvNode(props: HvUseNodeProps) {
     });
     return () => unregisterNode(id);
   }, [id, title, inputs, outputs, registerNode, unregisterNode]);
-
-  const inputEdges = useFlowNodeInputEdges();
-  const outputEdges = useFlowNodeOutputEdges();
 
   const handleDefaultAction = useCallback(
     (action: HvFlowNodeAction) => {
@@ -165,25 +166,45 @@ export function useHvNode(props: HvUseNodeProps) {
     [node, reactFlowInstance],
   );
 
-  return {
-    // state
-    id,
-    title,
-    icon,
-    color,
-    iconColor,
-    subtitle,
-    inputs,
-    inputEdges,
-    outputs,
-    outputEdges,
-    node,
-    nodeActions,
-    showActions,
-    // prop getters
-    getNodeToolbarProps,
-    // actions
-    toggleShowActions,
-    handleDefaultAction,
-  };
+  return useMemo(
+    () => ({
+      // state
+      id,
+      title,
+      icon,
+      color,
+      iconColor,
+      subtitle,
+      inputs,
+      inputEdges,
+      outputs,
+      outputEdges,
+      node,
+      nodeActions,
+      showActions,
+      // prop getters
+      getNodeToolbarProps,
+      // actions
+      toggleShowActions,
+      handleDefaultAction,
+    }),
+    [
+      id,
+      title,
+      icon,
+      color,
+      iconColor,
+      subtitle,
+      inputs,
+      inputEdges,
+      outputs,
+      outputEdges,
+      node,
+      nodeActions,
+      showActions,
+      getNodeToolbarProps,
+      toggleShowActions,
+      handleDefaultAction,
+    ],
+  );
 }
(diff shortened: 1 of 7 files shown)


plagoa:
I think this approach is really nice. Regarding the name, I think we could use `useNode` instead of `useBaseNode` .

HQFOX:
> I think this approach is really nice. Regarding the name, I think we could use `useNode` instead of `useBaseNode` .

I agree, but it maybe can be too similar to react flow's `useNodes` hook , maybe `useHvNode` could help distinguish?

packages/lab/src/Flow/hooks/useBaseNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+    inputs: inputsProp,
+    outputs: outputsProp,
+    groupId,
+    labels: labelsProps,
+    nodeToolbarProps,
+  } = props;
+
+  const { registerNode, unregisterNode } = useNodeMetaRegistry();
+
+  const id = idProp || uid();
```
zettca, line 71: This `id` is unstable like this. Shouldn't it be?:

```suggestion
  const id = useUniqueId(idProp);
```

or memo it instead

packages/lab/src/Flow/stories/Flow.stories.tsx
```diff
@@ -175,3 +176,19 @@ export const InvalidFlow: StoryObj<HvFlowProps> = {
...
+
+export const BaseHook: StoryObj<HvFlowProps> = {
+  parameters: {
+    docs: {
+      description: {
+        story: "A Flow with an initial state",
+      },
+      source: {
+        code: InitialStateRaw,
+      },
```
zettca, line 188: this should reflect the hook

packages/lab/src/Flow/stories/BaseHook/WebttleNode.tsx
```diff
@@ -0,0 +1,120 @@
...
+            {renderedIcon(action.icon)}
+          </HvIconButton>
+        ))}
+      </NodeToolbar>
+      <div
+        className={css({
+          position: "absolute",
+          top: `calc(-24px + ${theme.space.xs})`,
+          right: `calc(-24px + ${theme.space.xs})`,
+        })}
```
zettca, line 92: these styles should be hoisted

packages/lab/src/Flow/stories/BaseHook/WebttleNode.tsx
```diff
@@ -0,0 +1,120 @@
...
+          style={{
+            width: 48,
+            height: 48,
+            display: "flex",
+            justifyContent: "center",
+            alignItems: "center",
+            borderRadius: theme.radii.round,
+            backgroundColor: color,
+            marginRight: theme.space.xs,
+          }}
```
zettca, line 109: these styles should be hoisted (except `backgroundColor`)

packages/lab/src/Flow/stories/BaseHook/WebttleNode.tsx
```diff
@@ -0,0 +1,120 @@
...
+          position: "absolute",
+          top: `calc(-24px + ${theme.space.xs})`,
+          right: `calc(-24px + ${theme.space.xs})`,
+        })}
+      >
+        <Level4 />
+      </div>
+      <Handle type="target" position={Position.Left} id="0" />
+      <Handle type="source" position={Position.Right} id="0" />
+      <div className={cx(classes.content)}>
```
zettca, line 98:
```suggestion
      <div className={classes.content}>
```

packages/lab/src/Flow/stories/BaseHook/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+  model: MLModel,
+  kpi: KPI,
+  lineChart: LineChart,
+  table: Table,
+  dashboard: Dashboard,
+  DA: DANode,
+  webttle: WebttleNode,
+  assetDA: AssetDA,
+  modelDA: MLModelDA,
+  assetWorkbench: AssetWorkbench,
```
zettca, line 132: shouldn't we focus only on the nodes built with the hook? I think adding all the other nodes adds entropy

packages/lab/src/Flow/stories/BaseHook/BaseHook.tsx
```diff
@@ -0,0 +1,158 @@
...
+    minHeight: 200,
+  }),
+  outputLabel: css({
+    display: "flex",
+    alignItems: "center",
+    gap: 2,
+  }),
+};
+
+export const BaseHook: HvFlowNodeFC<AssetData> = (props) => {
```
zettca, line 42: Is this file being used? I couldn't find where.
Also, it has a confusing name. It's called `BaseHook` but `index.tsx` doesn't use it

packages/lab/src/Flow/hooks/useBaseNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+  outputs?: (HvFlowNodeOutput | HvFlowNodeOutputGroup)[];
+  /** Node actions */
+  nodeActions?: HvFlowNodeAction[];
+  /** Labels used on the default actions. */
+  labels?: Partial<typeof DEFAULT_LABELS>;
+  /** Props for the NodeToolbar Component */
+  nodeToolbarProps?: NodeToolbarProps;
+}
+
+export function useHvBaseNode(props: HvFlowUseBaseNodeProps) {
```
zettca, line 54: shouldn't we be leveraging this in `HvBaseNode`?

also, this hook isn't being exported

packages/lab/src/Flow/hooks/useBaseNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+  useFlowNodeInputEdges,
+  useFlowNodeOutputEdges,
+} from "./useFlowNode";
+
+export const DEFAULT_LABELS = {
+  deleteActionLabel: "Delete",
+  duplicateActionLabel: "Duplicate",
+};
+
+export interface HvFlowUseBaseNodeProps {
```
zettca, line 30: shouldn't we be sharing this with `HvBaseNode`?

zettca (commented):
The hook looks nice 👍

I think we should leverage it in `BaseNode` if we can.
There are also some issues with docs:

packages/lab/src/Flow/hooks/useNode.tsx
```diff
@@ -0,0 +1,191 @@
...
+          ]);
+          break;
+        default:
+          break;
+      }
+    },
+    [node, reactFlowInstance],
+  );
+
+  return {
```
MEsteves22, line 168: I think the following props shouldn't be exposed to the user:
- `registerNode`
- `unregisterNode`

MEsteves22 (commented):
Looks good 🙏

packages/lab/src/Flow/stories/BaseHook/TeapotNode.tsx
```diff
@@ -0,0 +1,156 @@
...
+          <Level0Good color="positive" />
+        ) : (
+          <Level2Average color="warning" />
+        )}
+      </div>
+      <Handle
+        className={classes.handle}
+        type="target"
+        position={Position.Left}
+        id="0"
```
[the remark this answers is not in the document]
HQFOX: Currently, the index of the array of inputs or outputs is being used as the id. This makes it so that when I add an id of 1 the checks of the connection won't work. I think we should open a separate issue to tackle this.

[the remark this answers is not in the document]
MEsteves22: Sorry, I also got it wrong: one is an output and the other one is an input. Nothing wrong with having the same id 🙏

packages/lab/src/Flow/hooks/useNode.tsx
```diff
@@ -0,0 +1,189 @@
...
+        {
+          id: "duplicate",
+          label: labels?.duplicateActionLabel,
+          icon: <Duplicate />,
+        },
+      ],
+    [labels?.deleteActionLabel, labels?.duplicateActionLabel, nodeActionsProp],
+  );
+
+  useMemo(() => {
```
zettca, line 121: This should be an effect - we're not memoizing anything

packages/lab/src/Flow/hooks/useNode.tsx
```diff
@@ -0,0 +1,189 @@
...
+          ]);
+          break;
+        default:
+          break;
+      }
+    },
+    [node, reactFlowInstance],
+  );
+
+  return {
```
zettca, line 168: We should probably memo this too

francisco-guilherme (approved):

PR #4214, opened 2024-07-03
feat: DragSplitView Component

rubenbarroshv:


francisco-guilherme:
We believe this component may not be the best fit for the UI Kit, as there are [better alternatives](https://www.npmjs.com/package/split-pane-react) with more extensive functionalities. Nonetheless, thank you for your contribution, @rubenbarroshv! 💪

PR #4223, opened 2024-07-05
test: tests added for canvas components & improvements

MEsteves22:
- Tests were added for the canvas components
- The canvas files were renamed to shorten the file names
- Missing types were added to the canvas components
- The samples were improved: decorators added to reduce sample code and code snippets improved
- Missing focus styles added 
- Missing <s>`onClickBack` handler</s> `backButtonProps` added to `HvCanvasToolbar`
- The width logic inside `HvCanvasPanel` was removed and now the `HvCanvasToolbar` width is defined by the user since it seemed like logic that should be on the client: let me know if you don't agree


MEsteves22:
Chromatic changes approved ✅

packages/lab/src/Canvas/Toolbar/Toolbar.tsx
```diff
@@ -61,7 +64,11 @@ export const HvCanvasToolbar = (props: HvCanvasToolbarProps) => {
     <div className={cx(classes.root, className)} {...others}>
       <div className={classes.back}>
         {backButton ?? (
-          <HvIconButton title={labels.back} variant="primaryGhost">
+          <HvIconButton
+            title={labels.back}
+            variant="primaryGhost"
+            onClick={onClickBack}
```
zettca, line 70: The component was just added, and this is already missing props 😝 

According to our guidelines, we should be calling this `onBackClick` instead:

> call handlers `on<Action>` or `on<Element><Action>` for sub-elements

We wouldn't need to name anything if we just did `backButton?: React.ReactNode` though 👀

zettca (approved):
LGTM 👍 

We should probably mark these components as WIP, to protect ourselves from breaking changes

packages/lab/src/Canvas/Toolbar/Toolbar.tsx
zettca: Actually... users shouldn't be using `onClick`, but `component={Link}` with `to={backRoute}`!

MEsteves22: I'll take a look a this

MEsteves22: I removed the click event and added `backButtonProps`: this way users can provide any prop they wish

PR #4233, opened 2024-07-17
docs(app): fix routing

zettca:


zettca:
This won't work because the `404.html` has to live in the root deployment. And we don't have a "root", but multiple deployments (eg. docs, app) **per pull request**

PR #4235, opened 2024-07-19
chore: `noImplicitAny` TS  rule enabled and errors fixed

MEsteves22:
- This PR adds `noImplicitAny: true` to `tsconfig` and fixes the errors raised by this rule: this ensures that the code is type safe unless we explicitly use the `any` type 
   - ℹ️Due to the complexity of some components (table, vertical navigation, among others), `any` types had to be used. In the future, when reviewing these components, the types as well as the components should definitely be reviewed carefully.


zettca (approved):
We're finally using TypeScript! 😝 💯

PR #4236, opened 2024-07-19
ci(release): change to bi-weekly releases

zettca:
- Change releases to run automatically mondays & thursdays at 05:00
  - We might want to eventually change this to weekly + manual, but let's run with this for now
  - The Nightly workflow doesn't make much sense anymore (all checks run on PR). We might want to eventually disable or rework it

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -48,17 +48,3 @@ jobs:
                 }
               ]
             }
-
-  release:
-    name: Release
-    if: github.event_name == 'schedule' && !contains(github.event.head_commit.message, '${ RELEASE_COMMIT_MESSAGE }')
-    needs: [tests, visual-tests]
-    runs-on: ubuntu-latest
-    steps:
-      - name: Trigger Release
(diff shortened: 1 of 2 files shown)


nantunes:
I had "all good" PRs in the past that failed after merge...

nantunes:
Maybe what is needed is to run the tests only if lerna reports any change?

zettca:
> I had "all good" PRs in the past that failed after merge...

Do you mean in those rare & unlucky cases where a branch is old (hasn't been rebased)? We're still performing the same "check" with the Nightly run

> Maybe what is needed is to run the tests only if lerna reports any change?

Not sure if I understood... To what purpose? To not run tests if it's unnecessary (saving CI compute)?

nantunes:
My bad... You are not disabling the nightly tests, only changing the release. May I ask why?

Also, on release days it will now run the tests twice (in the nightly and in the release workflow), right?

I'm not questioning the change, or against it, I'm just curious.

Basically, I thought you were trying to reduce the number of visual tests (as they rely on a third-party service).

PR #4241, opened 2024-07-23
chore: bump TypeScript to v5

MEsteves22:
- TypeScript version bump to latest v5.x version [as the build were already using v5](https://github.com/lumada-design/hv-uikit-react/actions/runs/10044666959/job/27760224774):
   - Fixes for v5.x were needed


zettca (approved):
Awesome stuff 👌 

I was expecting this upgrade to impact/break some of our components/stories/app, but it wasn't the case. I'll assume that Apps won't be impacted much 🤞

PR #4243, opened 2024-07-24
feat(Card): add Pentaho Card component

plagoa:


packages/pentaho/src/Card/stories/ButtonCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export interface HvButtonCardProps
+  extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+  icon?: React.ReactNode;
+  title: string;
+  description?: string;
+}
+
+export const ButtonCard = ({
```
MEsteves22, line 27: Maybe naming it `LinkCard` would be more accurate

packages/pentaho/src/Card/stories/ButtonCard.tsx
```diff
@@ -0,0 +1,44 @@
+import { css } from "@emotion/css";
+import { HvTypography, theme } from "@hitachivantara/uikit-react-core";
+
+import { HvCard } from "../Card";
```
MEsteves22, line 4: import from `@hitachivantara/uikit-react-pentaho` since it's a story

packages/pentaho/src/Card/stories/ImageCard.tsx
```diff
@@ -0,0 +1,39 @@
+import { css } from "@emotion/css";
+import { HvTypography, theme } from "@hitachivantara/uikit-react-core";
+
+import { HvCard } from "../Card";
```
MEsteves22, line 4: import from `@hitachivantara/uikit-react-pentaho` since it's a story

packages/pentaho/src/Card/stories/SimpleCard.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { css } from "@emotion/css";
+import { HvRadio, HvTypography, theme } from "@hitachivantara/uikit-react-core";
+
+import { HvCard } from "../Card";
```
MEsteves22, line 5: import from `@hitachivantara/uikit-react-pentaho` since it's a story

packages/pentaho/src/Card/stories/Stack.tsx
```diff
@@ -0,0 +1,148 @@
...
+  HvColorAny,
+  HvStack,
+  HvTag,
+  HvTagsInput,
+  HvTypography,
+  theme,
+} from "@hitachivantara/uikit-react-core";
+import { Copy, Delete, MoveFile, Tag } from "@hitachivantara/uikit-react-icons";
+
+import { ButtonCard } from "./ButtonCard";
```
MEsteves22, line 12: Maybe we could add a comment with link in order for users to know where to find these files

packages/pentaho/src/Card/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./Card";
```
MEsteves22, line 1: we are missing `HvCardMedia` export I think

packages/pentaho/src/Card/CardMedia/CardMedia.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useDefaultProps,
+} from "@hitachivantara/uikit-react-core";
+
+import { staticClasses, useClasses } from "./CardMedia.styles";
+
+export { staticClasses as cardMediaClasses };
+
+export type HvCardMediaClasses = ExtractNames<typeof useClasses>;
+
+export interface HvCardMediaProps extends HvBaseProps {
```
MEsteves22, line 14:
```suggestion
export interface HvCardMediaProps extends HvBaseProps<HTMLImageElement> {
```


We are also missing some descriptions below for props and the component ⬇️

packages/pentaho/src/Card/Card.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import { Meta, StoryObj } from "@storybook/react";
+import { HvCard, HvCardProps } from "@hitachivantara/uikit-react-pentaho";
+
+import { StackStory } from "./stories/Stack";
+import StackRaw from "./stories/Stack?raw";
+
+const meta: Meta<typeof HvCard> = {
+  title: "Pentaho/Card",
+  component: HvCard,
```
MEsteves22, line 9: we are missing `subcomponents: [HvCardMedia]`

packages/pentaho/src/Card/Card.tsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  return (
+    <Component
+      ref={ref}
+      className={cx(classes.root, className)}
+      onClick={handleClick}
+      onMouseDown={handleMouseDown}
+      {...others}
+    >
```
MEsteves22, line 59: shouldn't we also pass `disabled` to the component?

MEsteves22 (commented):
Looks really nice 🙏 

I just have the following comments (nothing major):

packages/pentaho/src/Card/Card.tsx
plagoa: I'm actually removing the disabled functionality as there's no requests for that feature at all and that would just add up to the confusion. If needed we'll add it later.

packages/pentaho/src/Card/CardMedia/CardMedia.styles.tsx
```diff
@@ -0,0 +1,13 @@
+import { createClasses } from "@hitachivantara/uikit-react-core";
+
+export const { staticClasses, useClasses } = createClasses("HvCardMedia", {
+  root: {
+    marginTop: "calc(var(--card-padding) * -1)",
```
MEsteves22, line 5: It looks like we no longer need this, right?

MEsteves22 (approved):

HQFOX:
I have noticed that because the component has the same name as the other card component. If you set default theme props for one they will be set for both. Right now it is not problematic because we don't have any overrides in the pentaho theme but it might be in the future.

MEsteves22:
> I have noticed that because the component has the same name as the other card component. If you set default theme props for one they will be set for both. Right now it is not problematic because we don't have any overrides in the pentaho theme but it might be in the future.

This is a good point 🙏 we should probably discuss about that before merging

PR #4247, opened 2024-07-29
docs(Flow): add sub group sample nodes

HQFOX:
A couple of icons are different from the samples because they are not in the uikit yet. I will change them as soon as they are added.

The sample main component name it's called Level, It was a placeholder but I haven't thought of a better one, suggestions are welcome 🙂


packages/lab/src/Flow/stories/Flow.stories.tsx
```diff
@@ -195,3 +197,19 @@ export const BaseHook: StoryObj<HvFlowProps> = {
...
+  parameters: {
+    docs: {
+      description: {
+        story: "This sample demonstrate the use of a sub flow",
+      },
+      source: {
+        code: SubFlowRaw,
+      },
+    },
+    // Enables Chromatic snapshot
```
MEsteves22, line 211: Since this is a custom node, I think we can disable Chromatic for this sample. No need to keep track of visual changes in my opinion.

packages/lab/src/Flow/stories/SubFlow/Level.styles.tsx
```diff
@@ -0,0 +1,43 @@
+import { createClasses, theme } from "@hitachivantara/uikit-react-core";
+
+export const { staticClasses, useClasses } = createClasses("HvLevel", {
```
MEsteves22, line 3: Since this is a story and not one of our components, maybe we should drop `createClasses` and create classes as we usually do for stories (using `css`). I don't think we need this "overhead" for the sample.

Also, I didn't find where is file is used. Is it needed?

packages/lab/src/Flow/stories/SubFlow/Level.tsx
```diff
@@ -0,0 +1,145 @@
...
+              >
+                {title}
+              </HvTypography>
+            }
+          />
+        </div>
+        {actions && (
+          <HvDropDownMenu
+            dataList={actions}
+            classes={{ root: css({ marginLeft: "auto" }) }}
```
MEsteves22, line 127: I think we should define all classes inside the `classes` object if possible.

packages/lab/src/Flow/stories/SubFlow/Node.tsx
```diff
@@ -0,0 +1,77 @@
+import { useMemo } from "react";
+import { css, cx } from "@emotion/css";
+import { NodeProps as ReactFlowNodeProps } from "reactflow";
+import { theme } from "@hitachivantara/uikit-styles";
+
+import { useHvNode } from "../../hooks";
```
MEsteves22, line 6: Should be imported from `lab` package inside stories.

packages/lab/src/Flow/stories/SubFlow/index.tsx
```diff
@@ -0,0 +1,264 @@
...
+    {...props}
+  />
+);
+
+const FactTableNode = (props: NodeProps) => (
+  <Node groupId="factTable" hierarchyData={factTableDataObject} {...props} />
+);
+
+const nodeGroups = {
+  dimensionTable: {
```
MEsteves22, line 164: Maybe we could provide some color to these groups

packages/lab/src/Flow/stories/SubFlow/FactTable.tsx
```diff
@@ -0,0 +1,89 @@
...
+    display: "flex",
+    flexDirection: "column",
+    justifyContent: "center",
+    backgroundColor: theme.colors.primary_20,
+    borderWidth: "1px",
+    borderColor: theme.colors.primary_20,
+    width: "100%",
+    height: "100%",
+  }),
+  header: css({
```
MEsteves22, line 22: I think we should have a bit more padding here

packages/lab/src/Flow/stories/SubFlow/Level.tsx
```diff
@@ -0,0 +1,145 @@
...
+    display: "flex",
+    flexDirection: "column",
+    justifyContent: "center",
+    backgroundColor: theme.colors.atmo1,
+    borderWidth: "1px",
+    borderColor: theme.colors.atmo3,
+    width: "100%",
+    height: "100%",
+  }),
+  header: css({
```
MEsteves22, line 26: I think we should have a bit more padding here

MEsteves22 (commented):
Really nice 🙏

packages/lab/src/Flow/stories/SubFlow/Level.styles.tsx
HQFOX: Leftover, It was not supposed to be on the commit, going to remove it.

packages/lab/src/Flow/stories/Flow.stories.tsx
MEsteves22: Right, I only looked at the comment 😅 We could actually remove `chromatic: { disableSnapshot: true },` altogether since all samples are disabled by default.

MEsteves22 (approved):

packages/lab/src/Flow/stories/SubFlow/Node.tsx
MEsteves22: I think this is missing from last commit

MEsteves22: Also, in stories, I think `theme` should come from the `core` package and not `styles` since users shouldn't need to use the `styles` package

francisco-guilherme:
The UI is looking good overall! However, I have noticed a couple of issues:

(PD reference)
Screenshot 2024-08-07 at 15 59 02

**Missing Shadows on Nodes:** The nodes appear to be lacking shadows.

**Button Padding:** The buttons seem to be too close to the margins.

Screenshot 2024-08-07 at 16 00 01

francisco-guilherme (approved):

PR #4248, opened 2024-07-30
fix(BottomPanel): tab title overflow

MEsteves22:
- `HvCanvasBottomPanel` component updates:
   - tabs title layout improved when content is overflowing (video below): `overflowActions` prop added
   - tests updated
   - `tab` prop renamed to `selectedTabId` (for consistency with the new toolbar tabs)
- Canvas template added to visual tests: to visually test the toolbar and side panel

https://github.com/user-attachments/assets/bd9f22ec-9dfe-4dab-8147-f472fdc79d38

### DIFF
--- a/packages/cli/src/templates/Canvas/index.tsx
+++ b/packages/cli/src/templates/Canvas/index.tsx
@@ -86,19 +86,17 @@ const Page = () => {
   const { selectedTable, openedTables, setOpenedTables, setSelectedTable } =
     useCanvasContext();
 
-  const bottomTabs = useMemo(
+  const bottomTabs: HvCanvasBottomPanelProps["tabs"] = useMemo(
     () =>
       openedTables?.map((table) => ({
         id: table.id,
-        title: (
+        title: (overflowing) => (
           <div className={classes.titleRoot}>
-            <Table />
-            <div className={classes.titleContainer}>
-              <HvOverflowTooltip data={table.label} />
-            </div>
+            {!overflowing && <Table />}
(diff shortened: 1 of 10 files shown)


packages/cli/src/templates/Canvas/index.tsx
```diff
@@ -86,19 +86,24 @@ const Page = () => {
...
-  const bottomTabs = useMemo(
+  const bottomTabs: HvCanvasBottomPanelProps["tabs"] = useMemo(
     () =>
       openedTables?.map((table) => ({
         id: table.id,
-        title: (
-          <div className={classes.titleRoot}>
-            <Table />
-            <div className={classes.titleContainer}>
+        title: (overflowing) =>
```
zettca, line 93: Is this the extra `<div>` & class to enforce the margin of the icons? I think it's better if we don't do that - we get a few more pixels for text and this only happens in an "extreme" scenario.

We could simplify it to something like this
```tsx
(overflowing) => (
  <div className={classes.titleRoot}>
    {!overflowing && <Table />}
    <HvOverflowTooltip data={table.label} />
  </div>
)
```

image

Also, the `marginRight` on the `rightActions` coming from `HvActionsGeneric` looks a bit excessive. This margin was intended for normal buttons

MEsteves22: I'll update this

MEsteves22:
PR updated ✅

HQFOX (approved):

zettca (approved):

PR #4249, opened 2024-07-30
fix(HvInlineEditor): typescript attribute recognition

HQFOX:
previously when using `HvInlineEditor` the spread attributes were not correctly recognized correctly so attributes of `HvInput` like `disableClear` would show as error with typescript. This intends to fix it even when you change the component.

### DIFF
--- a/packages/core/src/InlineEditor/InlineEditor.tsx
+++ b/packages/core/src/InlineEditor/InlineEditor.tsx
@@ -6,7 +6,7 @@ import { useControlled } from "../hooks/useControlled";
 import { useDefaultProps } from "../hooks/useDefaultProps";
 import { useTheme } from "../hooks/useTheme";
 import { HvInput, HvInputProps } from "../Input";
-import { HvBaseProps } from "../types/generic";
+import { fixedForwardRef, HvBaseProps, PolymorphicRef } from "../types/generic";
 import {
   HvTypography,
   HvTypographyProps,
@@ -20,8 +20,9 @@ export { staticClasses as inlineEditorClasses };
 
 export type HvInlineEditorClasses = ExtractNames<typeof useClasses>;
 
-export interface HvInlineEditorProps<T extends React.ElementType>
-  extends HvBaseProps<HTMLDivElement, "onBlur" | "onChange"> {
+export interface HvInlineEditorProps<
+  C extends React.ElementType = typeof HvInput,
+> extends HvBaseProps<HTMLDivElement, "onBlur" | "onChange"> {
   /** The value of the form element. */
   value?: string;
   /** The default value of the form element. */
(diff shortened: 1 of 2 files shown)


packages/core/src/InlineEditor/InlineEditor.tsx
```diff
@@ -20,7 +20,7 @@ export { staticClasses as inlineEditorClasses };
 
 export type HvInlineEditorClasses = ExtractNames<typeof useClasses>;
 
-export interface HvInlineEditorProps
+export interface HvInlineEditorProps<T extends React.ElementType>
```
MEsteves22, line 23: Really nice 🙏 I just think we should probably implement this like the other polymorphic components we have: button, typography,  time ago, etc.

packages/core/src/InlineEditor/InlineEditor.tsx
```diff
@@ -55,7 +56,13 @@ export interface HvInlineEditorProps
  * An Inline Editor allows the user to edit a record without making a major switch
  * between viewing and editing, making it an efficient method of updating a record.
  */
-export const HvInlineEditor = (props: HvInlineEditorProps) => {
+export const HvInlineEditor = fixedForwardRef(function HvInlineEditor<
+  C extends React.ElementType = typeof HvInput,
+>(
+  props: HvInlineEditorProps<C> &
```
MEsteves22, line 62: Can't we only do the following for `props`?

```suggestion
  props: HvInlineEditorProps<C>
```

HQFOX: I'm not seeing how, if I remove this it will complain that types of props are incompatible, maybe I'm miss understanding something.

MEsteves22: And what if we do something like the following and leverage `PolymorphicComponentRef` like our other polymorphic components? 🙏 

```tsx
export type HvInlineEditorProps<C extends React.ElementType = typeof HvInput> =
  PolymorphicComponentRef<
    C,
    {
      /** The value of the form element. */
      value?: string;
      /** The default value of the form element. */
      defaultValue?: string;
      /** Whether the Edit icon should always be visible */
      showIcon?: boolean;
      /** Variant of the HvTypography to display */
      variant?: HvTypographyVariants;
      /** Called when the input is blurred. */
      onBlur?: (
        event: React.FocusEvent<HTMLTextAreaElement | HTMLInputElement>,
        value: string,
      ) => void;
      /** Called when the input value changes. */
      onChange?: (event: React.SyntheticEvent, value: string) => void;
      /** Props passed to the HvButton component */
      buttonProps?: HvButtonProps;
      /** Props passed to the HvTypography text component */
      typographyProps?: HvTypographyProps;
      /** Whether the editor is disabled or not. */
      disabled?: boolean;
      /** A Jss Object used to override or extend the styles applied to the empty state component. */
      classes?: HvInlineEditorClasses;
      /** The placeholder value of the input. */
      placeholder?: string;
    }
  >;

export const HvInlineEditor = fixedForwardRef(function HvInlineEditor<
  C extends React.ElementType = typeof HvInput,
>(props: HvInlineEditorProps<C>, ref: PolymorphicRef<C>) {
  const {
    className,
    classes: classesProp,
    value: valueProp,
    defaultValue = "",
    showIcon,
    component: InputComponent = HvInput,
    variant = "body",
    placeholder = "Enter text",
    onBlur,
    onChange,
    onKeyDown,
    buttonProps,
    typographyProps,
    disabled,
    ...others
  } = useDefaultProps("HvInlineEditor", props);

  (...)
});
```

MEsteves22 (approved):

francisco-guilherme (approved):

PR #4253, opened 2024-07-31
ci(release): add custom GitHub releases

zettca:
Add custom GitHub Releases using [changelogithub](https://github.com/antfu/changelogithub)

- Disable `lerna` GitHub Releases generation (`createRelease`)
- Generate GitHub releases with [changelogithub](https://github.com/antfu/changelogithub)
  - version is picked up from core package
  - example:

image

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -1,5 +1,10 @@
 name: Dependencies
 description: Install UI Kit dependencies
+inputs:
+  install:
+    description: "Whether to npm install dependencies"
+    required: false
+    default: "true"
 
 runs:
   using: "composite"
@@ -10,6 +15,7 @@ runs:
         node-version: 18
 
     - name: Install dependencies
+      if: ${{ inputs.install == 'true' }}
       # ensure dev dependencies are installed
       run: npm ci --include=dev
       shell: bash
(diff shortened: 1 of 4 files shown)


HQFOX (approved):

francisco-guilherme:
I have changed this PR to Draft to allow for further investigation into the approach we should take.

After conducting some research, I couldn't find a clean solution or a library that directly accomplishes our goal. Consequently, I created a [Release Script](https://github.com/francisco-guilherme/hv-uikit-react/blob/master/scripts/release.sh) that generates these [Release Notes](https://github.com/francisco-guilherme/hv-uikit-react/releases) aligned with [MUI](https://github.com/mui/material-ui/blob/next/CHANGELOG.md).

I am not certain if this should be our final approach, but it outlines the steps we need to address:

- **Version and publish packages with Lerna:**
  - Lerna seems to be the best alternative to automatically version the packages following conventional commits.
  - We don't want Lerna to create tags by package, so we need to use the --noGitTagVersion flag.
- **Commit and tag changes:**
  - We want to create a single tag based on the core package version.
- **Create a GitHub release:**
  - Create the release notes by looping through all the packages and getting commits since the last released tag.
  - If new commits for the core package are found, trigger the release.

francisco-guilherme (approved):

PR #4268, opened 2024-08-09
feat(CodeEditor): IntelliSense support for XML

MEsteves22:
- IntelliSense support for XML added:
   - When the code editor is empty, a suggestion to add the root element appears when clicking on CTRL + SPACE. Clicking on it will add the root element.
   -  The code editor auto-completes a tag when closing it: for example, writing `>` after `<root` will automatically add `</root>`.
   - The XML code is validated against the schema and the errors are highlighted:
      - The schema is provided through the `xsdSchema` property.
      - The `xmllint-wasm` package is used for validation. In Vite we need to exclude it from `optimizeDeps`.
   - When the user types `<`, suggestions are shown with the possible tags. Clicking on a tag will add it.
   - Clicking on CTRL + SPACE inside an opened tag will show suggestions with the possible attributes for the tag. Clicking on an attribute will add it.
- Stories updated.
- In the future, if we need to add a new language, we only have to:
   - Create a new file in the `languages` folder and define the new utils.
   - Set the utils inside the plugins file. `handleMount` will take care of calling the utils for the language.

ℹ️ Let me know if you think more features are needed. I tried to add the most basic features.

### DIFF
--- a/packages/code-editor/src/CodeEditor/CodeEditor.tsx
+++ b/packages/code-editor/src/CodeEditor/CodeEditor.tsx
@@ -1,8 +1,9 @@
-import { useCallback, useEffect } from "react";
-import { Editor, EditorProps, useMonaco } from "@monaco-editor/react";
+import { useCallback, useEffect, useRef } from "react";
+import { Editor, useMonaco, type EditorProps } from "@monaco-editor/react";
 import { ExtractNames, useTheme } from "@hitachivantara/uikit-react-shared";
 
 import { staticClasses, useClasses } from "./CodeEditor.styles";
+import { languagePlugins } from "./plugins";
 
 export { staticClasses as codeEditorClasses };
 
@@ -17,6 +18,8 @@ export interface HvCodeEditorProps extends EditorProps {
   defaultValue?: string;
   /** The properties of the editor object in Monaco. */
   options?: EditorProps["options"];
+  /** XSD schema used to validate the code editor content when the language is set to `xml`. */
+  xsdSchema?: string;
 }
 
 const defaultCodeEditorOptions: EditorProps["options"] = {
@@ -37,34 +40,41 @@ const defaultCodeEditorOptions: EditorProps["options"] = {
 };
 
 /**
- * A wrapper to the React Monaco editor (https://github.com/suren-atoyan/monaco-react) with our styles.
+ * A wrapper to the [**React Monaco editor**](https://github.com/suren-atoyan/monaco-react) with our styles.
  * Please make sure you follow the instructions (found in the repository) to include the component.
- * Additional information regarding Tab trapping in Monaco, can be found here: https://github.com/microsoft/monaco-editor/wiki/Monaco-Editor-Accessibility-Guide#tab-trapping.
+ * Additional information regarding Tab trapping in Monaco, can be found [**here**](https://github.com/microsoft/monaco-editor/wiki/Monaco-Editor-Accessibility-Guide#tab-trapping).
  */
 export const HvCodeEditor = ({
   classes: classesProp,
-  defaultValue,
   options,
   editorProps,
+  defaultLanguage,
+  language: languageProp,
+  xsdSchema,
+  onMount: onMountProp,
+  beforeMount: beforeMountProp,
   ...others
 }: HvCodeEditorProps) => {
   const { classes } = useClasses(classesProp);
 
-  const { colors, selectedMode, selectedTheme, colorModes } = useTheme();
+  const language = languageProp ?? defaultLanguage;
 
-  const monaco = useMonaco();
+  const editorRef = useRef(null);
 
   // Merges the 2 objects together, overriding defaults with passed in options
-  const mergedOptions = {
+  const mergedOptions: EditorProps["options"] = {
     ...defaultCodeEditorOptions,
     ...options,
   };
 
-  const defineActiveThemes = useCallback(() => {
-    if (!monaco) return;
+  const { colors, selectedMode, selectedTheme, colorModes } = useTheme();
+  const monacoInstance = useMonaco();
+
+  const handleActiveThemes = useCallback(() => {
+    if (!monacoInstance) return;
 
     colorModes.forEach((mode) => {
-      monaco?.editor.defineTheme(`hv-${selectedTheme}-${mode}`, {
+      monacoInstance?.editor.defineTheme(`hv-${selectedTheme}-${mode}`, {
         base: colors?.type === "light" ? "vs" : "vs-dark",
         inherit: true,
         rules: [],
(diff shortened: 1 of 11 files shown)


MEsteves22:
PR accepted in Chromatic ✅ One change accepted: https://pentaho.chromatic.com/review?appId=66a1254d9d7ebf301ff1bebb&number=4268&type=linked&view=changes

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -36,35 +51,76 @@ const defaultCodeEditorOptions: EditorProps["options"] = {
...
+  xml: (event: any, editor: any, monaco: Monaco) =>
+    handleXmlKeyDown(event, editor, monaco),
+};
+const editorOptions: Record<string, object> = {
+  xml: xmlOptions,
+};
+
 /**
- * A wrapper to the React Monaco editor (https://github.com/suren-atoyan/monaco-react) with our styles.
+ * A wrapper to the <a href="https://github.com/suren-atoyan/monaco-react" target="_blank"><b>React Monaco editor</b></a> with our styles.
```
zettca, line 89: Markdown is simpler & more cross-compatible (eg.  it'll work in IDEs)

image

```suggestion
 * A wrapper to the [**React Monaco editor**](https://github.com/suren-atoyan/monaco-react) with our styles.
```

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -36,35 +51,76 @@ const defaultCodeEditorOptions: EditorProps["options"] = {
   },
 };
 
+/** Languages we have custom validations for. */
+const languages = { xml: "xml" };
+const completionProviders: Record<
+  string,
+  (monaco: Monaco, elements?: Elements) => object
+> = {
```
zettca, line 59: Looking at how this code is structured (all registering the `xml` key), it looks like a first-stage of a plugin system. We could evolve 

```tsx
const languagePlugins = [
  xmlLanguagePlugin(),
  // returns { completionProviders, validationMarkers, etc. }
  // or ["xml", { completionProviders, validationMarkers, etc. }]
]
```

Or maybe an object will be better for now (Array would be more versatile, but needlessly more complex?):

```tsx
const languagePlugins = {
  xml: xmlLanguagePlugin(), // returns completionProviders, validationMarkers, etc.
}
```

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -74,18 +130,82 @@ export const HvCodeEditor = ({
...
+    colors?.type,
+    colors?.atmo1,
+    colors?.secondary_60,
+  ]);
 
   useEffect(() => {
-    defineActiveThemes();
-  }, [selectedTheme, colorModes, defineActiveThemes]);
+    handleActiveThemes();
+  }, [selectedTheme, colorModes, handleActiveThemes]);
```
zettca, line 144: The blame on these useless dependencies is on me - let's not pass it onto you 😅 
```suggestion
  }, [handleActiveThemes]);
```

packages/code-editor/src/CodeEditor/stories/Xml.tsx
```diff
@@ -0,0 +1,322 @@
...
+    tree: Tree;
+    attributes: Attributes;
+  }>();
+  const [defaultExpandedKeys, setDefaultExpandedKeys] = useState<string[]>([]);
+
+  const editorRef = useRef<any>(null);
+
+  const buildTree = (xml: string) => {
+    const parser = new DOMParser();
+    const xmlDoc = parser.parseFromString(xml, "application/xml");
```
zettca, line 206: This looks like a pure function. Being that the case, we should hoist this outside the component.

The TreeView looks nice, but it adds lots of complexity to the sample 😬
... but by hoisting this I think out of the component, I think it'll be easier to follow 👍

packages/code-editor/src/CodeEditor/stories/Xml.tsx
```diff
@@ -0,0 +1,322 @@
...
+  };
+
+  const handleMount: HvCodeEditorProps["onMount"] = (editor) => {
+    editorRef.current = editor;
+  };
+
+  const handleOpenSearch: HvButtonProps["onClick"] = () => {
+    if (editorRef.current) {
+      editorRef.current.getAction("actions.find").run();
+    }
```
zettca, line 273:
```suggestion
    editorRef.current?.getAction("actions.find").run();
```

packages/code-editor/src/CodeEditor/stories/Xml.tsx
```diff
@@ -0,0 +1,322 @@
...
+      <Header onClickTree={handleClickTree} onClickSearch={handleOpenSearch} />
+      <HvCodeEditor
+        height={270}
+        language="xml"
+        value={editorValue}
+        onChange={(content) => setEditorValue(content ?? "")}
+        xsdSchema={xsdSchema}
+        onMount={handleMount}
+      />
+      {opened && (
```
zettca, line 297: Doesn't `open={opened}` already handle this? This is destroying more of the React tree than necessary

packages/code-editor/src/CodeEditor/stories/Xml.tsx
```diff
@@ -0,0 +1,322 @@
...
+      <Code />
+    </div>
+    <div className={classes.actionContainer}>
+      <HvButton variant="primaryGhost" onClick={onClickSearch}>
+        Search
+      </HvButton>
+      <HvButton variant="primaryGhost" onClick={onClickTree}>
+        XML Tree
+      </HvButton>
+    </div>
```
zettca, line 98: We could simplify this with a `flex: 1` "separator - it's what we do in our ActionBar samples (+ the `gap`)
```suggestion
    <HvTypography variant="label">XML</HvTypography>
    <Code />
    <div style={{ flex: 1 }} />
    <HvButton variant="primaryGhost" onClick={onClickSearch}>
      Search
    </HvButton>
    <HvButton variant="primaryGhost" onClick={onClickTree}>
      XML Tree
    </HvButton>
```

zettca (approved):
Nice work 💯 

Some minor suggestions/ideas:

packages/code-editor/src/CodeEditor/CodeEditor.tsx
MEsteves22: I'll update this 🙏

packages/code-editor/src/CodeEditor/stories/Xml.tsx
MEsteves22: Yep, don't know why I added this here 😅 I'll update

packages/code-editor/src/CodeEditor/stories/Xml.tsx
MEsteves22: I'll update

packages/code-editor/src/CodeEditor/stories/Xml.tsx
MEsteves22: I'll update 🙏

packages/code-editor/src/CodeEditor/CodeEditor.tsx
MEsteves22: I'll remove them 🙏

packages/code-editor/src/CodeEditor/CodeEditor.tsx
MEsteves22: Makes sense 🙏 I'll update the code to follow the object approach for now.

MEsteves22:
I updated the PR with the suggestions ✅

HQFOX (approved):

PR #4272, opened 2024-08-13
feat(Flow): add groups to subflow sample

HQFOX:
I tried creating some hooks in `useHvNode` so that the you wouldn't need to implement it in every node, like a onDragEnd event but was unsuccessful as it would always trigger to many re-renders. If you have an idea to remove some logic from the `Node` or `NodeGroup` let me know.

I also noted that we use the type property of the `reactflow` nodes in order to choose which component to use in a node, (i.e.: the Dimension Table node is of type: "dimensionTable"), but in order to create a group this property has to be 'group' creating multiple types of groups would have to be done some other way.

### DIFF
--- a/packages/lab/src/Flow/base.ts
+++ b/packages/lab/src/Flow/base.ts
@@ -232,8 +232,6 @@ export const flowStyles = css`
   .react-flow__node-input,
   .react-flow__node-output,
   .react-flow__node-group {
-    border-radius: 0px 16px 16px 16px;
-    border: unset;
     color: ${theme.colors.secondary};
     text-align: center;
   }
@@ -258,7 +256,6 @@ export const flowStyles = css`
   .react-flow__node-group.selectable.selected,
   .react-flow__node-group.selectable:focus,
   .react-flow__node-group.selectable:focus-visible {
-    box-shadow: 0 0 0 0.5px #1a192b;
   }
   .react-flow__node-group {
     background-color: rgba(240, 240, 240, 0.25);
--- a/packages/lab/src/Flow/base.ts
+++ b/packages/lab/src/Flow/base.ts
@@ -232,13 +232,10 @@ export const flowStyles = css`
   .react-flow__node-input,
   .react-flow__node-output,
   .react-flow__node-group {
-    padding: ${theme.space.sm};
-    border-radius: ${theme.radii.round};
-    width: 150px;
+    border-radius: 0px 16px 16px 16px;
+    border: unset;
     color: ${theme.colors.secondary};
     text-align: center;
-    border: 1px solid ${theme.colors.negative};
-    background-color: ${theme.colors.negative_20};
   }
   .react-flow__node-default::before {
     content: "Unknown node type";
@@ -248,7 +245,6 @@ export const flowStyles = css`
   .react-flow__node-input.selectable:hover,
   .react-flow__node-output.selectable:hover,
   .react-flow__node-group.selectable:hover {
-    box-shadow: 0 1px 4px 1px rgba(0, 0, 0, 0.08);
   }
   .react-flow__node-default.selectable.selected,
   .react-flow__node-default.selectable:focus,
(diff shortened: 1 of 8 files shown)


packages/lab/src/Flow/base.ts
```diff
@@ -232,13 +232,8 @@ export const flowStyles = css`
   .react-flow__node-input,
   .react-flow__node-output,
   .react-flow__node-group {
```
MEsteves22, line 234: This style was actually needed when showing an invalid node:

Screenshot 2024-08-20 at 11 54 01

We should recover this:

```css
  .react-flow__node-default,
  .react-flow__node-input,
  .react-flow__node-output {
     padding: ${theme.space.sm};
     border-radius: ${theme.radii.round};
     width: 150px;
     color: ${theme.colors.secondary};
     text-align: center;
     border: 1px solid ${theme.colors.negative};
     background-color: ${theme.colors.negative_20};
  }
  .react-flow__node-group {
    color: ${theme.colors.secondary};
    text-align: center;
  }
```

packages/lab/src/Flow/stories/SubFlow/NodeGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+            borderRadius: theme.radii.round,
+          }}
+        >
+          <Fullscreen />
+        </div>
+      </NodeResizeControl>
+      <div
+        className={cx(classes.headerBackground, css({ zIndex: node?.zIndex }))}
+      >
+        <div className={cx(classes.header)}>
```
MEsteves22, line 125: `cx` not needed

packages/lab/src/Flow/stories/SubFlow/NodeGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+} from "@hitachivantara/uikit-react-core";
+import { Fullscreen } from "@hitachivantara/uikit-react-icons";
+import { useHvNode } from "@hitachivantara/uikit-react-lab";
+
+const dashedBorder = `1px dashed ${theme.colors.atmo4}`;
+const containerBorderRadius = "0 16px 16px 16px";
+const minWidth = 300;
+const minHeight = 200;
+
+const classes = {
```
MEsteves22, line 22: We have styles inside the component and then defined here inside `classes`. For consistency, I think we should define all the styles inside `classes` when possible 🙏

packages/lab/src/Flow/stories/SubFlow/NodeGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+          {actions && (
+            <HvDropDownMenu
+              dataList={actions}
+              classes={{ root: classes.actions }}
+            />
+          )}
+        </div>
+      </div>
+      <div className={classes.content} />
+      <div />
```
MEsteves22, line 143: Is this needed?

packages/lab/src/Flow/stories/SubFlow/NodeGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export const NodeGroup = ({ id: idProp, groupId, actions = [] }: NodeProps) => {
+  const { title, icon, id, node, intersections, setNodeParent } = useHvNode({
+    id: idProp,
+    groupId,
+  });
+
+  const [draggingFlag, setDraggingFlag] = useState(false);
+
+  useEffect(() => {
```
MEsteves22, line 82: I had a bit of trouble understanding why this code is needed (same for the `useEffect` inside `Node.tsx`). I think adding a comment would be nice 🙏

packages/lab/src/Flow/hooks/useFlowNode.ts
```diff
@@ -126,10 +138,42 @@ export function useFlowNodeUtils<NodeData = any>(id?: string) {
...
     () => ({
       setNodeData,
+      setNodeParent,
     }),
-    [setNodeData],
+    [setNodeData, setNodeParent],
   );
 }
+
+export function useFlowNodeGetIntersections<NodeData = any>(id?: string) {
```
MEsteves22, line 173: I think naming it `useFlowNodeIntersections` would be enough

MEsteves22 (commented):
LGTM 🙏 I just have the following comments:

packages/lab/src/Flow/stories/SubFlow/NodeGroup.tsx
HQFOX: I agree,  added some comments, tell me what you think

MEsteves22 (approved):

PR #4273, opened 2024-08-13
feat(utils): add `@hitachivantara/uikit-react-utils` package

zettca:
- add `@hitachivantara/uikit-react-utils` package
- move `classes` utilities from `shared` to `utils` (that were changed in https://github.com/lumada-design/hv-uikit-react/pull/4255)
- also move other "base" utilities: `useDefaultProps`, `mergeStyles`

> [!IMPORTANT]
> Refactor is in different commit for ease of review. Please squash when merging

TODO:
- ~decide which other `core/src/utils` utils should be moved to `@hitachivantara/uikit-react-utils`~
- ~clean-up/remove long unused and/or internal core utils being exported?~
we can address this later 👆


packages/utils/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@hitachivantara/uikit-react-utils",
+  "version": "0.1.0",
```
nantunes, line 3: Version bellow 1.0 is intentional? It means consumers should consider the exposed APIs as unstable?

PR #4283, opened 2024-08-20
refactor(ToolbarTabs): accessibility and style

MEsteves22:
- Toolbar tabs refactored to improve accessibility and styles updated according to PD specs

⚠️❗️ **Please, squash when merging**


MEsteves22:
PR already accepted in Chromatic ✅ (one change accepted)

plagoa (approved):

packages/pentaho/src/Canvas/ToolbarTabs/ToolbarTabs.styles.tsx
```diff
@@ -1,15 +1,16 @@
...
 export const MIN_TAB_WIDTH = 120;
-const MAX_TAB_WIDTH = 220;
+export const MAX_TAB_WIDTH = 220;
+export const DROPDOWN_MENU_WIDTH = 64;
 const TAB_HEIGHT = 32;
-const TAB_LABEL_HEIGHT = 24;
+const CLOSE_ICON_SIZE = 32;
+const TAB_ICON_SIZE = 16;
+const TAB_PADDING = theme.space.xs;
+const TAB_COLOR = "#EFF6FF"; // TODO - change HEX for token when available (dark/light)
```
zettca, line 13: This doesn't look good in dark mode:

image

Don't we have a dark color that we can use?
If we do, we can define a CSS var (eg. `--tabColor`) with a fallback value, and override it in the `pentahoPlus` theme under `components.HvCanvasToolbarTabs.classes.root`
If we don't... I think we should decide a color from the palette ourselves, or keep what we previously had

What do you think?

packages/pentaho/src/Canvas/ToolbarTabs/ToolbarTabs.tsx
```diff
@@ -255,107 +240,82 @@ export const HvCanvasToolbarTabs = forwardRef<
...
+                          handleEdit(event, value, tab.id)
+                        }
+                        // We don't want the arrow keys to trigger the tab navigation
+                        onKeyDown={(e) => e.stopPropagation()}
+                      />
+                    )}
+                    <div className={classes.closeIconContainer}>
+                      <CloseXS
+                        aria-hidden
+                        data-testid="delete-icon" // For testing purposes
```
zettca, line 293: Icons have a `data-name` that we can use for assertions ([TagsInput example](https://github.com/lumada-design/hv-uikit-react/blob/master/packages/core/src/TagsInput/TagsInput.test.tsx#L41)). We would need to `getByRole("tab")` first though

packages/pentaho/src/Canvas/ToolbarTabs/ToolbarTabs.styles.tsx
```diff
@@ -22,85 +23,80 @@ export const { staticClasses, useClasses } = createClasses(
...
+          backgroundColor: TAB_COLOR,
+          borderColor: TAB_COLOR,
+          "&:first-of-type": { borderEndStartRadius: theme.radii.base },
+        },
+
+      // Hide icon when editor is hovered and focused
+      [`&:has($tabLabelEditor:hover) $tabIconContainer, &:has(.${toolbarTabEditorClasses.edit}) $tabIconContainer`]:
+        {
+          display: "none",
+        },
```
zettca, line 61: The UI feel a bit janky as it's "jumping around" on hover - maybe we should only hide the icon when we're editing?

zettca (commented):
Nice work, looks much better 💯 

Some small notes:

zettca (approved):

packages/pentaho/src/Canvas/ToolbarTabs/ToolbarTabs.styles.tsx
MEsteves22: As discussed, we'll talk with PD first 🙏

packages/pentaho/src/Canvas/ToolbarTabs/ToolbarTabs.tsx
MEsteves22: I'll address this in another PR

packages/pentaho/src/Canvas/ToolbarTabs/ToolbarTabs.styles.tsx
MEsteves22: I'll address this in another PR

PR #4289, opened 2024-08-21
test: move visual tests to components

zettca:
- move visual tests to components' story files, reducing file name & metadata duplication
  - stories whose names start with `Test*` will be hidden from `master` docs - and will always be hidden in "Docs" page
- add Hero page in `app` for debugging


packages/pentaho/src/Canvas/BottomPanel/BottomPanel.stories.tsx
```diff
@@ -33,3 +37,138 @@ export const Main: StoryObj<HvCanvasBottomPanelProps> = {
     return <MainStory {...args} />;
   },
 };
+
+export const Test: StoryObj = {
```
MEsteves22, line 41: We need to update the bottom panels tabs to use this story now: https://github.com/lumada-design/hv-uikit-react/blob/master/packages/pentaho/src/Canvas/BottomPanel/BottomPanel.spec.tsx#L12

zettca: yes, I noticed 🙈

MEsteves22 (approved):

PR #4293, opened 2024-08-21
feat(Card): add pentaho plus styles

plagoa:


packages/core/src/Card/stories/ExpandableCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+      >
+        <HvTypography>This is my hidden content</HvTypography>
+      </HvCardContent>
+
+      <div
+        style={{ display: "flex", justifyContent: "center", paddingBottom: 0 }}
+      >
+        <HvIconButton
+          title={isOpen ? "Collapse" : "Expand"}
+          onClick={() => setIsOpen(!isOpen)}
```
MEsteves22, line 53: Looking at [these docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded), I think we need to use the `aria-expanded` prop on the button and also use `aria-controls` (and set an `id` to the `HvCardContent`) 🙏

packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -730,6 +730,60 @@ const pentahoPlus = makeTheme((theme) => ({
         },
       },
     },
+    HvCard: {
```
MEsteves22, line 733: Shouldn't we remove the Pentaho card from the Pentaho package?

packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -730,6 +730,60 @@ const pentahoPlus = makeTheme((theme) => ({
...
+      classes: {
+        root: {
+          outline: `1px solid ${theme.palette.slate[200]}`,
```
MEsteves22, line 736: I think we should put a `TODO` on the colors from palette we are using to not forget we need to use a token when available.

MEsteves22 (commented):
LGTM 👌🏻

plagoa:
@MEsteves22 PR reviewed with feedback

zettca (approved):
Can we mark the removing Card commit as a breaking change, so that users (maybe none yet) get better visibility, and for us to test the versioning + release messaging

PR #4300, opened 2024-08-23
feat(input): enable passing props to search and password adornments & fix disabled adornment

MEsteves22:
- Fixes: when the input was `disabled`, the search and reveal adornments were not disabled.
- Adds: `searchButtonProps` and `revealButtonProps` to enable passing props to the search and reveal adornments. This was needed since some users may need to disabled the search adornment for instance. Mistakingly the `disableRevealPassword`, `disableSearchButton`, and `disableClear` were hiding the buttons and not disabling them.


MEsteves22:
PR approved in Chromatic already ✅

packages/core/src/Forms/Adornment/Adornment.tsx
```diff
@@ -28,6 +28,8 @@ export interface HvAdornmentProps
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
   /** If this property is defined the adornment visibility will be exclusively controlled by this value. */
   isVisible?: boolean;
+  /** Whether the adornment is disabled or not. */
+  disabled?: boolean;
```
zettca, line 32: I think we should be getting this value from the context `HvFormElementContext` below, same as `status`

packages/core/src/Input/Input.tsx
```diff
@@ -614,6 +614,7 @@ export const HvInput = forwardRef<InputElement, HvInputProps>((props, ref) => {
         onClick={hasOnEnter ? handleSearch : undefined}
         aria-label={labels?.searchButtonLabel}
         icon={<Search />}
+        disabled={disabled}
```
zettca, line 617: HvFormElementContext

zettca (approved):

PR #4306, opened 2024-09-02
feat: add icon container component

HQFOX:


packages/icons/src/PentahoIconWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return { width: baseSize, height: baseSize };
+};
+
+const StyledIconContainer = styled("div")({
+  display: "flex",
+  justifyContent: "center",
+  alignItems: "center",
+});
+
+export const PentahoIconWrapper = ({
```
zettca, line 41: I really don't like the name `Wrapper` 🤷 
Sometimes it's just a container (and `Container` is a clearer name). But most of the times it's more than a container, and the name is redundant & confusing

Here's it's just unnecessary. And we're in need of some docs

```suggestion
/** DOCS! */
export const PentahoIcon = ({
```

packages/icons/src/PentahoIconWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+  FontAwesomeIconProps,
+} from "@fortawesome/react-fontawesome";
+import { getColor, HvColorAny, HvSize } from "@hitachivantara/uikit-styles";
+
+import { IconSize, svgSizeMap } from "./IconBase";
+
+export interface PentahoIconWrapperProps
+  extends Omit<FontAwesomeIconProps, "color" | "size"> {
+  size?: HvSize;
+  iconSize?: IconSize;
```
zettca, line 14: If we're moving to the `size` prop, there's shouldn't be an `iconSize`

We could do `HvSize | IconSize` for increased compatibility though. Also, allowing `number` would be nice

```suggestion
  size?: HvSize | IconSize | number;
```

packages/icons/src/PentahoIconWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+  };
+
+  const iconComponent = (
+    <FontAwesomeIcon
+      icon={icon}
+      size={getSVGsizeTranslation(iconSize, size)}
+      fontSize={iconSize ? 16 : fontSize}
+      style={getSVGSize(iconSize ?? size)}
+      color={colorProp ? getColor(colorProp) : "inherit"}
```
zettca, line 74: `getColor` deals with fallback values

```suggestion
      color={getColor(colorProp, "inherit")}
```

packages/icons/src/PentahoIconWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+      {...others}
+    />
+  );
+
+  return container ? (
+    <StyledIconContainer
```
zettca, line 80: Did you try using the existing `StyledIconBase` component? Are there problems?

It would be nice if we could use the same component & behaviour

zettca (commented):
Looking nice 💪

packages/icons/src/PentahoIconWrapper.tsx
HQFOX: I agree, I will add number when you pr with icon base sizes enters so I can use the same logic.

packages/icons/src/PentahoIcon.tsx
```diff
@@ -0,0 +1,77 @@
...
+  color?: HvColorAny;
+  /** Gives the icon padding.  */
+  container?: boolean;
+}
+
+/**
+ * A wrapper for the [**react-fontawesome**](https://github.com/FortAwesome/react-fontawesome) component
+ * to easily apply our theme sizes colors and margins to the `font-awesome` icons.
+ */
+export const PentahoIcon = ({
```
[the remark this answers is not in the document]
HQFOX: I added Pentaho because I don´t think that is supposed to be mixed usage of icons, users using Next DS should use the current icons and Pentaho users should use the `font-awsome` icons. `HvExternalIcon` seems to general since this currently only works for `font-awsome`. I could change it to `HvFontAwesomeIcon` as that is the name of the react component it wraps, but it seems strange to me.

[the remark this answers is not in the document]
MEsteves22: Okay, got it. If the rest of the team is okay with the name, I don't mind. I thought this component could be used by anyone regardless of the theme. The `Hv` prefix is just missing though. We should add it.

docs/foundation/Icons/Icons.stories.tsx
```diff
@@ -114,6 +127,103 @@ export const IconSprites: StoryObj<HvIconSpriteProps> = {
...
+];
+
+const fontSizes = ["8", "16", "32", "48"];
+
+export const PentahoIcons = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "Example usage of Font Awesome icons to be used with Pentaho Design System",
```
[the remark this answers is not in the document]
MEsteves22: Nevermind this according to[ this thread.](https://github.com/lumada-design/hv-uikit-react/pull/4306#discussion_r1750574893)

packages/icons/src/IconContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * A component used to contain icons the established in the established margins.
+ * This allows for the use of external icons with our components that expect an icon container of a specific size.
+ *
+ */
+export const HvIconContainer = (
+  props: HvIconContainerProps & {
+    iconName?: string;
+    ref?: React.Ref<HTMLDivElement>;
+  },
+) => {
```
zettca, line 57: This doesn't depend on `iconName` - we shouldn't have it.
Also, we should `forwardRef` and pass it to the component below

packages/core/src/IconContainer/IconContainer.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Main: StoryObj<typeof HvIconContainer> = {
+  parameters: {
+    chromatic: { disableSnapshot: false },
+  },
+  args: {
+    size: undefined,
+  },
+  argTypes: {
+    size: { control: { type: "select" } },
```
zettca, line 20: This is not working, and it'd be nice if it did. I think we should leave it as the default input, so that we can pass in a `number` too

image

packages/core/src/IconContainer/IconContainer.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import { faAtom } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { Meta, StoryObj } from "@storybook/react";
+import { HvIconContainer } from "@hitachivantara/uikit-react-icons";
+
+const meta: Meta<typeof HvIconContainer> = {
+  title: "Components/Icon Container",
```
zettca, line 7: I think we should group this with the other icons, in the "Foundation" category. It might not be the perfect spot, but having this in the middle of the Components makes it harder to find/confusing IMO

HQFOX: That seems weird to me because Foundation is more of a documentation section with guides than a component showcase, and in this pr is already a story more in form of a guide that explains the usage of the component in that section. 
I can just rename it to the component name but then in my opinion is harder for a user to know that when using a external icon they need to use the icon container or the component breaks. What do you think?

zettca: all right - makes sense 👍

docs/foundation/Icons/Icons.stories.tsx
```diff
@@ -115,6 +124,20 @@ export const IconSprites: StoryObj<HvIconSpriteProps> = {
   },
 };
 
+export const ExternalIconsStory: StoryObj<HvIconContainerProps> = {
+  name: "External Icons",
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'Most of our components that use icons are expecting them to have certain size containers. This component allows you to use external components without breaking the component behavior. Here we exemplify this usage with <a href="https://docs.fontawesome.com/v5/web/use-with/react/">`Font Awesome`</a> icons.',
```
MEsteves22, line 133: We should specify which component we are referring to: `HvIconContainer`

package.json
```diff
@@ -49,6 +49,11 @@
     "@emotion/css": "^11.11.0",
     "@emotion/react": "^11.11.1",
     "@emotion/styled": "^11.11.0",
+    "@fortawesome/fontawesome-svg-core": "^6.6.0",
+    "@fortawesome/free-brands-svg-icons": "^6.6.0",
```
MEsteves22, line 53: We are not using these packages (`@fortawesome/free-solid-svg-icons` and `@fortawesome/free-regular-svg-icons`). I think they should be removed.

packages/core/src/IconContainer/IconContainer.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+  component: HvIconContainer,
+};
+export default meta;
+
+export const Main: StoryObj<typeof HvIconContainer> = {
+  parameters: {
```
MEsteves22, line 13: Let's remove `parameters` here: we are already testing this component visually inside `docs/foundation/Icons/Icons.stories.tsx` 🙏

packages/icons/src/IconContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+   */
+  size?: HvSize | IconSize | number;
+  /**
+   * Icon to render.
+   */
+  children?: ReactNode;
+}
+
+/**
+ * A component used to contain icons the established in the established margins.
```
MEsteves22, line 48: I don't understand this phrase

MEsteves22 (commented):
🙏

I rebased your PR since [this one](https://github.com/lumada-design/hv-uikit-react/pull/4336) was merged.

docs/foundation/Icons/Icons.stories.tsx
```diff
@@ -115,6 +124,20 @@ export const IconSprites: StoryObj<HvIconSpriteProps> = {
   },
 };
 
+export const ExternalIconsStory: StoryObj<HvIconContainerProps> = {
+  name: "External Icons",
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'Most of our components that use icons are expecting them to have certain size containers. The `HvIconContainer` component component allows you to use external components without breaking the component behavior. Here we exemplify this usage with <a href="https://docs.fontawesome.com/v5/web/use-with/react/">`Font Awesome`</a> icons.',
```
MEsteves22, line 133: `component` word repeated 👀

zettca (approved):

PR #4307, opened 2024-09-02
fix(Card): set Card height

plagoa:
set `HvCard` height to fit its contents.

### DIFF
--- a/packages/core/src/Card/Card.styles.tsx
+++ b/packages/core/src/Card/Card.styles.tsx
@@ -5,6 +5,7 @@ import { outlineStyles } from "../utils/focusUtils";
 
 export const { staticClasses, useClasses } = createClasses("HvCard", {
   root: {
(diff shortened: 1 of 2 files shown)


zettca (changes requested):
This is a visual breaking change that will affect many apps. We shouldn't do this outside of a major bump.

If we're aligning this behaviour, it should be the other way around. We'll most likely want for the cards to be (**by default**) the same height (`100%`), not to be as short as possible. Eg:

image

plagoa:
That

> This is a visual breaking change that will affect many apps. We shouldn't do this outside of a major bump.
> 
> If we're aligning this behaviour, it should be the other way around. We'll most likely want for the cards to be (**by default**) the same height (`100%`), not to be as short as possible. Eg:
> 
> image

That's fair, I'll close the PR.

PR #4310, opened 2024-09-03
chore: add status to card variants

plagoa:
- add status colors to the variants sample
- remove hover effect on non-selectable cards

### DIFF
--- a/packages/core/src/Card/stories/Variants.tsx
+++ b/packages/core/src/Card/stories/Variants.tsx
@@ -9,7 +9,12 @@ import {
   HvTypography,
   theme,
 } from "@hitachivantara/uikit-react-core";
-import { LocationPin, WorldGlobe } from "@hitachivantara/uikit-react-icons";
+import {
+  Level0Good,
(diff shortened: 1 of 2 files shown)


packages/core/src/Card/stories/Variants.tsx
```diff
@@ -71,7 +71,7 @@ export const Variants = () => {
         </HvCardContent>
       </HvCard>
 
-      <HvCard className={classes.card} bgcolor="atmo1">
+      <HvCard className={classes.card} bgcolor="atmo1" statusColor="negative">
```
zettca, line 74: As we're giving it a semantic color, we could also give it a semantic icon (`HvCard`'s `icon` prop) in a couple of cards

zettca (approved):
LGTM

packages/core/src/Card/stories/Variants.tsx
plagoa: Yes, I added this.

HQFOX (approved):

PR #4313, opened 2024-09-03
feat(icons): add rotate animation & size prop

zettca:
- add rotation transition to icons, as it's often used
- add `size` prop to replace `iconSize`, which is already deprecated without replacement
  - accepts any `number`, `HvSize` (`"xs"`, `"sm"`, etc.), and `IconSize` (for backwards compatibility)
  - svg container (`div`) grows dynamically according to the set svg size (eg. 120px svg -> 128px container)
- if the `size` prop is used, the colors are inherited by default (for forwards compatibility)

> [!IMPORTANT]
> Please don't squash the commits when merging

### DIFF
--- a/packages/icons/src/IconBase.tsx
+++ b/packages/icons/src/IconBase.tsx
@@ -7,55 +7,13 @@ import {
   theme,
 } from "@hitachivantara/uikit-styles";
 
-import { isSemantic, isXS } from "./utils";
-
-const calcSize = (size: number, isLarger = false) =>
-  isLarger ? size + 8 : size;
+import { getSizeStyles } from "./utils";
 
 const getColorVars = (colorArray: string[]) => {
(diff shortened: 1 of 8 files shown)


packages/icons/src/IconBase.tsx
```diff
@@ -91,11 +59,13 @@ export interface IconBaseProps
    * @example ["brand", "inherit"]
    */
   color?: HvColorAny | HvColorAny[];
-  /** Sets one of the standard sizes of the icons */
+  /** The size of the SVG icon */
```
HQFOX, line 62: It seems like an unexpected behavior that using the prop size can change the color, maybe we could add something a note to the prop description?

HQFOX (approved):

packages/icons/src/IconBase.tsx
zettca: good point - added

PR #4314, opened 2024-09-03
refactor(icons): remove div container

zettca:
testing the impact of replacing the `div` container with `svg` ([again](https://github.com/lumada-design/hv-uikit-react/pull/3615))


zettca:
significant impact - most samples changed but due to very few components (eg. `HvIconButton`, `HvWarningText`)

we'll check again next year!

PR #4319, opened 2024-09-06
feat(Button): review `HvButton` styles according to latest PD specs

plagoa:
- added `primaryStrong` and `primarySubtle` tokens
- also exporting the `amber` colors as those were not being exported.

### DIFF
--- a/packages/styles/src/themes/pentahoPlus.ts
+++ b/packages/styles/src/themes/pentahoPlus.ts
@@ -62,6 +62,8 @@ const pentahoPlus = makeTheme((theme) => ({
         pp: {
           primary: blue[600],
           primaryAction: blue[700],
+          primaryStrong: blue[800],
+          primarySubtle: blue[200],
           primaryDimmed: blue[100],
           success: green[600],
           successAction: green[700],
@@ -94,7 +96,7 @@ const pentahoPlus = makeTheme((theme) => ({
           bgSurface: slate[50],
           bgActive: slate[200],
           bgHover: blue[100],
-          bgDisabled: slate[200],
+          bgDisabled: neutral[200],
           bgOverlay: `color-mix(in srgb, ${slate[900]} 60%, transparent)`,
           dimmer: "#FFFFFF",
         },
@@ -139,6 +141,8 @@ const pentahoPlus = makeTheme((theme) => ({
         pp: {
           primary: blue[400],
           primaryAction: blue[300],
+          primaryStrong: blue[200],
+          primarySubtle: blue[900],
           primaryDimmed: blue[950],
           success: green[600],
           successAction: green[500],
@@ -171,7 +175,7 @@ const pentahoPlus = makeTheme((theme) => ({
           bgSurface: slate[900],
           bgActive: slate[900],
           bgHover: blue[950],
-          bgDisabled: slate[900],
+          bgDisabled: neutral[900],
           bgOverlay: `color-mix(in srgb, ${slate[900]} 40%, transparent)`,
           dimmer: "#000000",
         },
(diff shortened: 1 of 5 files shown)


packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -390,41 +390,114 @@ const pentahoPlus = makeTheme((theme) => ({
     HvButton: {
       classes: {
         root: {
-          borderRadius: theme.radii.full,
+          border: "none",
+          borderRadius: 48,
```
zettca, line 394: why move away from our theme token? aren't they the same?

packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -390,41 +390,114 @@ const pentahoPlus = makeTheme((theme) => ({
...
-        disabled: {
-          "&:active": {
-            boxShadow: "none",
+          [`&[data-color="negative"]`]: {
+            "&:hover": {
+              backgroundColor: theme.palette.red[700],
+            },
+            "&:active": {
+              backgroundColor: theme.palette.red[800],
+            },
```
zettca, line 410: These (and most other) colors aren't taking the dark theme into consideration.

We could use something like [`light-dark()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark) for now, or wait for the new semantic tokens to be set

Either way, we shouldn't disregard the theming

packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -390,41 +390,114 @@ const pentahoPlus = makeTheme((theme) => ({
...
         root: {
-          borderRadius: theme.radii.full,
+          border: "none",
+          borderRadius: 48,
           padding: theme.spacing(0, "sm"),
-          "&:active": {
-            boxShadow: `inset 0 1px 2px 0 #00000033`, // 33 = 20% opacity
+          [`&[data-color="positive"]`]: {
+            "&:hover": {
+              backgroundColor: theme.palette.green[700],
```
zettca, line 398: I think we should be using the colors directly, like elsewhere. These probably give a false sense old security (eg. they come from `theme.`, so they must be themmable) but they're not. It's just `green[700]` with extra steps.
I believe this would also be a blocker if we want to make the theme dynamic (not sure yet)

```suggestion
              backgroundColor: green[700],
```

packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -390,41 +390,114 @@ const pentahoPlus = makeTheme((theme) => ({
     HvButton: {
       classes: {
         root: {
-          borderRadius: theme.radii.full,
+          border: "none",
+          borderRadius: 48,
           padding: theme.spacing(0, "sm"),
-          "&:active": {
-            boxShadow: `inset 0 1px 2px 0 #00000033`, // 33 = 20% opacity
+          [`&[data-color="positive"]`]: {
```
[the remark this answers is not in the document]
plagoa: I actually needs to be as it is in this case.

packages/styles/src/themes/pentahoPlus.ts
plagoa: We have no token for this `48px` value, I think.

packages/styles/src/themes/pentahoPlus.ts
MEsteves22: Agreed, we can now take advantage of the token added in [this PR](https://github.com/lumada-design/hv-uikit-react/pull/4327) for the theming 🙏

plagoa: I'll review this PR with the new tokens to check where we can use those.

zettca (changes requested):

plagoa:
@zettca @MEsteves22 updated the PR with comments. the colors have been updated to use the new tokens.

MEsteves22 (commented):
Really nice 🙏 

I just found the following visual change for the Pentaho+ theme:

Screenshot 2024-09-18 at 12 21 24

The multi button needs to be fixed: the dropdown menu button (right side) has a different color than the left side.

plagoa:
@MEsteves22 fixed the styles for the `DropdownButton` in the pentaho plus theme

MEsteves22 (approved):
🚀

zettca (approved):

PR #4336, opened 2024-09-16
chore: Chromatic modes enabled & stories for visual testing reviewed

MEsteves22:
- Custom theme decorator removed and replaced by `@storybook/addon-themes` to enable Chromatic modes:
   - ℹ️ I was not able to keep the following feature: changing the mode (dark/light) for the docs updates the mode of the stories. Personally, I don't think it's a blocker.
- Chromatic modes enabled: stories are now visually tested for DS3 dawn/wicked, DS5 dawn/wicked, Pentaho+ dawn/wicked 
- Stories reviewed:
   - Tested in all 6 modes: "big stories" that combine various components (favoured instead of templates since they contain way more components). 
   - Only tested in DS5 (default): very basic components that are reused throughout various components or "structural" components.   
   - Tested for all themes but only in Dawn modes: specific sample of a component already tested for all 6 modes or components created by using smaller peaces/components already tested for all 6 modes. 
   - Only in DS5 dawn/wicked: flow samples.
   - Only in Pentaho+ dawn/wicked: Pentaho package and "Canvas" template.

ℹ️ **Previously, we were taking 119 snapshots. We now have 120 snapshots and a lot more coverage**.

💡 **Next steps:**
- Enable more browsers

⚠️ **Let me know if you think some tests are unnecessary or if more samples could be combined.**


packages/core/src/TestComponents.stories.tsx
```diff
@@ -0,0 +1,458 @@
...
+import { Test as TabsTestStory } from "./Tabs/Tabs.stories";
+import { Test as TagTestStory } from "./Tag/stories/Tag.stories";
+import { Variants as TagsInputVariantsStory } from "./TagsInput/TagsInput.stories";
+import { Variants as TextAreaVariantsStory } from "./TextArea/TextArea.stories";
+import { Multiple as ToggleButtonMultipleStory } from "./ToggleButton/ToggleButton.stories";
+import { Test as TypographyTestStory } from "./Typography/Typography.stories";
+import { setupChromatic } from ".storybook/setupChromatic";
+
+const meta: Meta = {
+  title: "Components/Test",
```
zettca, line 62: Maybe we should bring back the `Tests/*` directory under `docs/`. These tests feel a bit unsorted 🤔 

What do you think?

packages/core/src/TestComponents.stories.tsx
```diff
@@ -0,0 +1,458 @@
...
+            context,
+          )}
+        </HvSimpleGrid>
+      </div>
+      {DropdownTestStory.render?.(DropdownTestStory.args as any, context)}
+    </div>
+  ),
+};
+
+export const Test2: StoryObj = {
```
zettca, line 150: maybe we could give these more descriptive names, so that it's clearer in which sample to add a new component/variant

zettca (approved):
Awesome stuff 💯

.storybook/main.ts
```diff
@@ -35,8 +35,8 @@ export default {
     },
     getAbsolutePath("@storybook/addon-a11y"),
     getAbsolutePath("@storybook/addon-links"),
+    getAbsolutePath("@storybook/addon-themes"),
```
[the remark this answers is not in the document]
MEsteves22: I think it should be possible to change the UI a bit, I'll take a look

packages/core/src/TestComponents.stories.tsx
MEsteves22: Agreed, I'll add some comments

packages/core/src/TestComponents.stories.tsx
MEsteves22: Agreed, I'll do that

.storybook/manager.ts
```diff
@@ -19,9 +19,13 @@ addons.setConfig({
...
           )
         ) {
-          // Hide "Test" stories in the production build
+          // Hide "Test" and "AutomationTest" stories in the production build
           // We need them in other cases for Chromatic and development
-          return !item.name.startsWith("Test");
+          return (
+            !item.name.startsWith("Test") &&
+            !item.name.startsWith("Automation") &&
+            !item.title.includes("Test")
```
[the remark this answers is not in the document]
MEsteves22: Wrong name 😂 it should be playwright test instead

packages/core/src/TestComponents.stories.tsx
MEsteves22: I'll add some comments for now

.storybook/main.ts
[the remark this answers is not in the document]
MEsteves22: The only way I had of updating this was through CSS override

PR #4347, opened 2024-09-19
fix(ToolbarTabs): inline edit content sanitised

MEsteves22:
- The content of the toolbar tabs inline editor is sanitised: more info in the Jira issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,6 @@
         "@emotion/react": "^11.11.1",
         "@emotion/styled": "^11.11.0",
         "@fortawesome/fontawesome-svg-core": "^6.6.0",
-        "@fortawesome/free-brands-svg-icons": "^6.6.0",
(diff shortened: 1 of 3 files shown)


plagoa (approved):

packages/pentaho/package.json
```diff
@@ -44,13 +44,15 @@
     "@hitachivantara/uikit-react-icons": "^5.12.0",
     "@hitachivantara/uikit-react-utils": "^0.2.5",
     "@mui/base": "^5.0.0-beta.40",
-    "react-resize-detector": "^8.1.0"
+    "react-resize-detector": "^8.1.0",
+    "sanitize-html": "^2.13.0"
```
zettca, line 48: this library is really bloated/overkill for what we need - can't we use something else?

MEsteves22: I'll look into it 👍

MEsteves22:
@zettca updated ✅

PR #4348, opened 2024-09-19
fix(Badge): add pentaho plus styles

plagoa:
- The `Badge` component is `Stable` and `Ready for Dev`. The only change was on the typography (`label` > `caption2`)

### DIFF
--- a/packages/core/src/Badge/Badge.styles.tsx
+++ b/packages/core/src/Badge/Badge.styles.tsx
@@ -2,7 +2,7 @@ import { createClasses } from "@hitachivantara/uikit-react-utils";
 import { theme } from "@hitachivantara/uikit-styles";
 
 const labelBaseStyle: React.CSSProperties = {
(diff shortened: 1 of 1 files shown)


packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -363,6 +363,18 @@ const pentahoPlus = makeTheme((theme) => ({
     base: "6px",
   },
   components: {
+    HvBadge: {
+      classes: {
+        showCount: {
+          ...theme.typography.caption2,
```
zettca, line 369: NEXT 5 is also `caption2`. Let's change the "baseline" instead?

plagoa: @zettca updated to change the baseline instead.

plagoa:
@zettca updated to change the baseline instead.

PR #4365, opened 2024-10-01
docs(Input): add native examples & test with more variants

zettca:
- add more `HvInput` native `type` examples
  - move `TimePicker`'s sample to `HvInput`
- add `Test` with more variants, and `Variants` with fewer


MEsteves22 (approved):
PR approved in Chromatic ✅ 
Not merging since the new "Test" sample needs to changes due to a11y issues 🙏

PR #4370, opened 2024-10-02
fix(Badge): update pentaho+ styles

plagoa:
- update the `borderRadius` of the Badge component both on the DS theme and P+ theme
- update the `padding` values for the P+ theme

### DIFF
--- a/packages/core/src/Badge/Badge.styles.tsx
+++ b/packages/core/src/Badge/Badge.styles.tsx
@@ -13,7 +13,7 @@ export const { staticClasses, useClasses } = createClasses("HvBadge", {
   badgeContainer: { width: 0 },
   badgePosition: {},
   badge: {
-    borderRadius: theme.space.xs,
+    borderRadius: theme.space.sm,
     backgroundColor: theme.colors.secondary,
     float: "left",
     minHeight: "8px",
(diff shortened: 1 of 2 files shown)


packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -363,6 +363,16 @@ const pentahoPlus = makeTheme((theme) => ({
     base: "6px",
   },
   components: {
+    HvBadge: {
+      classes: {
+        showCount: {
+          padding: "2px 5px",
```
zettca, line 369: I don't think this change leads to the expected visual changes.
We already have this `2px` vertical padding, through `lineHeight` (it's `16px` instead of `12px`)

packages/core/src/Badge/Badge.styles.tsx
```diff
@@ -13,7 +13,7 @@ export const { staticClasses, useClasses } = createClasses("HvBadge", {
   badgeContainer: { width: 0 },
   badgePosition: {},
   badge: {
-    borderRadius: theme.space.xs,
+    borderRadius: theme.space.sm,
```
zettca, line 16: In one hand, shouldn't we be using one of our `radii` units (eg. `radii.full`)?
In the other hand, it won't matter either way (`sm`, `xs`, `full`, ...) because the badge's height is always `16px`

plagoa:
Closing this as the changes are not needed.

PR #4372, opened 2024-10-03
docs: multiple fixes and improve light/dark mode support

francisco-guilherme:


packages/core/src/Card/Card.tsx
```diff
@@ -46,7 +46,7 @@ export const HvCard = (props: HvCardProps) => {
     selectable = false,
     selected = false,
     statusColor = "sema0",
-    bgcolor,
+    bgcolor = "atmo1",
```
zettca, line 49: IIRC, we discussed making this change in the past, but considered it a breaking change as it affects many `HvCard` that don't want to have a background

PR #4379, opened 2024-10-08
fix(InlineEditor): value not updated on escape when controlled

MEsteves22:
When controlling the inline editor, the "Escape" functionality was not working: pressing Escape was not returning the cached value to the user in order for them to update the inline editor accordingly

### DIFF
--- a/packages/core/src/InlineEditor/InlineEditor.tsx
+++ b/packages/core/src/InlineEditor/InlineEditor.tsx
@@ -46,6 +46,13 @@ export type HvInlineEditorProps<C extends React.ElementType = typeof HvInput> =
       ) => void;
       /** Called when the input value changes. */
       onChange?: (event: React.SyntheticEvent, value: string) => void;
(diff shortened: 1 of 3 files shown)


packages/core/src/InlineEditor/InlineEditor.tsx
```diff
@@ -46,6 +46,13 @@ export type HvInlineEditorProps<C extends React.ElementType = typeof HvInput> =
       ) => void;
       /** Called when the input value changes. */
       onChange?: (event: React.SyntheticEvent, value: string) => void;
+      /** Called when there's a keydown event on the input. */
+      onKeyDown?: (
+        event:
+          | React.KeyboardEvent<HTMLTextAreaElement | HTMLInputElement>
+          | React.MouseEvent,
```
zettca, line 53: This `React.MouseEvent` on an `onKeyDown` is really strange... but it's coming from `HvInput` 🤷

zettca (approved):

PR #4383, opened 2024-10-11
feat(Loading): update to PD 0.2 specs

plagoa:
- added the color override for the color token `brand` in the P+ theme
- I needed to change the animation a bit: because we were using `scale` to change the height of each bar, now that it has a rounded border the change in scale changes the rounded borders as the element grows and shrinks. I opted to animate the height instead.

### DIFF
--- a/packages/core/src/Loading/Loading.styles.tsx
+++ b/packages/core/src/Loading/Loading.styles.tsx
@@ -11,6 +11,7 @@ export const { staticClasses, useClasses } = createClasses("HvLoading", {
   },
   barContainer: {
     display: "flex",
+    alignItems: "center",
     justifyContent: "space-around",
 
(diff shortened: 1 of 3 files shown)


packages/core/src/Loading/Loading.styles.tsx
```diff
@@ -19,7 +20,7 @@ export const { staticClasses, useClasses } = createClasses("HvLoading", {
     },
 
     ":has($small)": {
-      "--scaleY": "0.223",
+      "--scaleY": "60%",
```
zettca, line 23: This should `22%`, to match what we had. Also, let's call the var `height` or `h` - it's not longer a scale
```suggestion
      "--height": "22%",
```

zettca (approved):

PR #4385, opened 2024-10-14
feat(Footer): update to PD 0.2 specs

plagoa:


packages/core/src/Footer/Footer.tsx
```diff
@@ -29,8 +29,8 @@ export interface HvFooterProps extends HvBaseProps {
  */
 export const HvFooter = (props: HvFooterProps) => {
   const {
-    name = "Hitachi Vantara",
-    copyright = `© Hitachi Vantara Corporation ${new Date().getFullYear()}. All Rights Reserved.`,
+    name = "Pentaho+",
```
zettca, line 32: This will impact current DS5 users. Let's use the theme's defaultProps to change this. This way, each theme has their own "Brand name" automatically 😃

packages/core/src/Footer/Footer.tsx
```diff
@@ -29,8 +29,8 @@ export interface HvFooterProps extends HvBaseProps {
...
+    copyright = `© Hitachi Vantara LLC ${new Date().getFullYear()}. All Rights Reserved.`,
```
zettca, line 33: if we're changing to LLC, shouldn't we change it elsewhere too?

packages/core/src/Footer/Footer.stories.tsx
```diff
@@ -13,8 +13,8 @@ export default meta;
 
 export const Main: StoryObj<HvFooterProps> = {
   args: {
-    name: "Hitachi Vantara",
-    copyright: "© Hitachi Vantara Corporation 2022",
+    name: "Pentaho+",
+    copyright: "© Hitachi Vantara Corporation 2024",
```
zettca, line 17: I think we should remove the `args` instead. This way, each theme would get their default `name`/`copyright` values (which we can more easily test against), and Storybook users can still override the props

packages/core/src/Footer/Footer.tsx
plagoa: oh yeah, I forgot about that. I'll update.

packages/core/src/Footer/Footer.tsx
plagoa: I'll update.

plagoa:
@zettca updated the PR

zettca (approved):

PR #4389, opened 2024-10-16
feat(Switch): align with PD 0.2 specs

plagoa:


packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -66,6 +66,9 @@ const pentahoPlus = makeTheme((theme) => ({
           primaryStrong: blue[800],
           primarySubtle: blue[200],
           primaryDimmed: blue[100],
+          secondary: slate[700],
+          secondarySubtle: slate[500],
+          secondaryDimmed: slate[300],
```
zettca, line 71: these should be `text`/`textSubtle`/`textDimmed`, which already exist

zettca (commented):
The styles don't look ok. The "unselected" "thumb" has really low contrast, especially in dark theme - and the border seems off 

image
image

plagoa:
@zettca the PR was updated. This is only for the updated P+ 0.2 specs, it doesn't include the new feature for allowing other colors.

zettca (approved):

PR #4394, opened 2024-10-22
docs: add slider and radio control renderers

plagoa:
- add the possibility of overriding the custom control for enums (`Select`) with a `Slider` or a `RadioGroup`
- also add the `Banner` page

### DIFF
--- a/apps/docs/src/pages/components/Playground.tsx
+++ b/apps/docs/src/pages/components/Playground.tsx
@@ -9,12 +9,16 @@ import { themes } from "prism-react-renderer";
 import {
   HvCheckBox,
   HvOption,
+  HvRadio,
+  HvRadioGroup,
   HvSelect,
   HvSlider,
 } from "@hitachivantara/uikit-react-core";
 
+type ControlType = "radio" | "slider";
+
 type Control = {
-  type?: string;
+  type?: ControlType;
   defaultValue?: string;
   values?: string[];
 };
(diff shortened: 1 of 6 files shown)


apps/docs/src/pages/components/banner.mdx
```diff
@@ -0,0 +1,125 @@
...
+  actions={[
+    { id: "action_1", label: "Action 1", disabled: false },
+    { id: "action_2", label: "Action 2", disabled: false },
+  ]}
+  onAction={(event, action) => console.log("Clicked", action)}
+  style={{ position: "relative", top: 0 }}
+/>
+```
+
+### With custom icons
```
zettca, line 57: Can we get rid of the "With " prefixes? They don't add any value (any sample could be prefixed with "With " and the meaning is the same)

apps/docs/src/pages/components/Playground.tsx
```diff
@@ -149,13 +227,13 @@ export const Playground = ({
...
-      <div className="flex justify-between p-2 mt-1 border border-[var(--uikit-colors-atmo4)] rounded-t-round">
-        <div className="w-[70%] flex justify-center items-center">
+      <div className="flex justify-between mt-1 border border-[var(--uikit-colors-atmo4)] rounded-t-round">
+        <div className="w-[70%] flex justify-center items-center p-[var(--uikit-space-sm)]">
           <Component {...propsState} {...componentProps}>
             {children?.props.children}
           </Component>
         </div>
-        <div className="w-[30%] flex flex-col gap-[var(--uikit-space-xs)] justify-center items-center border-l border-[var(--uikit-colors-atmo3)] pl-[var(--uikit-space-sm)]">
+        <div className="w-[30%] flex flex-col gap-[var(--uikit-space-xs)] justify-center items-center border-l border-[var(--uikit-colors-atmo3)] p-[var(--uikit-space-sm)] pl-[var(--uikit-space-xs)]">
```
zettca, line 236: Don't the breakpoints work? This is really strange
```suggestion
        <div className="w-[30%] flex flex-col gap-xs justify-center items-center border-l border-[var(--uikit-colors-atmo3)] p-sm pl-xs">
```

zettca (approved):

apps/docs/src/pages/components/banner.mdx
plagoa: Yes, I'll remove those.

PR #4402, opened 2024-10-29
chore: add animation to tab change in floating mode

plagoa:
updated the floating variant to use the underlying tab selected indicator animation to keep consistency with remaining tabs


zettca (approved):
very sleek, very demure 💅

PR #4412, opened 2024-11-04
docs: style improvements

francisco-guilherme:


apps/docs/src/pages/components/avatar.mdx
```diff
@@ -1,7 +1,7 @@
 import { avatarClasses, HvAvatar } from "@hitachivantara/uikit-react-core";
 
-import { Description } from "@docs/components/Description";
-import { Page } from "@docs/components/Page";
+import { Description } from "@docs/components/component/Description";
+import { Page } from "@docs/components/component/Page";
```
zettca, line 4: If these are only for components, maybe we could simplify and just leave on the components page (like `Playground` is) instead of adding more nesting

```suggestion
import { Description } from "./Description";
import { Page } from "./Page";
```

zettca (approved):
some small notes that can be reviewed later
feel free to merge

apps/docs/src/components/Footer.tsx
```diff
@@ -0,0 +1,82 @@
+import { UIKitLogo } from "@docs/components/logo/uikit";
+
+const footerLinks = [
+  {
+    title: "Project",
+    links: [
+      { label: "Get Started", href: "/documentation/get-started" },
+      { label: "Contribute", href: "/documentation/contribute/" },
+      { label: "Changelog", href: "/documentation/changelog/" },
```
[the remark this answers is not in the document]
francisco-guilherme: We'll fix them later

apps/docs/src/pages/components/avatar.mdx
francisco-guilherme: We can take this offline. Personally i prefer to keep this separation.

PR #4419, opened 2024-11-08
docs: add `Card` documentation

plagoa:


apps/docs/src/pages/components/card.mdx
```diff
@@ -22,4 +22,160 @@ export const getStaticProps = async ({ params }) => {
...
+      title={
+        <HvTypography variant="title3">Start using UI Kit now</HvTypography>
+      }
+      subheader={
+        <HvTypography>
+          Get started with the Hitachi Vantara UI Kit and build your next
+          project.
+        </HvTypography>
+      }
+    ></HvCardHeader>
```
zettca, line 99:
```suggestion
    />
```

apps/docs/src/pages/components/card.mdx
```diff
@@ -22,4 +22,160 @@ export const getStaticProps = async ({ params }) => {
...
+    <HvTypography>
+      Set up a job to run automatically at your preferred time and frequency.
+    </HvTypography>
+  </HvCardContent>
+  <HvActionBar>
+    <div className="flex justify-between w-full">
+      <HvButton variant="secondaryGhost">Configure Job</HvButton>
+      <HvButton variant="secondarySubtle">Set Schedule</HvButton>
+    </div>
+  </HvActionBar>
```
zettca, line 57: Can we instead use a `flex-1` div to separate the buttons? `HvActionBar` already is a flex container, and this way it's easier add/remove buttons & keep the expected behaviour. `justify-between` only works for 2 buttons
```suggestion
  <HvActionBar>
    <HvButton variant="secondaryGhost">Configure Job</HvButton>
    <div aria-hidden className="flex-1" />
    <HvButton variant="secondarySubtle">Set Schedule</HvButton>
  </HvActionBar>
```

zettca (approved):

PR #4430, opened 2024-11-13
docs(app): upgrade to react-router@7

zettca:
testing upgrading to `react-router@7` + vite plugin (there are a few issues)


packages/core/src/BaseDropdown/BaseDropdown.tsx
```diff
@@ -8,8 +8,8 @@ import {
   useState,
 } from "react";
 import { PopperProps, usePopper } from "react-popper";
-import type { ClickAwayListenerProps } from "@mui/material/ClickAwayListener";
-import { useForkRef } from "@mui/material/utils";
+import type { ClickAwayListenerProps } from "@mui/base";
+import { unstable_useForkRef as useForkRef } from "@mui/utils";
```
andrew-kayne, line 12: Is use of the `unstable_useForRef` intentional? It seems that `useForkRef` is used in number of files, but in some cases you use the unstable version and in others, you don't. I don't know what the difference is but it seems odd to use unstable and stable interchangeably.

zettca: I'm not sure either, this was long ago. I could've just imported from `@mui/material` yes. They're the [same code](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/utils/useForkRef.js) btw.

That specific commit was never meant to be merged anyways. The vite plugin was not working with MUI's ([non-standard](https://unpkg.com/browse/@mui/material@6.3.1/Accordion/package.json)) subpaths "exports".

PR #4432, opened 2024-11-14
docs: add `Accordion` documentation

plagoa:


apps/docs/src/pages/components/accordion.mdx
```diff
@@ -0,0 +1,127 @@
...
+```jsx live
+<div className="w-full flex flex-col">
+  <HvTypography className="pb-4" variant="title3">
+    UI Kit
+  </HvTypography>
+
+  <HvAccordion
+    label="Components"
+    labelVariant="title4"
+    style={{ width: "50%" }}
```
zettca, line 101: instead of having this this on each Accordion, we move it to the the parent div (eg. `w-1/2`)
```suggestion
```

zettca (approved):

PR #4437, opened 2024-11-18
fix(SidePanel): prevent side panel content from being selected during resize

plagoa:


packages/pentaho/src/Canvas/SidePanel/useResizable.tsx
```diff
@@ -64,24 +64,33 @@ export const useResizable = (
...
+      panelRef.current.parentElement?.addEventListener(
+        "mousemove",
+        mouseMove as unknown as EventListener,
+      );
+      panelRef.current.parentElement?.addEventListener(
+        "mouseup",
+        handleMouseUp,
+      );
+      setIsDragging(true);
+    }
```
zettca, line 93: Can we please use [guard clauses](https://deviq.com/design-patterns/guard-clause) so the code is easier to read (also because the lines are long)?

I typically wouldn't complain about this, but the types are also wrong! We should be using the global native `MouseEvent`, not `React.MouseEvent` (above too). With this change, there's no need for any `as` cast

```suggestion
    if (!panelRef.current) return;

    panelRef.current.style.userSelect = "none";
    panelRef.current.parentElement?.addEventListener("mousemove", mouseMove);
    panelRef.current.parentElement?.addEventListener("mouseup", handleMouseUp);
    setIsDragging(true);
```

plagoa: sure, updated the PR.

zettca (approved):

francisco-guilherme (approved):

PR #4440, opened 2024-11-19
docs: style improvements

francisco-guilherme:
**Docs**: 
- Bumped dependencies.

**Code Editor:**
- Synced theme with light/dark modes.
- Added toolbar actions: show/hide, reset, copy code.

**Fix:** 
- Resolved ToC visibility issue on active tabs.

**Refactor:** 
- Applied code improvements for better maintainability.


apps/docs/package.json
```diff
@@ -27,12 +27,12 @@
...
-    "@unocss/postcss": "^0.58.5",
+    "@types/node": "^22.9.0",
+    "@types/react": "^18.3.12",
+    "@unocss/postcss": "^0.64.1",
     "rehype-mdx-code-props": "^3.0.1",
     "string-replace-loader": "^3.1.0",
-    "typescript": "^5.4.2",
-    "unocss": "^0.58.5"
+    "typescript": "^5.6.3",
+    "unocss": "^0.64.1"
```
zettca, line 36: Do we really need to bump these? We shouldn't bump `unocss` or `typescript` unless we really need to, to avoid the issues that'll arise from having multiple versions of them in the repo (`unocss` if a peer-dependency of one of our packages and is on `0.x`, and we need a fixed `typescript@5.4.2` to be used)

apps/docs/package.json
```diff
@@ -27,12 +27,12 @@
   },
   "devDependencies": {
     "@hitachivantara/uikit-uno-preset": "*",
-    "@types/node": "^20.11.6",
-    "@types/react": "^18.2.43",
-    "@unocss/postcss": "^0.58.5",
+    "@types/node": "^22.9.0",
```
zettca, line 30: Why?
We're on `node@20` on the root, which is already two majors ahead of the versions our CI run (and that we officially support)

apps/docs/package.json
francisco-guilherme: you're right. i just bulk bumped, my bad.

apps/docs/package.json
francisco-guilherme: you're right. i just bulk bumped, my bad.

zettca (approved):

PR #4446, opened 2024-11-21
docs: add `OverflowTooltip` documentation

plagoa:
- This also adds the possibility of passing a decorator to the playground to be used around the component

### DIFF
--- a/apps/docs/src/components/code/Playground.tsx
+++ b/apps/docs/src/components/code/Playground.tsx
@@ -11,6 +11,7 @@ type PlaygroundProps = {
   componentProps?: Record<string, unknown>;
   controls: Record<string, Control>;
   children?: React.ReactNode;
+  decorator?: (component: JSX.Element) => JSX.Element;
 };
 
 const generateCode = (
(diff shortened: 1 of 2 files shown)


apps/docs/src/pages/components/overflowTooltip.mdx
```diff
@@ -0,0 +1,59 @@
...
+  }}
+  decorator={(component) => (
+    <div
+      style={{
+        display: "flex",
+        alignItems: "center",
+        margin: "0 auto",
+        height: 200,
+        maxWidth: 200,
+      }}
```
zettca, line 52: can we shrink the container to `160px` or lower? Otherwise `paragraphOverflow` doesn't work
also, the parent element already takes care of the centering
```suggestion
      className="max-w-160px"
```

apps/docs/src/pages/components/_meta.ts
```diff
@@ -13,6 +13,7 @@ export default {
   input: "Input",
   loading: "Loading",
   loadingContainer: "LoadingContainer",
+  overflowTooltip: "OverflowTooltip",
```
zettca, line 16: no longer necessary (conflict). also, we should call the file `overflow-tooltip.mdx` as [URLs should be kebab-cased](https://redocly.com/docs/cli/rules/paths-kebab-case)

plagoa: @zettca made the changes you mentioned.

PR #4450, opened 2024-11-25
docs(FormElement): add docs & refactor component structure

zettca:
- hoist `HvFormElement` to root, aligning with other components (this is needed for the current `nextra` doc generation)
- keep `HvLabel`, `HvWarningText`, etc. as `HvFormElement`'s sub-components, keeping alignment with other components
- merge `HvFormElement`'s contexts (and remove/refactor redundant `.Provider` `.Consumer` components)

> [!WARNING]
> please rebase, don't squash


apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -123,79 +79,10 @@ export default function Demo() {
 }
 ```
 
-### External error message
```
zettca, line 126: this is a awkward example to add for each form component. it should be added in a complete form example/template instead

plagoa: Maybe we should add a callout on each form component documentation page to direct people to the `FormElement` page (not just on the related components section)

francisco-guilherme (approved):

PR #4453, opened 2024-11-26
fix: relative package imports

zettca:
relative package imports (ie. `package/utils/src`) is leading to a broken build

image


MEsteves22 (approved):
🚀

PR #4456, opened 2024-11-30
fix: Sidebar open state out of sync

plagoa:


packages/pentaho/src/Canvas/SidePanel/SidePanel.tsx
```diff
@@ -108,6 +108,10 @@ export const HvCanvasSidePanel = forwardRef<
     tabs?.[0]?.id ?? "none",
   );
 
+  useEffect(() => {
+    handleSidePanelOpen?.(open);
```
zettca, line 112: can we remove the eslint-disable below? it's exactly the same use-case

plagoa: Updated the PR.

zettca (approved):

PR #4457, opened 2024-12-02
docs: add `Skeleton` documentation

plagoa:


apps/docs/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,117 @@
...
+### Custom animation
+
+You can use custom CSS animations with the skeleton component by setting the `animation` prop to `undefined`
+and defining your own in your styles.
+
+```tsx live
+export default function Demo() {
+    import { useMemo } from "react";
+
+    const customAnimationKeyframes = useMemo(
```
zettca, line 85: these should be outside the component

apps/docs/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,117 @@
...
+      () => `
+      @keyframes customAnimation {
+        50% { box-shadow: inset 0px 0px 60px 0px rgba(0,0,0,0.25); }
```
zettca, line 88: this is not visible in dark theme; can we use a color that works in both?
```suggestion
        50% { box-shadow: inset 0px 0px 60px 0px #66669940; }
```

apps/docs/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,117 @@
...
+      <style>{customAnimationKeyframes}</style>
+      <HvSkeleton
+        width={300}
+        height={100}
+        animation={undefined}
+        variant="square"
+        style={{
+          animation: "customAnimation 1s infinite"
+        }}
+      />
```
zettca, line 106: Why not use `css()` for the custom animation, like we're doing elsewhere?
If we're avoiding Emotion, then we could leverage UnoCSS's built-in animations (eg. `animate-flash`); what do you think?

Also, `animation={undefined}` does nothing

```suggestion
      <HvSkeleton
        width={300}
        height={100}
        variant="square"
        className="animate-flash animate-count-infinite"
      />
```

plagoa:
@zettca updated PR.

apps/docs/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,96 @@
...
+and defining your own in your styles.
+
+```tsx live
+<HvSkeleton
+  width={300}
+  height={100}
+  animation={undefined}
+  variant="square"
+  className="animate-flash animate-count-infinite"
+/>
```
zettca, line 88: We'll need the `!` or the animation won't work.
We should probably review the `HvProvider`'s and UnoCSS's injection order

```suggestion
You can use custom CSS animations with the skeleton component by defining your own in your styles.

```tsx live
<HvSkeleton
  width={300}
  height={100}
  variant="square"
  className="!animate-flash !animate-count-infinite"
/>
```

plagoa: Updated.

zettca: `animation={undefined}` and "by setting the `animation` prop to `undefined`" is still there

plagoa: Updated.

PR #4465, opened 2024-12-04
fix(ColorPicker): hue slider goes out of bounds

plagoa:


packages/core/src/ColorPicker/Picker/Picker.styles.tsx
```diff
@@ -54,7 +54,7 @@ export const { staticClasses, useClasses } = createClasses(name, {
     marginLeft: "0px",
     border: "2px solid #fff",
     borderRadius: theme.radii.circle,
-    transform: "translate(0, -2px)",
+    transform: "translate(-6px, -2px)",
```
zettca, line 57: the visual changes seem unexpected, it goes out of bounds when vertical & on the start of the slider

plagoa: Yeah, I fixed it for the vertical case too. Still, some visual changes are expected.

zettca (approved):

PR #4466, opened 2024-12-04
fix(ColorPicker): unable to set R/G/B components as 0

plagoa:
value of 0 was being interpreted as `false`

### DIFF
--- a/packages/core/src/ColorPicker/Fields/Fields.tsx
+++ b/packages/core/src/ColorPicker/Fields/Fields.tsx
@@ -88,12 +88,16 @@ export const Fields = (props: FieldsProps) => {
         },
         event,
       );
-    } else if (data.r || data.g || data.b) {
+    } else if (
+      data.r !== undefined ||
+      data.g !== undefined ||
+      data.b !== undefined
+    ) {
       onChange(
         {
-          r: data.r || rgb?.r,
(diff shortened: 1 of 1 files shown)


packages/core/src/ColorPicker/Fields/Fields.tsx
```diff
@@ -88,12 +88,16 @@ export const Fields = (props: FieldsProps) => {
...
+      data.b !== undefined
+    ) {
       onChange(
         {
-          r: data.r || rgb?.r,
-          g: data.g || rgb?.g,
-          b: data.b || rgb?.b,
+          r: data.r !== undefined ? data.r : rgb?.r,
+          g: data.g !== undefined ? data.g : rgb?.g,
+          b: data.b !== undefined ? data.b : rgb?.b,
```
zettca, line 100: we have syntax for this case
```suggestion
          r: data.r ?? rgb?.r,
          g: data.g ?? rgb?.g,
          b: data.b ?? rgb?.b,
```

zettca (approved):

packages/core/src/ColorPicker/Fields/Fields.tsx
plagoa: Of course. Updated.

PR #4469, opened 2024-12-04
docs(Playground): fix children rendering & component name

zettca:
- fixes `children` rendering in `Playground`
  - we should use actual `children`, not `children={}` (requires fragments)
- this also seems to fix the `"No Display Name"` in children components

### DIFF
--- a/apps/docs/src/pages/components/section.mdx
+++ b/apps/docs/src/pages/components/section.mdx
@@ -24,7 +24,7 @@ export const getStaticProps = async ({ params }) => {
   controls={{
     title: {
       type: "text",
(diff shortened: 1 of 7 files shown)


apps/docs/src/pages/components/section.mdx
```diff
@@ -33,12 +33,9 @@ export const getStaticProps = async ({ params }) => {
       defaultValue: false,
     },
   }}
-  children={
-    <div>
-      <p>My section example content</p>
```
zettca, line 38: we should avoid `p` if we can, because nextra is currently overriding its styles

plagoa (approved):

PR #4482, opened 2024-12-10
docs: add Slider

zettca:


plagoa:
Maybe we should add the `divisionQuantity` prop to the Playground also, it's useful to play around with the `markStep` prop. Maybe the `markDigits` also?

plagoa:
Weird behavior, possibly a bug: when I try to backspace the `markStep` property until it's empty the page crashes. I think the same's happening in our current docs also. When I do it there the page sample crashes and throws an error "Error: Invalid array length".

plagoa (approved):

PR #4488, opened 2024-12-16
docs: add `EmptyState` documentation

plagoa:


apps/docs/src/pages/components/empty-state.mdx
```diff
@@ -0,0 +1,58 @@
...
+You can specify any React component as the action link. This can be a button, a link, or any other component that you want to use to guide the user to the next step.
+
+```tsx live
+<HvEmptyState
+  title="No visualization data available"
+  icon={<BarChart />}
+  action={
+    <div className="flex flex-col gap-1 items-center">
+      <HvButton variant="secondarySubtle">Upload data</HvButton>
+      <HvLink route="/">Back to homepage</HvLink>
```
zettca, line 52: actually, can we improve this now? using `href="#"` is also better, as clicking the link won't navigate away from from the docs
```suggestion
      <HvTypography link component="a" href="#">Back to homepage</HvTypography>
```

PR #4492, opened 2025-01-02
docs: add `Pagination` documentation

plagoa:


apps/docs/src/pages/components/pagination.mdx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <div>
+      <div className="flex flex-wrap">
+        {data.slice(pageSize * page, pageSize * (page + 1)).map((i) => (
+          <span
+            className="w-[100px] px-3 py-2 m-3 text-center rounded bg-[var(--uikit-colors-atmo1)] "
+            key={i}
+          >{`Item ${i + 1}`}</span>
+        ))}
+      </div>
```
zettca, line 70: The root needs a `w-full` so that the Pagination doesn't collapse.
Also, can we use an `HvPanel` for the items, so that we don't need so many styles? We should probably avoid things like `bg-[var(--uikit-colors-atmo1)]` if we can

```suggestion
    <div className="w-full">
      <div className="flex flex-wrap gap-sm rounded">
        {data.slice(pageSize * page, pageSize * (page + 1)).map((i) => (
          <HvPanel
            className="w-[100px] text-center"
            key={i}
          >{`Item ${i + 1}`}</HvPanel>
        ))}
      </div>
```

plagoa: Updated

zettca (approved):

PR #4515, opened 2025-01-16
docs: add `TreeView` documentation

plagoa:
Also includes a small change to the `Live` component to allow the importing of UI Kit hooks.

### DIFF
--- a/apps/docs/src/pages/components/tree-view.mdx
+++ b/apps/docs/src/pages/components/tree-view.mdx
@@ -1,0 +1,431 @@
+import { Callout } from "nextra/components";
+import {
+  HvPanel,
+  HvTreeItem,
+  HvTreeView,
+  treeView2Classes,
+} from "@hitachivantara/uikit-react-core";
+
+import Playground from "@docs/components/code/Playground";
+import { Description } from "@docs/components/page/Description";
+import { Page } from "@docs/components/page/Page";
+import { getComponentData } from "@docs/utils/component";
+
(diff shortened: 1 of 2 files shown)


apps/docs/src/pages/components/tree-view.mdx
```diff
@@ -0,0 +1,433 @@
...
+  Component={HvTreeView}
+  componentName="HvTreeView"
+  controls={{
+    multiSelect: { defaultValue: false },
+    disableSelection: { defaultValue: false },
+}}
+decorator={(component) => (
+<HvPanel style={{ width: 300 }}>{component}</HvPanel>
+)}
+
```
zettca, line 35: there must be errors somewhere in the samples, because the code isn't formatted correctly

apps/docs/src/pages/components/tree-view.mdx
```diff
@@ -0,0 +1,433 @@
...
+            <HvTreeItem nodeId="5" label="index.js" />
+          </HvTreeItem>
+        </HvTreeView>
+      </HvPanel>
+    </div>
+  );
+}
+```
+
+### Loading data
```
zettca, line 205: Note for later: I think this would be a good sample to move to `examples/` (and "Custom tree view" too)

apps/docs/src/pages/components/tree-view.mdx
plagoa: I couldn't find any formatted the code properly.

PR #4519, opened 2025-01-17
docs: add `SnackbarProvider` documentation

plagoa:


apps/docs/src/pages/components/snackbar-provider.mdx
```diff
@@ -0,0 +1,186 @@
...
+    >
+      Action
+    </HvButton>
+  );
+};
+```
+
+### Label
+
+The snackbar label acceps a `React.ReactNode` which means you can pass any component as the label. This allows you to use the `HvOverflowTooltip` component to display a long message with ellipsis and a tooltip.
```
zettca, line 142: I think we should reinforce/remind the user that snackbars shouldn't be more than 3 lines long. `HvOverflowTooltip` is a good workaround, but we could also suggest reducing/trimming the message size or using `HvBanner` instead

plagoa: I'll update the message.

plagoa:
@zettca updated the PR after your comments.

zettca (approved):

PR #4521, opened 2025-01-20
docs: add `Stack` documentation

plagoa:


apps/docs/src/pages/components/stack.mdx
```diff
@@ -0,0 +1,99 @@
...
+  {Array.from({ length: 4 }).map((_, index) => (
+    <div
+      key={index}
+      style={{
+        display: "grid",
+        placeContent: "center",
+        width: 50,
+        height: 50,
+        backgroundColor: theme.colors.primary_20,
+      }}
```
zettca, line 66: the `backgroundColor` is will be rendered as `'var(--uikit-colors-primary_20)'`

image

we might as well just use Uno instead

Btw, why `primary_20`? It's used for actions (eg. hovers). Shouldn't we be using an `atmo*`?

apps/docs/src/pages/components/stack.mdx
```diff
@@ -0,0 +1,99 @@
...
+    spacing: {
+      defaultValue: "sm",
+    },
+    divider: {
+      defaultValue: false,
+    },
+  }}
+>
+
+    {Array.from({ length: 4 }).map((_, index) => (
```
zettca, line 33: not a big deal, but `Array(n).keys()` is shorter, faster, and you get the index as values 😄 

https://jsperf.app/goqema/1/preview

apps/docs/src/pages/components/stack.mdx
```diff
@@ -0,0 +1,99 @@
...
+      key={index}
+      style={{
+        display: "grid",
+        placeContent: "center",
+        width: 50,
+        height: 50,
+        backgroundColor: theme.colors.primary_20,
+        borderRadius: theme.radii.circle,
+      }}
+      // className="flex items-center justify-center w-50px h-50px bg-primary"
```
zettca, line 91: are we using this? btw, `size-` is a nice helper
```suggestion
      // className="grid place-content-center size-50px bg-primary"
```

zettca (approved):

apps/docs/src/pages/components/stack.mdx
zettca: we could just use `HvPanel` instead

plagoa:
@zettca updated the PR

PR #4527, opened 2025-01-22
docs: add `BreadCrumb` documentation

plagoa:


apps/docs/src/pages/components/breadcrumb.mdx
```diff
@@ -0,0 +1,99 @@
...
+/>
+```
+
+### Custom path component
+
+You can use the `component` prop to pass a custom component to be renderer in each part of the breadcrumb.
+
+```tsx live
+export default function Demo() {
+  const CustomNavLink = ({ children, to, ariaLabel, ...others }: any) => (
```
zettca, line 62: this should be outside the component

apps/docs/src/pages/components/breadcrumb.mdx
```diff
@@ -0,0 +1,99 @@
...
+  Component={HvBreadCrumb}
+  componentName="HvBreadCrumb"
+  controls={{
+    maxVisible: {
+      defaultValue: 4,
+    },
+  }}
+  componentProps={{
+    listRoute: [
+      { label: "Label 1", path: "route1" },
```
zettca, line 32: can't we use the hash here? this way the user wouldn't be redirected await from the docs
```suggestion
      { label: "Label 1", path: "#route1" },
```

plagoa:
@zettca updated the PR with comments

zettca (approved):

PR #4528, opened 2025-01-22
docs: add examples page and improve code blocks

francisco-guilherme:


apps/docs/src/components/Header.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Description } from "./usage/Description";
+import { Props } from "./usage/Props";
+
+/**
+ * The `Header` component manages a tab-based layout
+ * and dynamically displays specific sections based on the selected tab.
+ */
+export const Header = () => {
+  // State to track the active tab (default is the first tab).
+  const [tab, setTab] = useState(0);
```
zettca, line 14: I don't see the value of these types of comments. Why do we need them? Comments are just like code, they have to be reviewed and maintained. Code should be self-commenting.

apps/docs/src/components/Pre.tsx
```diff
@@ -1,7 +1,7 @@
 import { isValidElement } from "react";
 import { Pre as NextraPre } from "nextra/components";
 
-import { Live } from "./Live";
+import { CodeBlock } from "./code/CodeBlock";
```
zettca, line 4: do we really need to rename this? we're losing the git history, and it's harder to understand what changed

zettca (commented):
I'm not a fan on hiding the component docs' code on a separate tab. IMO the source code is the primary value these component samples add, and it kind of defeats the purpose of `react-live`

apps/docs/src/components/home/Home.tsx
```diff
@@ -11,10 +11,22 @@ export const Home = () => (
...
   >
     {/* Background grid */}
-    <div className="absolute h-full w-full bg-[radial-gradient(var(--uikit-colors-secondary\\_80),transparent_1px)] [background-size:20px_20px] [mask-image:radial-gradient(ellipse_50%_50%_at_50%_50%,#000_70%,transparent_100%)]" />
+    <div
+      className={clsx(
+        "absolute h-full w-full",
+        "bg-[radial-gradient(var(--uikit-colors-secondary\\_80),transparent_1px)]",
+        "[background-size:20px_20px]",
+        "[mask-image:radial-gradient(ellipse_50%_50%_at_50%_50%,#000_70%,transparent_100%)]",
+      )}
```
zettca, line 20: I'm commenting this because I see the inclusion of `clsx` as a "cope" to deal with the complexity of this class.

We could instead just use style, which is easier to understand (and don't scare away new utility class users 😅).

```tsx
className="absolute h-full w-full"
style={{
  background: `radial-gradient(${theme.colors.secondary_80},transparent 1px)`,
  backgroundSize: "20px 20px",
  maskImage: `radial-gradient(ellipse 50% 50% at 50% 50%, #000 70%, transparent 100%)`,
}}
```

francisco-guilherme: removed from the scope of this PR

apps/docs/src/pages/components/tree-view.mdx
```diff
@@ -206,6 +203,7 @@ Sometimes the full tree data is unknown or paginated. This sample showcases how
 
 ```tsx live
 import { forwardRef, useState } from "react";
+import { useHvTreeItem } from "@hitachivantara/uikit-react-core";
```
zettca, line 206: Why is this necessary now? It wasn't it case previously. Can we keep the previous behaviour?

Also, there are places where this (I believe it's the same problem) isn't being applied/resolved:

image

apps/docs/src/components/code/ExpandableLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+        onToggle={() => setIsExpanded((prev) => !prev)}
+        isExpanded={isExpanded}
+        code={editorCode}
+        onReset={resetCode}
+      />
+
+      {/* Preview Section */}
+      <div
+        className={[
+          "p-4 pt-5 [&_tr]:table-row gap-2 bg-[var(--uikit-colors-atmo1)] overflow-hidden",
```
zettca, line 45: We keep should keep this as `atmo2` (`bgPage`). `atmo1` won't work for "surface" components (eg. `HvPanel`, `HvCard`, `HvSection`, etc.)

image

PR #4530, opened 2025-01-23
fix: headless mode in playwright

plagoa:


playwright.config.ts
```diff
@@ -50,7 +51,6 @@ export default defineConfig({
       name: "chrome",
       use: {
         ...devices["Desktop Chrome"],
-        channel: "chrome",
```
zettca, line 53: it seems like the default is `chromium`: https://playwright.dev/docs/browsers#chromium

can we add it explicitly, so we can better understand what's going on? and remove `chromium` section, because `chrome` and `chromium` are exactly the same

PR #4532, opened 2025-01-24
docs: add `AppSwitcher` documentation

plagoa:


apps/docs/src/pages/components/app-switcher.mdx
```diff
@@ -0,0 +1,129 @@
...
+  );
+}
+
+const handlesActionSelectedCallback: HvAppSwitcherProps["isActionSelectedCallback"] =
+  (application) => {
+    if (typeof window !== "undefined") {
+      return window.location.href.startsWith(application.url || "");
+    }
+    return false;
+  };
```
zettca, line 104: I don't think it's worth it to keep the `window` checks. Most users won't be using SSR, and even those who do, this component will be hidden inside a client-side pop-up.
Also, I believe we don't want an empty/nullish URL to match as `true`

Or we could just remove it all together, It won't work in this page with the current links anyways
```suggestion
        isActionSelectedCallback={(app) => app.url && location.href.startsWith(app.url)}
      />
    </div>
  );
}
```

zettca (approved):

PR #4538, opened 2025-01-28
docs: add `Header` documentation

plagoa:


apps/docs/src/pages/components/header.mdx
```diff
@@ -0,0 +1,279 @@
...
+    }
+  };
+
+  const isLgUp = width === "lg" || width === "xl";
+
+  return (
+    <div className="min-h-120px w-full">
+      <HvHeader position="relative">
+        {!isLgUp && (
+          <HvIconButton aria-label="Menu">
```
zettca, line 50: `title` is a required prop of `HvIconButton`. same below
```suggestion
          <HvIconButton title="Menu">
```

apps/docs/src/pages/components/header.mdx
```diff
@@ -0,0 +1,279 @@
...
+  return { props: { ssg: { meta } } };
+};
+
+<Description />
+
+<Page>
+
+### Usage
+
+The example below shows a simple header with a brand, navigation, and actions. If you resize the window to a smaller size, the navigation will be hidden and a menu button will appear.
```
zettca, line 24: Can we have a simpler sample without Navigation? It adds quite a bit of complexity, and Pentaho+ Header designs don't seem to use/promote it anymore anyways.

Also, can we update the docs to better explain what `HvHeader` **is**? I think the key points are that it's composable (user can add any pieces), but we provide `HvHeaderBrand`, `HvHeaderNavigation`, `HvHeaderActions`.

apps/docs/src/pages/components/header.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+```tsx live
+import { useState } from "react";
+
+export default function Demo() {
+  const width = useWidth();
+  const [selected, setSelected] = useState<string>("2");
+
+  const handleChange: HvHeaderNavigationProps["onClick"] = (
```
zettca, line 33: can we inline this? it's only used in 1 place & no need for `HvHeaderNavigationProps["onClick"]`

apps/docs/src/pages/components/header.mdx
```diff
@@ -0,0 +1,279 @@
...
+    } else if (selectedItem.data?.length) {
+      setSelected(selectedItem.data[0].id);
+    }
+  };
+
+  const isLgUp = width === "lg" || width === "xl";
+
+  return (
+    <div className="min-h-120px w-full">
+      <HvHeader position="relative">
```
zettca, line 48: no need for the wrapper. we can give it a margin instead
```suggestion
      <HvHeader position="relative" className="mb-lg w-full">
```

plagoa:
@zettca updated.

zettca (approved):

PR #4539, opened 2025-01-29
docs: add `GlobalActions` documentation

plagoa:


apps/docs/src/pages/components/global-actions.mdx
```diff
@@ -0,0 +1,100 @@
...
+<div className="w-full">
+  <HvGlobalActions
+    backButton={
+      <div className="flex items-center gap-xs">
+        <HvIconButton title="Go back">
+          <Backwards />
+        </HvIconButton>
+        <Home title="Home" />
+        <HvBreadCrumb
+          listRoute={[
```
zettca, line 75: Shouldn't this be in `title` instead of `backButton`? We wouldn't need the `flex` container, as `HvGlobalActions` takes care of that for us

apps/docs/src/pages/components/global-actions.mdx
```diff
@@ -0,0 +1,100 @@
...
+            { label: "Label 1", path: "route1" },
+            { label: "Label 2", path: "route2" },
```
zettca, line 77:
```suggestion
            { label: "Label 1", path: "#route1" },
            { label: "Label 2", path: "#route2" },
```

plagoa:
@zettca updated

apps/docs/src/pages/components/global-actions.mdx
```diff
@@ -0,0 +1,100 @@
...
+The content of the Global Actions can be customized. This is useful when you want to add different elements other than the title and actions.
+
+```tsx live
+<div className="w-full">
+  <HvGlobalActions
+    title={
+      <>
+        <HvIconButton title="Go back">
+          <Backwards />
+        </HvIconButton>
```
zettca, line 72: the back button should still be in the `backButton` prop...
...unless we're deprecating it; it seems a bit redundant

plagoa: Of course, my mistake. Updated.

zettca (approved):

PR #4544, opened 2025-01-31
docs: add `Blade` documentation

plagoa:


zettca (approved):
We should eventually separate these components to a "Lab" section or equivalent, as these are experimental components. And in this is case, this component isn't a part of either Design System

zettca:
added here https://github.com/lumada-design/hv-uikit-react/pull/4545

PR #4552, opened 2025-02-04
docs: mark deprecated props

plagoa:
- Move `@deprecated` props in the props table to the end and mark them with strikethrough

image


zettca (approved):
👌

PR #4563, opened 2025-02-13
style: review focus styles

zettca:


zettca:
applying the `-1px` borderOffset overrides component's borders, which sometimes causes loss of information (eg. input error border)
image

PR #4595, opened 2025-03-10
chore: review theme related issues

plagoa:


packages/core/src/Carousel/Carousel.styles.ts
```diff
@@ -133,7 +133,7 @@ export const { staticClasses, useClasses } = createClasses("HvCarousel", {
     width: 5,
     height: 5,
     margin: theme.space.xs,
-    borderRadius: "50%",
+    borderRadius: theme.radii.circle,
```
zettca, line 136: We should probably deprecate `circle` in favor of `full`. In most cases where this is used, we'll have a `width=height`, so they'll be the same.

```suggestion
    borderRadius: theme.radii.full,
```

In cases where `width!=height`, we probably don't want `circle` anyways, because it'll lead to weird UI. Example of `circle` vs `full` in HvButton:
image image

Tailwind also has a `none` and a `full`, but (from what I can see) no utility for "circle":
https://tailwindcss.com/docs/border-radius

zettca (approved):

packages/core/src/Carousel/Carousel.styles.ts
plagoa: Yeah, I agree. I see no use case for the `circle` variant when width and height are not the same. Also, we could deprecate `none`? I see no semantic value in it.

PR #4597, opened 2025-03-11
feat(theme): add `form.errorColor` token

zettca:
- add `theme.form.errorColor` and use it in the various input/form elements

### DIFF
--- a/packages/core/src/BaseInput/BaseInput.styles.tsx
+++ b/packages/core/src/BaseInput/BaseInput.styles.tsx
@@ -62,7 +62,7 @@ export const { staticClasses, useClasses } = createClasses("HvBaseInput", {
     },
   },
   invalid: {
-    borderColor: theme.colors.negativeDeep,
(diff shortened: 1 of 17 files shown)


plagoa:
This is nice, I just don't know about the naming. In the `pentahoPlus` theme we already have `inputColors` which are specific to the `Input` component. I'd rather have this be something like `theme.forms.errorColor` or something like that.

zettca:
> This is nice, I just don't know about the naming. In the `pentahoPlus` theme we already have `inputColors` which are specific to the `Input` component. I'd rather have this be something like `theme.forms.errorColor` or something like that.

If the issue is only the naming "collisions", I'd rather change the `inputColors` to something else instead. It also isn't used only in inputs (eg. `inputColors.bg` usage), and  `inputColors` is just a private variable in a file, while `theme.*` is public API. So `theme.*` should have the "best" name, whichever that is 😅

PR #4599, opened 2025-03-12
refactor(Button): use outline instead of border

zettca:
allows removing negative margins so the button doesn't grow more than it should


zettca:
`HvMultiButton` has a strong dependency on borders

PR #4600, opened 2025-03-12
feat(CheckBox,Radio,Switch): improve styling code and update P+ and DS5 styles

plagoa:


packages/core/src/BaseCheckBox/BaseCheckBox.styles.ts
```diff
@@ -42,36 +46,33 @@ export const { staticClasses, useClasses } = createClasses("HvBaseCheckBox", {
   icon: {},
   checked: {
     "& svg": {
-      border: `1px solid ${theme.colors.text}`,
-      backgroundColor: theme.colors.text,
+      border: `1px solid ${theme.colors.primaryStrong}`,
```
zettca, line 49: the background already does the coloring
```suggestion
      borderColor: "transparent",
```

packages/core/src/CheckBox/CheckBox.styles.tsx
```diff
@@ -19,6 +19,9 @@ export const { staticClasses, useClasses } = createClasses("HvCheckBox", {
   disabled: {
     cursor: "not-allowed",
     "& $label": { color: theme.colors.textDisabled, cursor: "not-allowed" },
+    "&:hover": {
+      backgroundColor: "transparent",
+    },
```
zettca, line 24: isn't this redundant? we already have it in `BaseCheckBox` (with lots of specificity)

packages/core/src/BaseCheckBox/BaseCheckBox.styles.ts
```diff
@@ -42,36 +46,33 @@ export const { staticClasses, useClasses } = createClasses("HvBaseCheckBox", {
...
+      backgroundColor: theme.colors.primaryStrong,
```
zettca, line 50: this affects DS3 😞

zettca (commented):
looking nice 👌
but there are some unintended side-effects:

packages/core/src/BaseCheckBox/BaseCheckBox.styles.ts
```diff
@@ -13,14 +13,15 @@ export const { staticClasses, useClasses } = createClasses("HvBaseCheckBox", {
...
+      borderRadius: theme.radii.round,
     },
 
     "& svg": {
       width: 16,
       height: 16,
       color: theme.colors.bgContainer,
-      borderRadius: theme.radii.none,
-      border: `1px solid ${theme.colors.text}`,
+      borderRadius: theme.radii.base,
```
zettca, line 23: unsure if it's related to this change, but the new radius is showing a duplicated & rounded error border.

image

packages/styles/src/tokens/colors.ts
```diff
@@ -320,8 +320,8 @@ const newLight = {
 
   border: atmosphereLight.atmo4,
   borderSubtle: atmosphereLight.atmo3,
-  borderStrong: atmosphereLight.atmo2,
-  borderDisabled: atmosphereLight.atmo3,
+  borderStrong: accentLight.secondary_80,
```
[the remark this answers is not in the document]
plagoa: Yeah. In the `Tooltip` we were actually doing it wrong. We were using `atmo2` instead of `atmo3` (`borderSubtle`). In the `VerticalNavigation` it's the same thing.

packages/core/src/BaseCheckBox/BaseCheckBox.styles.ts
plagoa: Well, that's why we have theme overrides 👀

packages/core/src/BaseCheckBox/BaseCheckBox.styles.ts
plagoa: Yeah, we have two "invalid state" lines in the component. One for when we only have the checkbox and the other to the container (that includes) the label. When we have a label is just a matter of removing the line from the checkbox (the container already has it hence the double border). But when we only have the checkbox the case is trickier. I went for a `::after` pseudo-class and I think it works. Let me know what you think.

PR #4602, opened 2025-03-12
test: remove icons dependency

zettca:
using UI Kit icons is unnecessary to testing the components
this should also (theoretically) improve test performance and reliability


plagoa:
Good idea 👌

PR #4614, opened 2025-03-19
fix(MultiButton): incorrect border styles when the `MultiButton` has only one button

plagoa:
- Fixes the incorrect border styles when the `MultiButton` only has one `Button`
  - The left border radius was being incorrectly overridden by the last button styles
- Added visual test

### DIFF
--- a/packages/core/src/MultiButton/MultiButton.styles.tsx
+++ b/packages/core/src/MultiButton/MultiButton.styles.tsx
@@ -11,7 +11,6 @@ export const { staticClasses, useClasses } = createClasses("HvMultiButton", {
   },
   multiple: {
     backgroundColor: theme.colors.bgPage,
-
     borderWidth: 0,
     borderColor: `${theme.colors.border} transparent`,
     borderRadius: theme.radii.base,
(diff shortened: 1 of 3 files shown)


packages/core/src/MultiButton/MultiButton.styles.tsx
```diff
@@ -47,6 +47,9 @@ export const { staticClasses, useClasses } = createClasses("HvMultiButton", {
           borderRightColor: theme.colors.border,
         },
       },
+      "&$firstButton$lastButton": {
+        borderRadius: theme.radii.base,
+      },
```
zettca, line 52: this is doing nothing. DS5 doesn't have this bug.
also, we should use `"inherit"` for these cases. this is making it harder to override

packages/core/src/MultiButton/stories/MultiButton.stories.tsx
```diff
@@ -288,6 +288,17 @@ export const Test: StoryObj = {
...
+          <HvMultiButton>
+            <HvButton>Label 1</HvButton>
+          </HvMultiButton>
+          <HvMultiButton>
+            <HvButton selected>Label 1</HvButton>
+          </HvMultiButton>
+          <HvMultiButton>
+            <HvButton disabled>Label 1</HvButton>
+          </HvMultiButton>
+        </div>
```
zettca, line 301: this is dragging all the other samples unnecessarily. let's inline this alongside the `width: fit-content` icon-only `HvMultiButton`

```tsx
<div className="flex flex-1 [&>*]:flex-1 gap-xs">
  <HvMultiButton>
    ...
```

image

zettca (changes requested):

zettca (approved):

PR #4634, opened 2025-04-01
feat(NumberInput): add `NumberInput` component

plagoa:


packages/core/src/NumericInput/NumericInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+export { staticClasses as numericInputClasses };
+
+export type HvNumericInputClasses = ExtractNames<typeof useClasses>;
+
+export interface HvNumericInputProps
+  extends Omit<HvInputProps<HTMLInputElement>, "inputProps"> {
+  classes?: HvNumericInputClasses;
+  /** Properties passed on to the input element. */
+  inputProps?: React.InputHTMLAttributes<HTMLInputElement>;
+}
```
zettca, line 22: We should move `inputProps` to `HvBaseInput`. The problem with MUI's is that isn't the `HTMLInputElement | HTMLTextArea` elements, but the fact that is extends `React.HTMLAttributes` and not `React.InputHTMLAttributes`. Since `React.InputHTMLAttributes` itself extends `React.HTMLAttributes`, it's a safe/compatible change.

Also, we should be able to pass other `HvInputClasses` classes, instead of being limited to `HvNumericInputClasses`

```suggestion
export interface HvNumericInputProps extends HvInputProps {}
```

packages/core/src/Input/stories/Controlled.tsx
```diff
@@ -13,7 +13,10 @@ export default function Controlled() {
         placeholder="Enter value"
         value={value}
         // to be possible to change the input value by user action
-        onChange={(event, newValue) => setValue(newValue)}
+        onChange={(event, newValue) => {
+          console.log("YEY", newValue);
+          setValue(newValue);
+        }}
```
zettca, line 19: 👀

plagoa:
@zettca updated PR.

apps/docs/src/pages/components/number-input.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+<br />
+
+<HvBannerContent variant="warning">
+  <HvTypography component="span">
+    The `HvNumberInput` component shares the same API as the `HvInput`
+    component. Please check the [HvInput component page](/components/input) for
+    documentation.
+  </HvTypography>
+</HvBannerContent>
```
zettca, line 41: Why not use Nextra's `Callout` component like we're doing elsewhere?
`HvBannerContent` has no icon, show a close icon that doesn't work, etc.

[Callout](https://nextra.site/docs/built-ins/callout) also has a warning variant, but honestly it feels a bit excessive - what's there to be careful about? 😅

packages/core/src/Input/Input.tsx
```diff
@@ -107,32 +108,41 @@ export interface HvInputProps<
   /** The error message to show when `status` is "invalid". */
   statusMessage?: string;
   /** @inheritdoc */
-  onChange?: (event: React.ChangeEvent<InputElement>, value: string) => void;
+  onChange?: (
+    event: React.ChangeEvent<HTMLInputElement>,
```
zettca, line 112: We can't do this... `InputElement` is the generic (maybe I could've named it something more obvious like `TInput` or `InputType`), and could be a TextArea or something that extends input.

We just need to do `HvInputProps<HTMLInputElement>` on the other side

packages/core/src/utils/input.ts
```diff
@@ -0,0 +1,16 @@
+/** Changes a given `input`'s `value`, triggering its `onChange` */
+export const changeInputValue = (
```
zettca, line 2: If this util is for the input, shouldn't it stay closer to the Input component? Can we move this to `Input/utils.ts`?
`HvNumberInput` is already importing other stuff from `Input`!

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { HvIcon } from "../icons";
+import { HvInput, HvInputProps } from "../Input";
+import { changeInputValue } from "../utils/input";
+import { staticClasses, useClasses } from "./NumberInput.styles";
+
+export { staticClasses as numericInputClasses };
+
+export type HvNumberInputClasses = ExtractNames<typeof useClasses>;
+
+export interface HvNumberInputProps extends HvInputProps {}
```
zettca, line 18:
```suggestion
export interface HvNumberInputProps extends HvInputProps<HTMLInputElement> {}
```

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import { HvAdornment } from "../FormElement";
+import { HvIcon } from "../icons";
+import { HvInput, HvInputProps } from "../Input";
+import { changeInputValue } from "../utils/input";
+import { staticClasses, useClasses } from "./NumberInput.styles";
+
+export { staticClasses as numericInputClasses };
+
+export type HvNumberInputClasses = ExtractNames<typeof useClasses>;
```
zettca, line 16: this is unused. we need to find a way to be able to use `HvInput`'s `classes`, and be able to override it

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export interface HvNumberInputProps extends HvInputProps {}
+
+export const HvNumberInput = forwardRef<
```
zettca, line 20: missing the doc block!
maybe we could mention it uses the `HvInput` API here?

packages/core/src/NumberInput/NumberInput.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <HvTypography
+              className="color-primary underline underline-primary"
+              component="a"
+              href="/?path=/docs/components-input--docs"
+            >
+              HvInput component page
+            </HvTypography>{" "}
+            for documentation.
+          </HvTypography>
+        </HvBannerContent>
```
zettca, line 45: let's please not put docs inside the code samples.
the component is also missing a doc block, we could add something similar to this there

zettca (changes requested):

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import { HvAdornment } from "../FormElement";
+import { HvIcon } from "../icons";
+import { HvInput, HvInputProps } from "../Input";
+import { changeInputValue } from "../Input/utils";
+import { staticClasses, useClasses } from "./NumberInput.styles";
+
+export { staticClasses as numericInputClasses };
+
+export type HvNumberInputClasses = ExtractNames<typeof useClasses>;
```
zettca, line 16: these are unused and redundant. the component uses `HvInputClasses`

```suggestion
```

packages/core/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <HvInput
+      ref={forkedRef}
+      type="number"
+      className={cx(classes.root, className)}
+      endAdornment={!readOnly && buttons}
+      classes={{
+        adornmentsBox: classes.adornmentsBox,
+      }}
```
zettca, line 75: we need for forward all other `classes.*` too, otherwise we can't override them
```suggestion
      endAdornment={!readOnly && buttons}
      classes={classes}
```

zettca (approved):

PR #4638, opened 2025-04-03
chore: update radii token use

plagoa:


packages/core/src/SnackbarProvider/SnackbarProvider.styles.ts
```diff
@@ -15,7 +16,7 @@ export const { useClasses, staticClasses } = createClasses(
         fontFamily: "inherit",
         fontWeight: "inherit",
         lineHeight: "inherit",
-        borderRadius: "0",
+        borderRadius: theme.radii.none,
```
zettca, line 19: Can we leave these as they were? We gain nothing from this...

Reasons:
- we're adding a new external dependency on the `theme` object
- the output will be `border-radius: var(--uikit-radii-none)` instead of simply `border-radius: 0`
- consistency. we're not using `theme.radii.none` in other places (apart from a single instance in `BannerContent.styles.tsx`)

The "none" is indeed useful for using with utility classes and components props:
```tsx
<div className="rounded-none" />
<HvButton radius="none" />
```

packages/pentaho/src/Canvas/PanelTab/PanelTab.styles.tsx
```diff
@@ -12,7 +12,7 @@ export const { staticClasses, useClasses } = createClasses("HvCanvasPanelTab", {
     alignItems: "center",
     justifyContent: "center",
     flex: 1,
-    borderRadius: "16px 16px 0 0",
+    borderRadius: `${theme.radii.large} ${theme.radii.large} 0 0`,
```
zettca, line 15: this would be (which I'm not suggesting, please don't 😅 ): 
```
borderRadius: `${theme.radii.large} ${theme.radii.large} ${theme.radii.none} ${theme.radii.none}`,
```

packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -585,9 +587,9 @@ const pentahoPlus = makeTheme((theme) => ({
         hasIcon: {
           paddingLeft: theme.space.xs,
         },
-        xs: { borderRadius: 4 },
-        sm: { borderRadius: 6 },
-        md: { borderRadius: 8 },
+        xs: { borderRadius: theme.radii.base },
+        sm: { borderRadius: theme.radii.base },
+        md: { borderRadius: theme.radii.round },
```
zettca, line 592: these are wrong. they need to be 4/6/8

plagoa:
@zettca updated the PR.

zettca (approved):

packages/core/src/SnackbarProvider/SnackbarProvider.styles.ts
```diff
@@ -1,4 +1,5 @@
 import { createClasses } from "@hitachivantara/uikit-react-utils";
+import { theme } from "@hitachivantara/uikit-styles";
```
zettca, line 2:
```suggestion
```

PR #4639, opened 2025-04-03
docs: add inputs examples section

plagoa:


apps/docs/src/examples/inputs/TagsDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+    setSelectedColors((prev) => prev.filter((c) => c !== color));
+  };
+
+  const focusOnContainer = () => {
+    focusTarget.current?.focus();
+  };
+
+  return (
+    <div className="w-300px">
+      <HvLabel>Tags dropdown input</HvLabel>
```
zettca, line 57: If we're using `HvLabel`, then it needs to be accessible. (ie. `id`/`aria-labelledby`). Also, the `HvLabel`'s "label" is via the `label` prop, not `children`

Alternatively, we can use `aria-label` on the `HvBaseDropdown`

apps/docs/src/examples/inputs/TagsDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+    });
+  };
+
+  const handleRemoveColor = (color: string) => {
+    setSelectedColors((prev) => prev.filter((c) => c !== color));
+  };
+
+  const focusOnContainer = () => {
+    focusTarget.current?.focus();
+  };
```
zettca, line 53: dead code

apps/docs/src/examples/inputs/PhoneInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+  },
+];
+
+export default function Demo() {
+  const [selectedCountry, setSelectedCountry] = useState("Portugal");
+  const [formatttedPhoneNumber, setFormattedPhoneNumber] = useState("");
+
+  return (
+    <div className="w-300px">
+      <HvInput
```
zettca, line 54: let's avoid useless `div`s
```suggestion
      <HvInput
        className="w-300px"
```

apps/docs/src/examples/inputs/PhoneInput.tsx
```diff
@@ -0,0 +1,91 @@
...
+        classes={{
+          inputRoot: "overflow-visible",
+        }}
```
zettca, line 57: let's use `enablePortal` on `HvSelect` instead. this is slightly clipping the border:

image  image

apps/docs/src/examples/inputs/TagsDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const containerRef = useRef<HTMLDivElement>(null);
+  const focusTarget = useRef<HTMLDivElement>(null);
+
+  const [selectedColors, setSelectedColors] = useState<string[]>([]);
+
+  useEffect(() => {
+    if (containerRef.current) {
+      containerRef.current.scrollLeft = containerRef.current.scrollWidth;
+    }
+  }, [selectedColors]);
```
zettca, line 38: we own `selectedColors`, so this `useEffect` could easily be avoided by moving the code to where we do `setSelectedColors()`

apps/docs/src/examples/inputs/TagsDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+                <HvTag
+                  autoFocus={idx === 0}
+                  key={color}
+                  label={color}
+                  onClick={() => handleAddColor(color)}
+                  onKeyDown={(e) => {
+                    if (e.key === "Enter" || e.key === " ") {
+                      handleAddColor(color);
+                    }
+                  }}
```
zettca, line 93: these tags _are_ buttons, we don't need this
```suggestion
```

apps/docs/src/examples/inputs/TagsSuggestions.tsx
```diff
@@ -0,0 +1,114 @@
...
+  };
+
+  const handleRemoveColor = (color: string) => {
+    setSelectedColors((prev) => prev.filter((c) => c !== color));
+  };
+
+  return (
+    <div className="w-300px">
+      <HvLabel>Tags with suggestions</HvLabel>
+      <HvTagsInput
```
zettca, line 58: same as above, but `HvTagsInput` takes in a `label`

```suggestion
      <HvTagsInput
        label="Tags with suggestions"
```

apps/docs/src/examples/inputs/TagsSuggestions.tsx
```diff
@@ -0,0 +1,114 @@
...
+            className="cursor-pointer"
+            size="xs"
+          />
+        }
+        classes={{
+          tagsList: "h-32px pr-0 overflow-hidden",
+        }}
+      />
+      <Popper anchorEl={containerRef.current} open={open}>
+        <HvPanel className="flex flex-wrap flex-col gap-xs w-300px bg-bgContainer border-rounded-16px border border-atmo3">
```
zettca, line 83: maybe there's a better middle ground, but the `HvPanel` defaults currently look better IMO. `HvPanel` already has the correct `backgroundColor`, and a better `borderColor`
```suggestion
        <HvPanel className="flex flex-wrap flex-col gap-xs w-300px border">
```

image 
`HvPanel` default:
image

apps/docs/src/pages/examples/inputs.mdx
```diff
@@ -0,0 +1,27 @@
...
+import tagsSuggestions from "../../examples/inputs/TagsSuggestions?raw";
+
+# Inputs examples
+
+## Phone input
+
+A phone input that includes a dropdown to select the country code. The input is automatically formatted based on the selected country code.
+In the example the formatting function is always the same and formats the number as `XX XXX XX XX`.
+
+<CodeBlock code={{ main: phoneInput }} />
```
zettca, line 14: kinda unrelated, but reviewing these samples with the "toggable" layout was a pain.
maybe we could use `layout="expandable"` for these?

apps/docs/src/components/code/ToggableLayout.tsx
```diff
@@ -59,7 +59,7 @@ export const ToggableLayout = ({ title, scope, code }: ToggableLayoutProps) => {
       {showPreview ? (
         <div
           className={clsx(
-            "p-md pt-lg min-h-400px mb-lg",
+            "p-md pt-lg min-h-100px mb-lg",
```
zettca, line 62: if we use the (IMO much more suitable) `"expandable"` layout, we don't need to change this 🙃

zettca:
samples looking great overall btw 👌

apps/docs/src/pages/examples/inputs.mdx
```diff
@@ -0,0 +1,15 @@
...
+
+import copyButton from "../../examples/inputs/CopyButton?raw";
+import dropdownPrefix from "../../examples/inputs/DropdownPrefix?raw";
+import tagsDropdown from "../../examples/inputs/TagsDropdown?raw";
+import tagsSuggestions from "../../examples/inputs/TagsSuggestions?raw";
+
+# Inputs examples
+
+<div className="w-full grid [grid-template-columns:repeat(auto-fit,minmax(320px,1fr))] gap-md">
+  <CodeBlock code={{ main: dropdownPrefix }} layout="popup" />
```
zettca, line 11: when we have only 1 file, we can do:
```suggestion
  <CodeBlock code={dropdownPrefix} layout="popup" />
```

zettca (approved):
nice 👌

francisco-guilherme (approved):

PR #4650, opened 2025-04-09
feat(SearchInput): add `SearchInput` component

plagoa:


packages/core/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { HvIcon } from "../icons";
+import { HvInput, HvInputProps } from "../Input";
+import { fixedForwardRef } from "../types/generic";
+
+export const HvSearchInput = fixedForwardRef(function HvSearchInput(
+  props: HvInputProps<HTMLInputElement>,
+  ref: React.Ref<HTMLInputElement>,
+) {
+  return (
+    <div>
```
zettca, line 11: image

```suggestion
```

packages/core/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const { classes } = useClasses(classesProp);
+
+  return (
+    <HvInput
+      ref={ref}
+      type="search"
+      disableSearchButton
+      startAdornment={
+        <HvAdornment icon={<HvIcon compact name="Search" title={"Search"} />} />
```
zettca, line 33: this isn't/doesn't need to be a11y visible, so we don't need the label. and that's good, because we shouldn't hard-code it otherwise
```suggestion
        <HvAdornment icon={<HvIcon compact name="Search" />} />
```

zettca (approved):

PR #4652, opened 2025-04-09
refactor(InputHeader): add component & use in BaseDropdown, Select, TagsInput

zettca:


zettca:
Currently, only this `HvBaseDropdown` and `HvTagsInput` can leverage this refactor, as `HvBaseButton` is coupled to MUI's `ButtonBase`, and `HvSelect` is coupled to `HvButton`

PR #4655, opened 2025-04-10
docs(CheckBoxGroup): add component

zettca:
also improves input adornments

### DIFF
--- a/apps/docs/src/pages/components/checkbox-group.mdx
+++ b/apps/docs/src/pages/components/checkbox-group.mdx
@@ -1,0 +1,77 @@
+import {
+  checkBoxGroupClasses,
+  HvCheckBox,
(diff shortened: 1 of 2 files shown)


plagoa:
I noticed that when we have the `CheckBoxGroup` marked as required then the asterisk is not displayed:
image

The same happens with the `RadioGroup` but it doesn't happen with the `Input` for instance:
image

zettca:
> I noticed that when we have the `CheckBoxGroup` marked as required then the asterisk is not displayed if no label is present: image
> 
> The same happens with the `RadioGroup` but it doesn't happen with the `Input` for instance: image

I'll open an issue for it. We should probably have a more uniform implementation/dedupe our form component label/descriptions

plagoa (approved):

PR #4666, opened 2025-04-16
docs: fix broken links

HQFOX:
I found some links in table hook links that were not working properly. I couldn't launch the storybook locally due to an error (`SB_CORE-SERVER_0005 (MainFileESMOnlyImportError): Storybook failed to load .storybook/main.ts`) to verify the fix but it should be it.


zettca (approved):
thanks 👌

PR #4669, opened 2025-04-19
refactor: theming refactor for improved maintainability 

francisco-guilherme:


packages/core/src/index.ts
```diff
@@ -140,7 +140,6 @@ export * from "./hooks/useUniqueId";
 export * from "./hooks/useWidth";
 
 export * from "./providers/Provider";
-export * from "./providers/ThemeProvider";
```
zettca, line 143: this is a breaking change

zettca (changes requested):

packages/core/src/index.ts
francisco-guilherme: I know, but is there even any use case where the ThemeProvider is being used instead of the Provider?

Issue #4592, opened 2025-03-07
The "Select All" checkbox in Multi-Select Dropdowns don't respect the "disabled" attribute of some options


ronakkaria:
### Which UI Kit version is this bug for?

v5.x

### Latest version

- [x] I have tested the latest version

### No Hitachi confidential content

- [x] I'm not including any Hitachi confidential content

### Current behavior 😯

If you create a multi-select dropdown, with some of the options being "disabled" using the "disabled" flag in the value object of that option. 
The option indeed becomes disabled and the user is unable to select / un-select these options individually, however, they are still able to click on the "select-all" checkbox and select / unselect all the options.

### Expected behavior 🤔

If we have some disabled options in the list, then the select-all checkbox should not modify the state of the disabled options (these disabled options may be selected / unselected - but their value should never change). 

### Steps to reproduce 🕹

Sandbox link: 

Steps to reproduce:

1. Create a multi-select dropdown
2. In the values prop, add some options with a "disabled: true" flag. 
3. The options are individually disabled
3. However, clicking on the select-all checkbox on top of the list, will modify the state of these options as well
4.

### Context 🔦

We have a scanario where the user is not allowed to select some options,
however, it's still useful for them to see those options in the list with a custom tooltip showing why they are disabled (so that in case they really want to select these options, they know what they need to do in order to enable them).

plagoa:
@ronakkaria thanks for the feedback, we opened an issue to address this:
https://hv-eng.atlassian.net/browse/HVUIKIT-7288

PR #4673, opened 2025-04-22
fix(Dropdown): ignore `disabled` items on select all

plagoa:
Addresses https://github.com/lumada-design/hv-uikit-react/issues/4592

### DIFF
--- a/packages/core/src/Dropdown/Dropdown.test.tsx
+++ b/packages/core/src/Dropdown/Dropdown.test.tsx
@@ -177,4 +177,34 @@ describe("Dropdown", () => {
     await userEvent.click(getCheckBox());
     expect(getApplyButton()).toBeDisabled();
   });
+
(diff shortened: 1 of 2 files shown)


packages/core/src/Dropdown/Dropdown.test.tsx
```diff
@@ -177,4 +177,34 @@ describe("Dropdown", () => {
     await userEvent.click(getCheckBox());
     expect(getApplyButton()).toBeDisabled();
   });
+
+  it.only("should not select disabled items when Select All is clicked", async () => {
```
zettca, line 181: [`.only`](https://vitest.dev/guide/filtering) disables other tests
```suggestion
  it("doesn't select disabled items when Select All is clicked", async () => {
```

plagoa: Sorry, left that behind by mistake. Fixed.

PR #4674, opened 2025-04-22
fix: update radius on several input components

plagoa:


packages/core/src/BaseDropdown/BaseDropdown.styles.tsx
```diff
@@ -92,16 +92,19 @@ export const { useClasses, staticClasses } = createClasses("HvBaseDropdown", {
...
     padding: 0,
     border: `1px solid ${theme.colors.text}`,
+    "&[data-popper-placement*='top']": {
+      top: 1,
+      borderRadius: `${theme.radii.round} ${theme.radii.round} 0 0`,
+    },
+    "&[data-popper-placement*='bottom']": {
+      top: -1,
+      borderRadius: `0 0 ${theme.radii.round} ${theme.radii.round}`,
+    },
```
zettca, line 102: looking at the visual changes, this should probably be moved to the `ds5.ts` theme, because:
- these are harder to override
- to not affect DS3
- `ds3` and `pentahoPlus` always the same border (0 and `round` respectively)

plagoa: Yeah, that was wrong. I updated the PR.

zettca (approved):

PR #4676, opened 2025-04-22
docs: improve layout of charts examples page

plagoa:
This changes the layout of the charts examples page to make better use of the available space.

⚠️ This also adds a new echarts component to the default components loaded by the `HvBaseChart`:

```echarts.use([CanvasRenderer, AriaComponent, TitleComponent]);```

 As it is, the `HvBaseChart` component loads two components only, necessary for rendering charts (`CanvasRenderer`) and add accessibility information (`AriaComponent`). Then each chart will add the components that it needs. For example, the `HvBarChart` will add:

 ```
 echarts.use([
  BarChart,
  DatasetComponent,
  GridComponent,
  TooltipComponent,
  LegendComponent,
  DataZoomSliderComponent,
  DataZoomInsideComponent,
  MarkLineComponent,
]);
```

but if the user needs to add a title to the chart he's using he can't do it. He needs to register the `TitleComponent` also. We may want to revisit this in the future and add some mechanics to allow the user to pass some extra modules for the `BaseChart` to register besides the default ones.


packages/viz/src/BaseChart/BaseChart.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { AriaComponent, TitleComponent } from "echarts/components";
 import * as echarts from "echarts/core";
 import { CanvasRenderer } from "echarts/renderers";
 
 import { useVizTheme } from "../hooks";
 import { HvChartCommonProps, HvEChartsOption } from "../types/common";
 
 // Register chart components
-echarts.use([CanvasRenderer, AriaComponent]);
+echarts.use([CanvasRenderer, AriaComponent, TitleComponent]);
```
zettca, line 12: Can we not?
We're adding "bloat" to the base chart, to add custom text in a docs sample that doesn't add much to it.
Besides, it doesn't adjust in dark mode, and feels misaligned 

image

apps/docs/src/components/code/PopupLayout.tsx
```diff
@@ -80,9 +80,7 @@ export const PopupLayout = ({ scope, code }: PopupLayoutProps) => {
       />
 
       {/* Preview Section */}
-      <div className="p-md flex items-center justify-center">
-        <div>{element}</div>
-      </div>
+      <div className="p-md flex items-center justify-center">{element}</div>
```
zettca, line 83: This will affect other samples. The parent of the samples must be "clean", without custom layouts

Please revert.

plagoa:
@zettca updated the PR to remove the addition of the `TitleComponent`. The change to the `PopupLayout` was intentional.

PR #4680, opened 2025-04-24
fix(Popper): align styles of poppers with specs

plagoa:
⚠️ Changes in Chromatic are expected and valid.

### DIFF
--- a/packages/styles/src/themes/pentahoPlus.ts
+++ b/packages/styles/src/themes/pentahoPlus.ts
@@ -1171,6 +1171,20 @@ const pentahoPlus = makeTheme((theme) => ({
             borderColor: theme.colors.textDimmed,
           },
         },
+        baseContainer: {
+          "&[data-popper-placement=bottom-end] .HvBaseDropdown-panel": {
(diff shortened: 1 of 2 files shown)


packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -1098,8 +1132,13 @@ const pentahoPlus = makeTheme((theme) => ({
             borderColor: theme.colors.textDimmed,
           },
         },
-        panel: {
-          borderColor: theme.colors.textDimmed,
+        panelOpenedDown: {
+          top: 0,
+          ...popperStyles(theme),
```
zettca, line 1137: image

If this is the spec, then the spec must be wrong... 😅
there's a huge mismatch between the popper and the input "header"

looking at the "current" TimePicker spec, this "popper" shouldn't be used:

image

We need to better understand where exactly this new popper is to be used first

zettca (changes requested):

plagoa:
@zettca updated after feedback.

PR #4681, opened 2025-04-24
docs: add ThemeSwitcher to docs

plagoa:
The goal here is to add a Theme Switcher to the docs so that we can view the docs in the several themes we support. The goal is *not* to change the theme on the whole docs, just change it at sample level. Visuals can still be tweaked, of course.


apps/docs/src/contexts/DocsThemeContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface DocsThemeContextValue {
+  docsTheme: DocsTheme;
+  setDocsTheme?: Dispatch<SetStateAction<DocsTheme>>;
+}
+
+const DEFAULT_THEME = {
+  theme: "pentahoPlus",
+  mode: "dawn",
+};
```
zettca, line 23: Having a static `pentahoPlus-dawn` default is bad UX IMO.
Shouldn't we instead leverage the existing light/dark switcher, and keep this exclusively for the theme (`ds3`, `ds5`, `pentahoPlus`) instead?

image

apps/docs/src/components/home/ThemeSwitcher.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useDocsThemeContext } from "../../contexts/DocsThemeContext";
+
+const themesModes = [
+  "pentahoPlus-dawn",
+  "pentahoPlus-wicked",
+  "ds5-dawn",
+  "ds5-wicked",
+  "ds3-dawn",
+  "ds3-wicked",
+];
```
zettca, line 13: what's I was suggesting above is something like this
```suggestion
const themes = ["pentahoPlus","ds5","ds3"];
```

apps/docs/src/contexts/DocsThemeContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+});
+
+interface DocsThemeProviderProps {
+  children?: React.ReactNode;
+}
+
+export const DocsThemeProvider = ({ children }: DocsThemeProviderProps) => {
+  const [docsTheme, setDocsTheme] = useState<DocsTheme>(DEFAULT_THEME);
+
+  const value = useMemo(
```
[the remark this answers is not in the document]
plagoa: @zettca  I can add `localStorage` to persist state across reloads, but removing the context we no longer have the shared state between the components (that is, changing the theme on the header won't trigger the samples to change the theme they're using). what do you think?

plagoa:
@zettca made several changes:
 - the `ThemeSwitcher` only changes theme. 
 - the color mode is defined by Nextra's `ThemeSwitch` component we were already using
 - local storage is used to persist state between reloads

apps/docs/src/components/code/ExpandableLayout.tsx
```diff
@@ -41,18 +49,25 @@ export const ExpandableLayout = ({ scope, code }: ExpandableLayoutProps) => {
...
-      <div
-        className={clsx(
-          "p-md pt-lg bg-bgPage border rounded-inherit",
-          isExpanded && "rounded-b-0",
-        )}
+      <HvProvider
+        themes={[pentahoPlus, ds5, ds3]}
+        theme={docsTheme}
+        colorMode={docsMode}
+        cssTheme="scoped"
```
zettca, line 56: Let's extract this to a separate component (eg. `HvDocsProvider`)?
We're repeating it quite a few times (per "layout" renderer + `Playground`) - and most likely want to keep the exact same behaviour. Also, we'll probably want to add a few more renders down the road

apps/docs/src/components/code/Playground.tsx
```diff
@@ -105,8 +113,16 @@ export const Playground = ({
       {/* Component preview and controls */}
       <div className="grid grid-cols-[2fr_1fr] border rounded-t-round">
         {/* Preview Area */}
+
         <div className={"grid place-items-center p-sm"}>
-          {decorator ? decorator(componentElement) : componentElement}
+          <HvProvider
```
zettca, line 118: we need before the `div.grid` above, otherwise the UI gets messed up. we're using the styles above for the layout

apps/docs/src/pages/examples/inputs.mdx
```diff
@@ -10,10 +10,10 @@ import tagsSuggestions from "../../examples/inputs/TagsSuggestions?raw";
 
 <div className="overflow-hidden">
   <div className="w-101% grid  divide-border [grid-template-columns:repeat(auto-fit,minmax(360px,1fr))] -mt-px -mb-px">
-    <CodeBlock code={{ main: dropdownPrefix }} layout="popup" />
-    <CodeBlock code={{ main: copyButton }} layout="popup" />
-    <CodeBlock code={{ main: tagsDropdown }} layout="popup" />
-    <CodeBlock code={{ main: tagsSuggestions }} layout="popup" />
-    <CodeBlock code={{ main: editableValue }} layout="popup" />
+    <CodeBlock id="1" code={{ main: dropdownPrefix }} layout="popup" />
```
zettca, line 13: what is this for? it's not even part of `CodeBlock`'s API 🤔

apps/docs/src/components/code/ExpandableLayout.tsx
```diff
@@ -41,18 +49,25 @@ export const ExpandableLayout = ({ scope, code }: ExpandableLayoutProps) => {
...
+        <div
+          className={clsx(
+            "p-md pt-lg bg-bgPage border rounded-inherit",
+            isExpanded && "rounded-b-0",
+          )}
+        >
+          {liveError ? (
+            <div className="text-negative">{liveError}</div>
+          ) : (
+            <>{element}</>
```
zettca, line 67: why? the empty `div` is there so we don't mess up the samples when we need to change the container's layout (which we've done before multiple times). we talked about this

zettca: we should probably abstract this away and document it properly

apps/docs/src/components/code/ExpandableLayout.tsx
```diff
@@ -41,18 +42,27 @@ export const ExpandableLayout = ({ scope, code }: ExpandableLayoutProps) => {
...
+            "p-md py-[40px] bg-bgPage border rounded-inherit",
+            isExpanded && "rounded-b-0",
+          )}
+        >
+          <div
+            className={clsx(
+              "p-md pt-lg bg-bgPage border rounded-inherit",
+              isExpanded && "rounded-b-0",
+            )}
+          >
```
zettca, line 57: this shouldn't be here - probably issues with the rebase
image
image

apps/docs/src/components/code/PopupLayout.tsx
```diff
@@ -43,42 +47,53 @@ export const PopupLayout = ({ scope, code }: PopupLayoutProps) => {
...
-      <HvDialog
-        onClose={() => setIsExpanded(false)}
-        open={isExpanded}
-        fullWidth
-        maxWidth="md"
-      >
-        <HvDialogTitle>Code</HvDialogTitle>
-        <HvDialogContent className="py-sm">
+    <div id={id} className="bg-transparent h-full">
+      <section className="relative border-1 border-t-0 -ml-px h-full">
```
zettca, line 51: is this being used? the `id="1"` samples seem to be gone. 🤔 and if so, it'd be nice if we could merge these

```suggestion
    <section id={id} className="bg-transparent relative border-1 border-t-0 -ml-px h-full">
```

PR #4688, opened 2025-05-05
chore(theme): add makeColors utility & map UI Kit <-> NEXT tokens

zettca:
- add `makeColors` utility, reducing boilerplate when defining theme colors
  - allows passing a `[light, dark]` colors tuple, instead of having to (or forgetting to) update a separate object
  - easier to maintain & update tokens from new DS updates, as tokens are typically side-by-side
- add a mapping function `extendCompatColors` that ensures the "old" colors are also updated
  - eg. updating `bgPage` ensures `backgroundColor` & `atmo2` are also updated/synced


packages/styles/src/makeTheme.ts
```diff
@@ -18,3 +19,42 @@ export const makeTheme = <Mode extends string = string>(
 
   return newTheme;
 };
+
+/** Takes in a Record of `HvThemeColors`, where the values can */
```
plagoa, line 73: _"where the values can"_ is a great cliffhanger 😄

zettca: oops, updating docs is hard (or boring) 😅

plagoa (approved):

PR #4691, opened 2025-05-05
chore: several fixes

plagoa:
This adds or fixes a few small things:
- Selected tab on the docs header had the incorrect color (not visible in dark mode)
- Added an icon to the KPIs card on the examples home page
- Added a banner and a KPI to the docs home page to showcase a couple more UI Kit components

### DIFF
--- a/apps/docs/src/components/home/CardsSection.tsx
+++ b/apps/docs/src/components/home/CardsSection.tsx
@@ -1,3 +1,4 @@
+import { ActiveUsers } from "./cards/ActiveUsers";
 import { ApiUsage } from "./cards/ApiUsage";
 import { DataConfig } from "./cards/DataConfig";
 import { DataInsights } from "./cards/DataInsights";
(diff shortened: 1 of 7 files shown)


apps/docs/src/pages/globals.css
```diff
@@ -94,7 +94,7 @@ body,
 }
 
 [aria-current="true"].nextra-focus {
-  @apply font-medium text-dark;
+  @apply font-medium text-secondary;
```
zettca, line 97: the token is `text`
```suggestion
  @apply font-medium text-text;
```

apps/docs/src/components/home/CardsSection.tsx
```diff
@@ -17,6 +18,7 @@ export const CardsSection = () => {
       <div className="space-y-4">
         <FinancialPerformance />
         <StorageOverview />
+        <ActiveUsers />
```
zettca, line 21: This looks a bit out of place, as it differs from all other containers. We do have something similar in the [App](https://lumada-design.github.io/uikit-app/master/)'s components page, but that page was kind of meant for debugging - this one is meant to be "pretty", right? 😅 

This looks good enough for me, but I'd like @francisco-guilherme's opinion on this too

image

zettca (approved):

PR #4697, opened 2025-05-06
docs: add Status  and Theming pages

francisco-guilherme:


apps/docs/src/pages/docs/status.mdx
```diff
@@ -0,0 +1,39 @@
+# Project Status
+
+**UI Kit** is actively maintained by a dedicated team committed to enabling high-quality digital product development. Stay informed about the project’s current status and ongoing development.
+
+## Design System Support
+
+Originally built on the [Next Design System](https://designsystem.hitachivantara.com/6c705d900/p/43c9a1-next-design-system), **UI Kit** has since evolved into a **design system–agnostic** library, powered by a flexible [theming system](/docs/theming) that allows teams to meet diverse branding and design requirements.
+
+Recent development has prioritized support to the Pentaho Design System, so updates for NEXT DS are temporarily paused and are not fully reflected in the current **UI Kit** version.
```
zettca, line 9: Would it make more sense to use "deprioritized", as we eventually (indirectly) end up updating NEXT components when reviewing P+? We also mention it's a matter of priority on the start of the sentence
Also, **Pentaho+**
```suggestion
Recent development has prioritized support to the Pentaho+ Design System, so updates for NEXT DS are temporarily deprioritized and are not fully reflected in the current **UI Kit** version.
```

apps/docs/src/pages/docs/status.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+To get involved, [open an issue](https://github.com/lumada-design/hv-uikit-react/issues/new) or reach out through our channels.
+
+## Version Support
+
+Below is an overview of the supported **UI Kit** versions, along with their corresponding design system and core dependencies. We recommend using the **latest version** for all new development and production environments.
+
+| Release Date   | UI Kit Version                                              | Design System Version                                                                                     | React Version | MUI Version                    | Emotion Version |
+| -------------- | ----------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- | ------------- | ------------------------------ | --------------- |
+| March 2023     | [UI Kit v5.x](https://lumada-design.github.io/uikit/master) | [Design System 5.x](https://designsystem.hitachivantara.com/)                                             | 17, 18        | [MUI 5.x](https://mui.com/)    | 11.x            |
```
zettca, line 25: shouldn't we add DS3 and Pentaho+ to the DS Versions too? We support those

apps/docs/src/pages/docs/status.mdx
```diff
@@ -0,0 +1,39 @@
...
+| September 2022 | [UI Kit v4.x](https://lumada-design.github.io/uikit/v4.x)   | [Design System 3.x](https://hitachivantara.sharepoint.com/sites/DesignSystem/Pattern%20Library/Home.aspx) | 16, 17        | [MUI 5.x](https://mui.com/)    | 11.x            |
+| October 2020   | [UI Kit v3.x](https://lumada-design.github.io/uikit/v3.x)   | [Design System 3.x](https://hitachivantara.sharepoint.com/sites/DesignSystem/Pattern%20Library/Home.aspx) | 16, 17        | [MUI 4.x](https://v4.mui.com/) | —               |
+
+To learn more about our release cadence and versioning principles, see the [Release & Versioning](/docs/releases) documentation.
+
+## Browsers Support
+
+**UI Kit** supports the latest stable versions of all major modern browsers:
```
zettca, line 33: We can review this later, but we definitely won't be using features released on the latest versions of browsers (eg. May 2025)
We should probably add a mention to https://web.dev/baseline, which is what we essentially follow (through MDN & [caniuse](https://caniuse.com)).

apps/docs/src/pages/docs/_meta.ts
```diff
@@ -6,6 +6,7 @@ export default {
   introduction: "Introduction",
   "get-started": "Get Started",
   changelog: "Changelog",
+  status: "Status",
```
zettca, line 9: can we align the sidebar & page titles. We call it "Status" here, but "Project Status" on the page - it feels like something's missing

zettca (approved):

plagoa (approved):

apps/docs/src/pages/docs/status.mdx
francisco-guilherme: I'm not sure.
Lets do it later. We need to add Pentaho+ somewhere for sure 👍

PR #4708, opened 2025-05-08
fix(Select): invalid border color

MEsteves22:
- `HvSelect` didn't have a red border when invalid.

Before:

Screenshot 2025-05-08 at 12 24 50

After:

Screenshot 2025-05-08 at 12 31 47

### DIFF
--- a/packages/core/src/Select/Select.styles.ts
+++ b/packages/core/src/Select/Select.styles.ts
@@ -10,13 +10,13 @@ export const { staticClasses, useClasses } = createClasses("HvSelect", {
   },
   disabled: {},
   readOnly: {},
-  invalid: {
-    borderColor: theme.form.errorColor,
-  },
+  invalid: {},
   labelContainer: {},
   label: {},
   description: {},
-  select: {},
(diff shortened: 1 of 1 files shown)


packages/core/src/Select/Select.styles.ts
```diff
@@ -11,7 +13,8 @@ export const { staticClasses, useClasses } = createClasses("HvSelect", {
   disabled: {},
   readOnly: {},
   invalid: {
-    borderColor: theme.form.errorColor,
+    // Increase specificity to override the default button styles
+    [`&&.${buttonClasses.root}`]: { borderColor: theme.form.errorColor },
```
zettca, line 17: Can we instead leverage the `HvSelect`'s classes and move this to `classes.select`?

We could simply do this, without resorting to external component's classes:
```tsx
"&&$invalid": { borderColor: theme.form.errorColor },
```

MEsteves22: Updated

zettca (approved):

PR #4721, opened 2025-05-09
fix(TagsInput): add start adornment

plagoa:
- Add a start adornment to the `TagsInput` component (it will be needed in the future and we already support an end adornment
- Add a new example that illustrates how to keep the tags input empty while adding tags (based on the real use case scenario from PD)
- Small fixes to samples layout
- Revert inputs samples layout because the Examples index page counts the number of `<CodeBlock>` elements to tell the number of samples

### DIFF
--- a/apps/docs/src/examples/inputs/TagsSuggestions.tsx
+++ b/apps/docs/src/examples/inputs/TagsSuggestions.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useRef, useState } from "react";
-import { Popper } from "@mui/material";
+import { ClickAwayListener, Popper } from "@mui/material";
 import {
   HvAdornment,
   HvPanel,
@@ -66,7 +66,7 @@ export default function Demo() {
           handleRemoveColor(value as string);
         }}
         value={selectedColors}
-        commitTagOn={["Comma"]}
+        commitTagOn={["Comma", "Enter"]}
         endAdornment={
           <HvAdornment
             tabIndex={0}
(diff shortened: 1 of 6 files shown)


apps/docs/src/examples/inputs/TagsSuggestions.tsx
```diff
@@ -75,38 +75,45 @@ export default function Demo() {
           />
         }
         classes={{
-          tagsList: "h-32px pr-0 overflow-hidden",
+          tagsList: "h-32px pr-0! overflow-hidden",
```
zettca, line 78: this seems to do nothing
we should avoid using `!important` unless it's really necessary

apps/docs/src/examples/inputs/TagsSuggestionsEmpty.tsx
```diff
@@ -0,0 +1,106 @@
...
+    setSelectedColors((prev) => prev.filter((c) => c !== color));
+  };
+
+  return (
+    <>
+      <div className="w-300px">
+        <HvInput
+          label="Tags with suggestions and empty input"
+          onEnter={(event, value) => {
+            handleAddColor(value as string);
```
zettca, line 58:
```suggestion
            handleAddColor(value);
```

apps/docs/src/examples/inputs/TagsSuggestionsEmpty.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const handleAddColor = (color: string) => {
+    setSelectedColors((prev) => {
+      const newColor =
+        typeof color === "string"
+          ? color
+          : ((color as HvTagProps).label as string);
+      const colorExists = prev.includes(newColor);
+      return colorExists ? prev : [...prev, newColor];
+    });
```
zettca, line 45: What is this for? There is no `HvTagProps` - we're using an `HvInput`
```suggestion
    setSelectedColors((prev) => [...new Set([...prev, newColor])]);
```

apps/docs/src/examples/inputs/TagsSuggestionsEmpty.tsx
```diff
@@ -0,0 +1,106 @@
...
+        />
+
+        <Popper
+          anchorEl={containerRef.current}
+          open={open}
+          placement="bottom-start"
+          className="top-1px!"
+        >
+          <ClickAwayListener onClickAway={() => setOpen(false)}>
+            <HvPanel className="flex flex-wrap flex-col gap-xs w-300px border rounded-[var(--uikit-radii-large)]!">
```
zettca, line 78:
```suggestion
            <HvPanel className="grid gap-xs w-300px border rounded-large">
```

packages/core/src/TagsInput/TagsInput.tsx
```diff
@@ -536,6 +537,7 @@ export const HvTagsInput = forwardRef<HTMLElement, HvTagsInputProps>(
               />
             );
           })}
+          {!disabled && !readOnly && startAdornment}
```
zettca, line 540: let's add a test for this please

zettca (approved):

PR #4731, opened 2025-05-13
feat(Badge): add `color` prop

plagoa:


packages/core/src/Badge/Badge.stories.tsx
```diff
@@ -157,6 +158,22 @@ export const Test: StoryObj = {
...
+      <HvBadge variant="subtle" showCount label={888} icon={<Alert />} />
+      <HvBadge variant="subtle" count={0} text="Events" textVariant="label" />
+      <HvBadge variant="subtle" label={10} text="Events" />
+      <HvBadge
+        variant="subtle"
+        showCount
+        label={10}
+        maxCount={5}
+        text="Events"
+      />
```
zettca, line 173: There's no need to duplicate the whole API - especially the deprecated props
```suggestion
      <HvBadge variant="subtle" showCount label={88} icon={<Alert />} />
```

packages/core/src/Badge/Badge.tsx
```diff
@@ -12,7 +12,11 @@ export { staticClasses as badgeClasses };
 
 export type HvBadgeClasses = ExtractNames<typeof useClasses>;
 
+export type HvBadgeVariant = "primary" | "subtle";
+
 export interface HvBadgeProps extends HvBaseProps {
+  /** The badge variant. */
+  variant?: HvBadgeVariant;
```
zettca, line 19: if we're doing this to change the colors, then we should use the `color` prop like in other components, and add proper support in the component
```suggestion
  color?: HvColorAny;
```

plagoa: Updated to use `color` instead of `variant`.

packages/core/src/Badge/Badge.styles.tsx
```diff
@@ -25,7 +25,7 @@ export const { staticClasses, useClasses } = createClasses("HvBadge", {
     ...theme.typography.caption2,
     color: theme.colors.textDimmed,
     borderRadius: theme.radii.full,
-    backgroundColor: theme.colors.text,
+    backgroundColor: "var(--bg-color)",
```
zettca, line 28: Can we optimise this to fallback to `text`? This way we can remove the default value, and we get less code all around (no default prop, no CSS var, etc.)
```suggestion
    backgroundColor: `var(--bg-color, ${theme.colors.text})`,
```

apps/docs/src/pages/components/badge.mdx
```diff
@@ -17,6 +17,7 @@ export const getStaticProps = async ({ params }) => {
   Component={HvBadge}
   componentName="HvBadge"
   controls={{
+    color: { type: "color", defaultValue: "text" },
```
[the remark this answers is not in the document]
plagoa: It's the playground, why not let users play with it? :)

zettca (approved):

PR #4746, opened 2025-05-19
docs: add `color` prop to styling section

plagoa:


apps/docs/src/pages/docs/styling.mdx
```diff
@@ -46,6 +46,23 @@ buttonClasses.disabled; // "HvButton-disabled"
 buttonClasses.startIcon; // "HvButton-startIcon"
 ```
 
+### `color` prop
+
+Several **UI Kit** components have a [`color`](/docs/styling#color-prop) prop, which allows the user to override the default color of the component.
+The prop accepts any UI Kit or CSS color. These include UI Kit [color tokens](/docs/color-tokens) (`primary`, `positive`, `bgPageSecondary`, etc.)
+or CSS valid colors (`blue`, `#0000FF`, `rgb(0, 0, 255)`, `rgba(0, 0, 255, 0.5)`, etc.).
```
zettca, line 53: We could add a link to the MDN CSS colors; and since we're already linking both the docs, I'd say we don't need so many examples. `bgPageSecondary ` is a bit of an awkward token too
```suggestion
The prop accepts any UI Kit [color token](/docs/color-tokens) (`primary`, `positive`, etc.) or [CSS color](https://developer.mozilla.org/en-US/docs/Web/CSS/color) (`blue`, `#0000FF`, etc.).
```

zettca (approved):

apps/docs/src/pages/docs/styling.mdx
plagoa: Updated.

PR #4750, opened 2025-05-19
fix(Avatar): fix border radius styles

plagoa:
- fix the `variant` classes (`square`, `circular`) not being applied in the appropriate element and thus not having effect when users wanted to override the styles of each specific variant.
- Update to NEXT 5.9.5 specs

### DIFF
--- a/packages/core/src/Avatar/Avatar.styles.tsx
+++ b/packages/core/src/Avatar/Avatar.styles.tsx
@@ -26,6 +26,7 @@ export const { staticClasses, useClasses } = createClasses("HvAvatar", {
     alignItems: "center",
     position: "relative",
     padding: theme.space.xxs,
+
     height: "fit-content",
 
     ":focus-visible": {
@@ -38,7 +39,7 @@ export const { staticClasses, useClasses } = createClasses("HvAvatar", {
   md: { width: 40, height: 40, fontSize: "1rem" },
   lg: { width: 52, height: 52, fontSize: "1.5rem" },
   xl: { width: 88, height: 88, fontSize: "2rem" },
-  avatar: {},
+  avatar: { borderRadius: "inherit" },
   badge: {
     width: 8,
     height: 8,
--- a/packages/core/src/Avatar/Avatar.tsx
+++ b/packages/core/src/Avatar/Avatar.tsx
@@ -54,9 +54,9 @@ export interface HvAvatarProps extends HvBaseProps {
   /** A string representing the type of avatar to display, circular or square. */
   variant?: HvAvatarVariant;
   /** A string representing the color of the avatar border that represents its status. */
-  status?: string;
+  status?: HvColorAny;
   /** A string representing the color of the avatar badge. */
-  badge?: string;
+  badge?: HvColorAny;
   /** Attributes applied to the avatar element. */
   avatarProps?: MuiAvatarProps;
   /** A Jss Object used to override or extend the styles applied to the component. */
--- a/apps/docs/src/pages/components/avatar.mdx
+++ b/apps/docs/src/pages/components/avatar.mdx
@@ -32,6 +32,12 @@ export const getStaticProps = async ({ params }) => {
       type: "color",
       defaultValue: "secondary",
     },
+    status: {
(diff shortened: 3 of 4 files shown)


packages/core/src/Avatar/Avatar.styles.tsx
```diff
@@ -49,5 +49,5 @@ export const { staticClasses, useClasses } = createClasses("HvAvatar", {
     zIndex: 1,
   },
   circular: { borderRadius: theme.radii.full },
-  square: {},
+  square: { borderRadius: theme.radii.round },
```
zettca, line 52: this makes no sense to me... 😅

We could instead align with the MUI API and add the `rounded` variant. Our `variant` API is already a subset of MUI's, with only the `rounded` missing.
It probably would be better to align with `HvRadius`, but now it'd just add more confusion - maybe we can add a `// TODO: align with `HvRadius` if it makes sense 

Alternatively, this should be at least in the `ds5` overrides, not here in the "UI Kit default styles"

packages/core/src/Avatar/Avatar.tsx
```diff
@@ -168,7 +168,13 @@ export const HvAvatar = forwardRef<
       <MuiAvatar
         component={component}
         // Consider not using the root and className classes in this component
-        className={cx(classes.root, classes.avatar, classes[size], className)}
+        className={cx(
+          classes.root,
+          classes.avatar,
+          classes[size],
+          classes[variant],
```
zettca, line 175:
```suggestion
```

We shouldn't duplicate styles in _different_ components. Also, the "variants" should be in the root, so that we can use a selector from anywhere:

```ts
classes: {
  square: {
    "& .HvAvatar-avatar": {
      borderRadius: 0
    }
  }
}

**Comment by plagoa** [comment]
@zettca updated.

File: packages/core/src/Avatar/Avatar.styles.tsx [hunk 3]
```diff
@@ -48,6 +48,17 @@ export const { staticClasses, useClasses } = createClasses("HvAvatar", {
...
   },
-  circular: { borderRadius: theme.radii.full },
+  circular: {
+    borderRadius: theme.radii.full,
+    " $avatar": {
+      borderRadius: theme.radii.full,
+    },
+  },
   square: {},
+  rounded: {
```
zettca, line 58: where is this coming from? there isn't a rounded class being added

packages/core/src/Avatar/Avatar.styles.tsx
```diff
@@ -48,6 +48,17 @@ export const { staticClasses, useClasses } = createClasses("HvAvatar", {
     borderRadius: theme.radii.full,
     zIndex: 1,
   },
-  circular: { borderRadius: theme.radii.full },
+  circular: {
+    borderRadius: theme.radii.full,
+    " $avatar": {
+      borderRadius: theme.radii.full,
+    },
```
zettca, line 55: can't we just `borderRadius: "inherit"` on `classes.avatar` instead? it's a direct child, and saves us of these duplicated overrides

plagoa: Yep, much cleaner.

plagoa:
@zettca updated.

packages/core/src/Avatar/Avatar.styles.tsx
```diff
@@ -25,7 +25,7 @@ export const { staticClasses, useClasses } = createClasses("HvAvatar", {
     display: "flex",
     alignItems: "center",
     position: "relative",
-    padding: theme.space.xxs,
+    padding: 3,
```
zettca, line 28: where is this `3` coming from? we should avoid using values outside our spacing tokens
there's some miss-alignment now:
image

zettca: the spec says `4px`
image image

plagoa: Sorry, that's leftovers from a test I did. Fixed.

zettca (approved):

PR #4751, opened 2025-05-20
chore(docs): remove hitachivantara refences and improve README

francisco-guilherme:


zettca (approved):

plagoa:
The only thing here is that by removing any company references (other than the small "by Pentaho") we now are stuck with a generic **UI Kit** name. I'd still love to have an actual name for the library :)

plagoa (approved):

PR #4753, opened 2025-05-21
fix(Badge): fix P+ styles for primary badges

plagoa:


packages/styles/src/themes/pentahoPlus.ts
```diff
@@ -304,6 +304,9 @@ const pentahoPlus = makeTheme((theme) => ({
           "&[data-color='bgPageSecondary']": {
             color: theme.colors.text,
           },
+          "&[data-color='primaryDimmed']": {
+            color: theme.colors.primary,
+          },
```
zettca, line 309: Can we invert this?
```suggestion
          "&[data-color=primary]": {
            color: theme.colors.primary,
            backgroundColor: theme.colors.primaryDimmed,
          },
```

1. It's nicer for the user to just use `color="primary"` (we'd eventually do the same for other _semantic_ colors)
3. It fixes the color when there's no label/counter, which should be `primary`:

image

zettca (approved):

PR #4759, opened 2025-05-26
docs: add input example with label on the side

plagoa:


apps/docs/src/pages/components/form-element.mdx
```diff
@@ -90,19 +90,22 @@ along with the following building blocks provided by UI Kit:
...
-      endAdornment={
-        <>
-          <HvAdornment showWhen="invalid" icon={<Fail color="negative" />} />
-          <HvAdornment showWhen="valid" icon={<Success color="positive" />} />
-        </>
-      }
-    />
-  </HvLabel>
-  <HvWarningText disableBorder>This is the error message.</HvWarningText>
+  <div className="flex gap-xs">
```
zettca, line 93: `HvFormElement` accepts in a `className` - no need for another `div`!

plagoa: D'oh! fixed it, spot the `div`erence.

PR #4778, opened 2025-06-04
docs: include descriptions on examples that use the popup layout

plagoa:


francisco-guilherme:
Close in favor of #4809

PR #4783, opened 2025-06-16
fix(TagsInput): input styles

MEsteves22:
- I fixed some styling discrepancies for the Tags Input (when besides another input, you could see the Tags Input was different):
   - Border color for the Pentaho+ theme
   - Disabled background and border colors for the Pentaho+ theme
   - Placeholder text style

ℹ️ I think I found more stuff but I only fixed the styles: the placeholder disappears when read only or disabled is true; it looks like in Pentaho+ the description should be under the input.

### DIFF
--- a/packages/core/src/themes/pentahoPlus.ts
+++ b/packages/core/src/themes/pentahoPlus.ts
@@ -220,6 +220,9 @@ export const pentahoPlus = mergeTheme(pentahoPlusBase, {
               borderColor: theme.colors.textDisabled,
             },
           },
+          "& .HvTagsInput-chipRoot": {
+            outlineColor: theme.colors.textDisabled,
+          },
         },
       },
     },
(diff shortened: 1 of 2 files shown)


packages/core/src/themes/pentahoPlus.ts
```diff
@@ -208,10 +208,19 @@ export const pentahoPlus = mergeTheme(pentahoPlusBase, {
...
         singleLine: {
           height: 32,
         },
+        disabled: {
+          "& .HvTagsInput-tagsList": {
+            backgroundColor: theme.colors.bgDisabled,
+            "&,:hover": {
+              borderColor: theme.colors.textDisabled,
+            },
+          },
```
zettca, line 222: Can we also add the Tag border, so that it looks like the spec? 🙏 
```suggestion
          },
          "& .HvTagsInput-chipRoot": {
            outlineColor: theme.colors.textDisabled,
          },
```

PR #4794, opened 2025-07-02
fix(Banner,Snackbar): add transparency overlay to background

plagoa:


packages/core/src/themes/pentahoPlus.ts
```diff
@@ -64,7 +64,7 @@ export const pentahoPlus = mergeTheme(pentahoPlusBase, {
           Object.entries(notificationMap).map(([variant, color]) => [
             variant,
             {
-              backgroundColor: theme.colors[`${color}Dimmed`],
+              background: `linear-gradient(${theme.mix("dimmer", 0.5)}), ${theme.colors[`${color}Dimmed`]}`,
```
zettca, line 67: I tested, and the colors are equivalent to a simple `theme.mix` interpolation 😃 
```suggestion
              backgroundColor: theme.mix(`${color}Dimmed`, 0.5, "dimmer"),
```

zettca (approved):

PR #4798, opened 2025-07-03
feat(Dialog): update Dialog according to P+ specs

plagoa:


zettca (changes requested):
Can you please close & reopen this PR (with another branch name)?


There's something wrong with Chromatic (I enabled A11y, probably my fault).
There are too many changes (that are not actual changes), some of which are not OK (example below)
image



image

PR #4801, opened 2025-07-08
feat(Card): update Card to P+ 0.6.2 specs 

plagoa:
**Changes:**
- Card styles, specially the `selectable` card styles
- New card samples on the examples section


apps/docs/src/app/examples/cards/page.mdx
```diff
@@ -0,0 +1,18 @@
...
+import smallRadio from "./SmallRadio?raw";
+
+<div className="overflow-hidden mt-20">
+  <div className="w-101% grid  divide-border [grid-template-columns:repeat(auto-fit,minmax(480px,1fr))] -mt-px -mb-px [&>*]:p-20px">
+    <CodeBlock code={{ main: small }} layout="popup" />
+    <CodeBlock code={{ main: big }} layout="popup" />
+    <CodeBlock code={{ main: smallRadio }} layout="popup" />
+    <CodeBlock code={{ main: bigRadio }} layout="popup" />
+    <CodeBlock code={{ main: smallCheck }} layout="popup" />
+    <CodeBlock code={{ main: bigCheck }} layout="popup" />
```
zettca, line 16: The sample are too similar to each other - this is basically the same sample 6 times!

We're also doing selection, but each sample only has 1 Card, so we can't actually see the radio/checkboxes "working"... Can't we instead have 1-2 examples with multiple cards instead?

zettca (approved):

francisco-guilherme (approved):

PR #4803, opened 2025-07-10
fix(Snackbar,Banner): fix incorrect color on DS3 and DS5 themes

plagoa:


zettca:
see #4804, which also addresses the size (extra padding) issues

PR #4805, opened 2025-07-11
feat(Dialog): update Dialog according to P+ specs

plagoa:
Notable changes Dialog component in this PR:

- no more semantic bar on the top of the semantic dialogs
- use HvStatusIcon on the dialog title
- no more top border on the dialog actions bar
- updated border radius on the dialogs
- fixed border radius on the fullscreen dialog (some unwanted border radius was showing)
- changed order of buttons on sample stories to adhere to PD norms

### DIFF
--- a/packages/core/src/Dialog/Title/Title.tsx
+++ b/packages/core/src/Dialog/Title/Title.tsx
@@ -8,7 +8,6 @@ import {
 
 import { HvStatusIcon } from "../../StatusIcon";
 import { HvTypography } from "../../Typography";
-import { iconVariant } from "../../utils/iconVariant";
 import { useDialogContext } from "../context";
 import { staticClasses, useClasses } from "./Title.styles";
 
--- a/packages/core/src/Dialog/Title/Title.tsx
+++ b/packages/core/src/Dialog/Title/Title.tsx
@@ -6,6 +6,7 @@ import {
   type ExtractNames,
 } from "@hitachivantara/uikit-react-utils";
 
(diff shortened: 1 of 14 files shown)


packages/core/src/Dialog/Title/Title.tsx
```diff
@@ -48,7 +48,7 @@ export const HvDialogTitle = (props: HvDialogTitleProps) => {
   const { classes, cx } = useClasses(classesProp);
   const { fullScreen } = useDialogContext();
 
-  const icon = customIcon || (showIcon && iconVariant(variant));
+  const icon = customIcon || (showIcon && <HvStatusIcon variant={variant} />);
```
zettca, line 51: there are regressions in other themes, that Chromatic isn't picking up

image

plagoa: @zettca this should now be fixed

packages/core/src/themes/ds5.ts
```diff
@@ -151,6 +151,26 @@ export const ds5 = mergeTheme(ds5Base, {
...
+          "& .HvIconContainer-root": {
+            color: `${theme.colors.secondary}!important`,
+          },
+        },
+      },
+    },
+    HvDialogContent: {
+      classes: {
+        root: {
+          borderTop: "none!important",
```
zettca, line 169: we most likely don't need `!important` if we drop the redundant `classes.contentBorder`

zettca (approved):

PR #4822, opened 2025-07-16
chore: don't show close button when `onClose` isn't configured

plagoa:


packages/core/src/Banner/BannerContent/BannerContent.tsx
```diff
@@ -79,7 +79,7 @@ export const HvBannerContent = forwardRef<
       ref={ref}
       id={id}
       variant={variant}
-      showClose
+      showClose={onClose !== undefined}
```
zettca, line 82: without the icon, some Banners are becoming too small, and some actions are positioned incorrectly

PR #4836, opened 2025-07-18
feat(Flow): update reactflow to v12

plagoa:


packages/lab/src/Flow/base.ts
```diff
@@ -1,10 +1,115 @@
 import { css } from "@emotion/react";
-import { theme } from "@hitachivantara/uikit-react-core";
```
zettca, line 2: shouldn't we use our own tokens instead? eg. `theme.colors.*`, `theme.space.xs` instead of "10px", etc.

packages/lab/src/Flow/stories/Base/Asset.tsx
```diff
@@ -39,7 +40,7 @@ const classes = {
   }),
 };
 
-export const Asset: HvFlowNodeFC<AssetData> = (props) => {
+export const Asset: HvFlowNodeFC<RFNode<AssetData>> = (props) => {
```
zettca, line 43: Can't we keep the `HvFlowNodeFC<AssetData>` API? It'd go a long way in reducing breaking changes

PR #4862, opened 2025-08-08
docs: add ui kit tokens section

plagoa:


apps/docs/src/content/docs/ui-kit-tokens.mdx
```diff
@@ -0,0 +1,66 @@
...
+);
+```
+
+These tokens are categorized into several groups, each serving a specific purpose in the UI Kit design system and can be
+accessed through the `theme` object.
+
+When creating a new theme, you can extend any of the supported themes by overriding these tokens (or adding new ones).
+This allows you to customize the UI Kit to fit your brand or application needs while maintaining the core design principles.
+
+Below is a summary of the available token categories:
```
zettca, line 36: Can we add a `# Colors` and a link to it? 🙏

apps/docs/src/content/docs/_meta.ts
```diff
@@ -19,11 +19,15 @@ export default {
     type: "separator",
     title: "Foundation",
   },
-  theming: "Theming",
+  "ui-kit-tokens": {
+    title: "UI Kit Tokens",
```
zettca, line 23: can we call this "theme tokens"? at least in the URL, it looks weird/redundant

```suggestion
  "theme-tokens": {
    title: "Theme Tokens",
```

zettca (approved):
👌

PR #4915, opened 2025-09-17
feat(DropDownMenu): add support for item separators

nantunes:


francisco-guilherme (approved):
Great work 💪
The focus management already feels way less intrusive. When navigating with the keyboard it seems to skip to the next element, which looks a bit odd visually. But overall  it’s much more pleasant now, and we can refine it later.

https://github.com/user-attachments/assets/68aa1964-d364-4b27-9cc6-e5825e15fe72

PR #4930, opened 2025-10-02
chore: make apps be as-is on importmap [PPUC-199]

loonatic:
The `apps` properties keys are bing concatenated with a `/` which breaks some of the importmap supported [functionalities ](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap#scoped_module_specifier_maps) like `"square": "./modules/shapes/square.js"`.

Not sure if this is a fix or the unlocking of a feature, or maybe not really desirable which can be discussed.


dcleao (approved):
It's possible that `apps` was designed to be more limited than import maps' `imports` section on purpose -- and that its entries would always be folder mappings.

This change already provides a great more control of the generated import map.

Note that "we" will possibly need even more control, such as the ability to specify scoped mappings, via the import maps' `scopes` section (to support dev mode and `$self` keyword configuration per package/base url). In the end, maybe we'd be better off allowing for a full import map to be specified (in the app shell config or the vite plugin options?), which could then be merged with what gets built from the `apps` configuration. Or, these needs may be seen as specific to the Pentaho App Shell, and not to be addressed using the HV App Shell vite plugin...

francisco-guilherme (approved):

PR #4941, opened 2025-10-15
fix(AppShell): improve external marking and id prefix replacement [PPUC-199]

loonatic:
Because the plugin `vite:react-refresh` has an object for the `resolveId` hook, it was not executing the one on the `replaceIdPrefix` function and thus not marking the importmap elements as external. This PR addresses it by refactoring that function into 2: one to mark as external that forcibly runs before the mentioned plugin, and other to replace the prefixes that keeps the old behavior. What this was doing in reality is that, instead of using the files pointed by the importmap mapping, it was using the ones in the node_modules.

After that, we noticed that the addition of the packageName to the RegExp was failing to find matches and thus this PR reverts the change done [here](https://github.com/lumada-design/hv-app-shell/commit/76798cd8431667443097245c8f584d2e757aa55e) (and also this [fix](https://github.com/lumada-design/hv-app-shell/commit/1049c1263db95ed8623a94a1f44fb5fe2e60587b) later on).

### DIFF
--- a/packages/app-shell-vite-plugin/src/vite-importmap-plugin.ts
+++ b/packages/app-shell-vite-plugin/src/vite-importmap-plugin.ts
@@ -24,25 +24,13 @@ export const getExtraDependenciesString = (): string => {
   return extraDependenciesString;
 };
 
-function replaceIdPrefix(
-  keys: (string | RegExp)[],
-  packageName: string,
-): PluginOption {
-  const reg = new RegExp(
-    `(/${packageName})?${ID_PREFIX}(${keys.join("|")})`,
-    "g",
-  );
-
+function markExternal(keys: (string | RegExp)[]): PluginOption {
   return {
-    name: "vite-plugin-importmap-replace-idprefix",
+    name: "vite-plugin-importmap-mark-external",
     enforce: "pre",
     apply: "serve",
 
-    // while in dev, we need to prevent 'vite:import-analysis' on runtime modules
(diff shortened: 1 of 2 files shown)


dcleao (approved):

packages/app-shell-vite-plugin/src/vite-importmap-plugin.ts
```diff
@@ -180,8 +180,22 @@ document.currentScript.after(im);
...
+        // however, because vite:react-refresh overrides the `resolveId` hook,
+        // we need to add our `markExternal` before it
+        const index = resolvedConfig.plugins.findIndex(
+          (plugin) => plugin.name === "vite:react-refresh",
+        );
+
+        // if the plugin changes its name or is removed, this needs to be revised
+        if (index === -1) {
+          throw new Error("Could not find vite:react-refresh plugin");
+        }
```
nantunes: Why be so drastic? Couldn't you just add it to the end (push) if `vite:react-refresh` is not found?

loonatic: The thought behind it was that if `vite:react-refresh` ceases to exist, it will mostly likely be from a major version bump somewhere, which would make this code look odd and maybe it should be reviewed and refactored.
Neverthless, if you think it is still better to keep the old behavior as fallback, I can change, just let me know.

nantunes (approved):

PR #4974, opened 2025-11-11
test(storybook): use vitest runner

zettca:
replace jest with vitest for storybook tests #4793


zettca:
tests still hang for an unknown reason. we can try again after an eventual `storybook@10` + `vitest@4` upgrade, which stabilise browser mode tests

PR #4976, opened 2025-11-12
docs: change docs repo to static

zettca:
- set docs repo to `lumada-design/lumada-design.github.io`, while the Pentaho org one isn't ready/available


zettca:
it's now ready https://pentaho.github.io/uikit-docs/master/

PR #4985, opened 2025-11-17
feat(Banner): update to PD 0.9 specs

plagoa:
@zettca @francisco-guilherme one thing regarding this PR and concerning this new variant of the banner that PD added to their specs:

image

I first added it as a variant but I ended up removing it in favor of just having a sample illustrating how to achieve the desired effect (it's a very simple style override). I don't think having yet another variant for such a specific use case and bloating our API would be a good approach. 

In the future, if indeed we find that there's plenty use of this variant, we might see if it makes sense to add it. For now, I think this is enough.

### DIFF
--- a/packages/core/src/utils/Callout.tsx
+++ b/packages/core/src/utils/Callout.tsx
@@ -23,7 +23,7 @@ const { useClasses } = createClasses("HvCallout", {
     boxShadow: "none",
     flexWrap: "nowrap",
     borderRadius: theme.radii.round,
-    alignItems: "flex-start",
+    alignItems: "center",
 
     "&[data-size='large']": {
       padding: theme.space.sm,
--- a/packages/core/src/Snackbar/Snackbar.stories.tsx
+++ b/packages/core/src/Snackbar/Snackbar.stories.tsx
@@ -1,6 +1,5 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
 import {
-  HvIconContainer,
   HvOverflowTooltip,
   HvSnackbar,
   HvSnackbarContent,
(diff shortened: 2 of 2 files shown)


packages/core/src/utils/Callout.tsx
```diff
@@ -176,23 +191,25 @@ export const HvCallout = forwardRef<
       ref={ref}
       id={id}
       classes={{
-        root: cx(classes.root, classes[variant], className),
+        root: cx(classes.root, classes[variant], classes[size], className),
```
zettca: let's use a `data-size` class instead, like we do elsewhere

packages/core/src/themes/pentahoPlus.ts
```diff
@@ -145,6 +146,9 @@ export const pentahoPlus = mergeTheme(pentahoPlusBase, {
         actionClose: {
           color: theme.colors.text,
         },
+        content: {
+          ...theme.typography.caption1,
```
zettca: this is too small for "normal" text...

packages/core/src/utils/Callout.tsx
```diff
@@ -176,23 +191,25 @@ export const HvCallout = forwardRef<
...
+              size={size === "large" ? "md" : "sm"}
               className={classes.messageIcon}
               variant={variant === "default" ? "info" : variant}
               customIcon={customIcon}
             />
           )}
           <div className={classes.messageContent}>
             {title && <b className={classes.messageTitle}>{title}</b>}
-            {children}
+            <div className={classes.content}>{children}</div>
```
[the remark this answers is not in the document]
plagoa: I need it to apply some styles to that part of the message content only and just for the `pentaho` theme.

packages/core/src/themes/pentahoPlus.ts
zettca: please remove, this is no longer the case

plagoa: Updated this as per the latest PD specs

PR #5014, opened 2025-12-11
fix: prevent preload of external modules

loonatic:


package-lock.json
```diff
@@ -296,7 +296,6 @@
       "integrity": "sha512-UlLAnTPrFdNGoFtbSXwcGFQBtQZJCNjaN6hQNP3UPvuNXT1i82N26KL3dZeIpNalWywr9IuQuncaAfUaS1g6sQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
```
zettca: please revert this. these keys are added by `npm@11.6+`, which we need to use

feel free to run `corepack use npm@11.6.2`, which ensure (if you're using `corepack`) that the correct version runs

PR #5015, opened 2025-12-11
fix(AppShell): prevent preload of shared external modules

zettca:
- fixes https://github.com/pentaho/hv-uikit-react/pull/5014 in dev mode
  - only replaces the known `sharedDependencies`
- upgrades corepack npm version to `npm@11.6.2`
  - we require this version publishing, and older versions have `package-lock.json` differences

### DIFF
--- a/packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
+++ b/packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
@@ -5,6 +5,7 @@ import type { PluginOption, UserConfig } from "vite";
 import type { HvAppShellConfig } from "@hitachivantara/app-shell-shared";
 
 import { getAppModules, getBasePath } from "./config-utils.js";
+import sharedDependencies from "./shared-dependencies.js";
 
 /**
  * Process configuration, executing several tasks:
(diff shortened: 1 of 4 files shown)


loonatic (approved):

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
```diff
@@ -162,3 +155,24 @@ export default function processConfiguration(
...
+  const sharedDeps = sharedDependencies
+    .map((dep) => dep.moduleId)
+    // escape the dependencies
+    .map((dep) => dep.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"))
+    .join("|");
+
+  const sharedDepsRegex = new RegExp(
+    `<script type="module"[^>]+src="(${sharedDeps})"[^>]*><\\/script>`,
+    "g",
+  );
```
dcleao: Don't understand the trouble to only catch shared dependencies.
The previous code had it right, in that regard, imo.
The syntax for a script tag whose `src` is a bare module identifier is not supported at all and is used only as hack for an older browser that we do not support.

zettca: Well, the previous code wasn't working on dev mode because it was removing more than it should. I believe the culprit was a `"/@vite/client"`, which _looks like_ both a path & "bare" specifier at the same time.
And so I opted for a more conservative approach.

PR #5039, opened 2026-01-08
fix(StatusIcon): remove extra padding when in simple mode

plagoa:
According to the [specs](https://www.figma.com/design/aRqtbvWWrPf7asrEbd3g8x/branch/X3WOsVc7h7xG8tSu4JDrlI/Pentaho--Design-System-v0.9?node-id=25257-77961&m=dev), when `type="simple"` the icon shouldn't have the extra padding that it has when in `full` mode, only the padding from the underlying icon container.


plagoa:
@zettca take a look now.

PR #5071, opened 2026-01-27
fix(TimeAgo): fix yesterday timestamps showing as today due to incorrect date comparison calculation. CAT-10453

francisco-guilherme:
Fix flawed comparison logic that compared seconds elapsed in timestamp's day
with time difference, causing incorrect "today"/"yesterday" classification.
Replaced with proper calendar day boundary comparisons.

- Replace seconds-based comparison with calendar day boundary checks
- Add formatUtils.test.ts with comprehensive test coverage 
- Improve code readability and structure

### DIFF
--- a/packages/core/src/TimeAgo/TimeAgo.test.tsx
+++ b/packages/core/src/TimeAgo/TimeAgo.test.tsx
@@ -37,6 +37,17 @@ describe("TimeAgo with timestamp", () => {
     expect(screen.getByText(/^yesterday/)).toBeVisible();
   });
 
+  it("should show yesterday for timestamp less than a day ago from previous day", () => {
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
+    yesterday.setHours(23, 58, 0, 0); // Late night timestamp
+    const timestamp = yesterday.getTime();
+
+    render(<HvTimeAgo timestamp={timestamp} locale="en" />);
+
+    expect(screen.getByText(/^yesterday/i)).toBeVisible();
+  });
+
   it("should contain the relative time when the day is tomorrow", () => {
     const timestamp = new Date().setDate(new Date().getDate() + 1);
     render(<HvTimeAgo timestamp={timestamp} />);
--- a/packages/core/src/TimeAgo/formatUtils.test.ts
+++ b/packages/core/src/TimeAgo/formatUtils.test.ts
@@ -1,0 +1,142 @@
+import { describe, expect, it } from "vitest";
+
+import { formatTimeAgo } from "./formatUtils";
+
+describe("formatTimeAgo", () => {
+  const referenceDate = new Date("2026-01-15T12:00:00.000Z");
+
+  describe("Past dates", () => {
+    it("should show 'now' for timestamps less than 20 seconds ago", () => {
+      const timestamp = new Date(referenceDate.getTime() - 10 * 1000).getTime();
+      const result = formatTimeAgo(timestamp, "en", false, referenceDate);
+      expect(result).toBe("now");
+    });
+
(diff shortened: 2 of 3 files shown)


packages/core/src/TimeAgo/formatUtils.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      timestamp.setDate(timestamp.getDate() - 3); // 3 days ago
+      timestamp.setHours(10, 0, 0, 0);
+      const result = formatTimeAgo(
+        timestamp.getTime(),
+        "en",
+        false,
+        referenceDate,
+      );
+      // Should show weekday format (e.g., "Mon, 10:00 AM")
+      expect(result).toMatch(/^(Mon|Tue|Wed|Thu|Fri|Sat|Sun)/i);
```
zettca: this seems too vague too

packages/core/src/TimeAgo/formatUtils.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    it("should show 'today' for timestamps earlier today", () => {
+      const timestamp = new Date(referenceDate);
+      timestamp.setHours(8, 0, 0, 0); // 8 AM same day
+      const result = formatTimeAgo(
+        timestamp.getTime(),
+        "en",
+        false,
+        referenceDate,
+      );
+      expect(result).toMatch(/today/i);
```
zettca: what about the the time?

packages/core/src/TimeAgo/formatUtils.ts
```diff
@@ -35,53 +12,93 @@ export const formatTimeAgo = (
...
+
+    // Future: < 2 minutes ahead
+    case secsAgo < 0:
+      return relFormatter.format(Math.abs(secsAgo), "seconds");
+
+    // Just now (< 20 seconds ago)
     case secsAgo < 20:
       return relFormatter.format(0, "seconds");
+
+    // < 2 minutes ago
```
zettca: do these really need an explanation? ...

packages/core/src/TimeAgo/TimeAgo.test.tsx
```diff
@@ -37,6 +37,17 @@ describe("TimeAgo with timestamp", () => {
...
 
+  it("should show yesterday for timestamp less than a day ago from previous day", () => {
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
+    yesterday.setHours(23, 58, 0, 0); // Late night timestamp
+    const timestamp = yesterday.getTime();
+
+    render(<HvTimeAgo timestamp={timestamp} locale="en" />);
+
+    expect(screen.getByText(/^yesterday/i)).toBeVisible();
```
zettca: what about the time?

packages/core/src/TimeAgo/formatUtils.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      const result = formatTimeAgo(timestamp, "en", false, referenceDate);
+      expect(result).toMatch(/\d+\s*(second|sec)/i);
+    });
+
+    it("should show minutes ago for timestamps 2-60 minutes ago", () => {
+      const timestamp = new Date(
+        referenceDate.getTime() - 30 * 60 * 1000,
+      ).getTime();
+      const result = formatTimeAgo(timestamp, "en", false, referenceDate);
+      expect(result).toMatch(/\d+\s*(minute|min)/i);
```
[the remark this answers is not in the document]
francisco-guilherme: Good point! The regex was a bit overkill — since our tests always use the same locale, we can just check for the exact strings instead of a loose pattern. I’ll update it.

PR #5081, opened 2026-02-06
fix(DialogContent): dividers & backgroundColor

zettca:
- fix `backgroundImage` in dark mode
- improve dividers/showBorder feature
  - built on top of MUI's `dividers` prop that does the same, and follow it if passed
  - `useResizeDetector` for performance


zettca:
> @zettca maybe we can address the background color on the content section? I don't really understand why it's not working on dark mode, but a `backgroundColor: "inherit"` there should fix it.

It's because of this property coming from MUI:
image

My bad, the `backgroundImage` unset should be in `HvDialog`, not `HvDialogContent`. And just `background` will override it correctly. Updated
IIRC, we want `HvDialogContent` to be transparent (like it is now), and not the parent's color.

PR #5145, opened 2026-04-17
feat(Theme): update shadows on the Pentaho theme to the latest specs

plagoa:
- Update default shadow definition on the Pentaho theme to the latest spec
- Add custom overrides to the `HvDialog` and `HvTooltip` components
- Add shadows to components that didn't have shadow before: `HvCard` and `HvSnackbar`

### DIFF
--- a/packages/core/src/themes/pentaho.ts
+++ b/packages/core/src/themes/pentaho.ts
@@ -130,7 +130,7 @@ const popperStyles = {
   backgroundColor: theme.colors.bgContainer,
   border: `1px solid ${theme.colors.borderSubtle}`,
   borderRadius: theme.radii.large,
-  boxShadow: `0px 0px 6px 0px rgba(65, 65, 65, 0.08)`,
+  boxShadow: theme.colors.shadow,
 };
 
 export const pentaho = mergeTheme(pentahoBase, {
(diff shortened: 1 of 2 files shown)


francisco-guilherme (approved):

packages/core/src/themes/pentaho.ts
```diff
@@ -672,6 +672,7 @@ export const pentaho = mergeTheme(pentahoBase, {
       classes: {
         paper: {
           borderRadius: theme.radii.large,
+          boxShadow: `0 32px 32px 0 ${theme.alpha(slate[900], 0.1)}, 0 20px 20px 0 ${theme.alpha(slate[900], 0.08)}, 0 12px 12px 0 ${theme.alpha(slate[900], 0.06)}, 0 5px 5px 0 ${theme.alpha(slate[900], 0.04)}, 0 1px 1px 0 ${theme.alpha(slate[900], 0.02)}`,
```
zettca: can we promote this to a variable, like we do in other "component"/internal tokens?
so it's easier to understand the PD mapping & its reusable?

ie

```ts
const shadows = {
  container: theme.colors.shadow, // we can also omit this
  elevated: `0 32....`
  modal: `0 ...`
}

File: packages/core/src/themes/pentaho.ts [hunk 2]
```diff
@@ -851,6 +852,7 @@ export const pentaho = mergeTheme(pentahoBase, {
       classes: {
         root: {
           outlineColor: theme.colors.borderSubtle,
+          boxShadow: theme.colors.shadow,
```
zettca, line 861: this means cards will always be raised 🤔

zettca (approved):

packages/core/src/themes/pentaho.ts
plagoa: I'll make this change 👍

PR #5154, opened 2026-04-27
App Shell i18n improvements [PPUC-512]

dcleao:
This pull request improves App Shell's internationalization (i18n), moving from inline translations only to supporting a file-based locale structure, and adds new documentation and a helper package for `i18next` library integration.

**New Internationalization Features**

* Inline `translations` from `app-shell.config.json` now represent "preloaded" translations, when present, but which are refreshed by the translation resources in a file-based locale structure under `/locales/{lng}/app.json` for each supported language, once fetched.
* The translation resources for the App Shell's own UI/chrome are preloaded (for the "en" language), like before, and refreshed from the translations repository under `/locales/{lng}/appShell.json`.
* Added a manifest file `supported-locales.json` to declare supported locales, avoiding 404 network requests. The Vite plugin creates the file if missing, or else keeps it in sync with the repository locale folders.
* Added a new hook, `useHvAppShellI18n` hook, to `@hitachivantara/app-shell-shared` which allows reading and changing the App Shell's current language, in an i18n library agnostic way.
* Added translations for all Pentaho BA's stock languages.

**Documentation Updates**

* Added a new `internationalization.md` guide explaining the new translations file layout, migration steps, namespaces, and advanced usage scenarios.
* Updated configuration and API reference docs to reflect the new i18n system
* Added navigation entry for Internationalization in the docs sidebar.

Check out the added `internationalization.md` guide for a more complete overview of the I18n features!

**New i18next Integration Package**

* Introduced the `@hitachivantara/app-shell-i18next` package, providing a utility for conveniently, safely initializing `i18next` (`createI18n` and `useI18n`), and a backend, `HttpResourcesBackend`, that supports the new translations repository format. Used by the App Shell UI itself, and potentially by application bundles using `i18next` following the same translations repository format.

The code for this library was prototyped in the context of https://github.com/pentaho/pentaho-app-shell/pull/1134.

Main features:
- create i18n instances with sensible defaults (`createI18n`);
- initialize those instances from React components with predictable loading and
  error handling (`useI18n`); and
- reactive current language update via (`useI18n({lng})`)
- load translation bundles from HTTP using the new resource repository format (`HttpResourcesBackend`)

Check out the package's `README.md` and the `createI18n` doclet for more information!

**Minor: compilation changes to `app-shell-shared`**

**TODO**: check if this is still useful, now that, with the latest revision of this PR, `app-shell-shared` is back to referencing `i18next` types only.

The `@hitachivantara/app-shell-shared` package was _pre-built_ by its own package and its `.esm.js` bundle was _statically copied_ to the output by the Vite plugin.  The `@hitachivantara/app-shell-ui` package, was, however, _compiled from source_ during the Vite plugin build (via the virtual `/virtual/main.tsx` entry point).

Because they lived in separate compilations, they couldn't seamlessly share code such as `i18next` and `react-i18next`. Hence why `app-shell-shared` would only depending on `i18next` types, not implementations, as two different bundled modules would result, which would be incompatible with each other.

To solve this, introduced the concept of _built-from-source shared dependencies_ in `shared-dependencies.ts` via a `bundleSrc: null` convention. When `bundleSrc` is `null`, the dependency is compiled from source as part of the App Shell build instead of being statically copied from a pre-built file.

The result is that now `app-shell-shared` and `app-shell-ui` are built together in the same Vite compilation and will share modules such as `i18next` and `react-i18next` as required by other features in this PR.


packages/app-shell-ui/package.json
```diff
@@ -48,6 +49,11 @@
   "files": [
     "dist"
   ],
+  "exports": {
+    ".": "./src/index.ts",
+    "./package.json": "./package.json",
+    "./locales/*": "./dist/locales/*"
```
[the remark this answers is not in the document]
dcleao: Definitely a bug. Will fix.

packages/app-shell-shared/src/i18n/index.ts
```diff
@@ -1 +1 @@
-export const CONFIG_TRANSLATIONS_NAMESPACE = "configTranslations";
+export const CONFIG_TRANSLATIONS_NAMESPACE = "app";
```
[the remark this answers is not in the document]
dcleao: While publicly exported, this is (was) not a public API of the App Shell, it is used internally by App Shell packages (shared, ui, etc.). So, not a real compatibility breakage.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
```diff
@@ -1,70 +1,141 @@
...
+      configUrl,
+      config.baseUrl,
+    );
 
-  const config = useMemo(
-    () => configProp ?? loadedConfig,
-    [configProp, loadedConfig],
-  );
+    initializeI18Next(i18n, resolvedBaseUrl);
+  }, [config, configUrl]);
```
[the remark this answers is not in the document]
dcleao: This is not an issue in this project, imo.

This project does not use concurrent rendering.

The reason this is using `useMemo` and not `useEffect` is because of the former's _synchronous_ initialization function, while still being guarded by `config` and `configUrl` having changed. So, `useEffect` or `useLayoutEffect` are not real alternatives.

The `initalizeI18Next` is already idempotent, so replays are not really an issue. The use of `useMemo` is mostly one of performance.

The "proper" alternative seems way more complex, via use of `useRef`, synchronous/inline initialization guarded by explicit comparison.

packages/app-shell-ui/src/i18n/index.ts
```diff
@@ -19,24 +22,94 @@ export const addResourceBundles = (
...
+        baseUrl: resolvedBaseUrl,
+      },
+    })
+    .catch(() => {});
 
-  return newInstance;
+  // The backend connector skips fetching resources already in the store
+  // (from `resources` above). Force a reload so the server-side
+  // translations override the pre-loaded ones when the fetch completes.
+  instance.reloadResources().catch(() => {});
```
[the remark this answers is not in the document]
dcleao: The race is intentional. i18next already internally handles concurrent reloads. In any case, I will update to limit the reload to the preloaded bundles (which is exactly what would happen, but this will make it clearer).

packages/app-shell-ui/src/locales/supported-locales.json
```diff
@@ -0,0 +1 @@
+["en", "pt"]
```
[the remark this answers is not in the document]
dcleao: Well spotted. Will fix. Despite the Vite plugin already fixing this for those using it.

packages/app-shell-i18next/src/HttpResourcesBackend.ts
```diff
@@ -0,0 +1,221 @@
...
+  const bundleUrl = getBundleUrl(baseUrl, lng, ns);
+  const res = await fetch(bundleUrl, fetchOptions);
+  if (!res.ok) {
+    throw new Error(
+      `Failed to load translation bundle ${bundleUrl}: HTTP ${res.status}`,
+    );
+  }
+
+  return await res.json();
+};
```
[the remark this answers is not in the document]
dcleao: This is what's expected from an i18next backend implementation.

packages/app-shell-i18next/src/HttpResourcesBackend.ts
```diff
@@ -0,0 +1,221 @@
...
+  BackendModule,
+  CallbackError,
+  ReadCallback,
+  ResourceKey,
+  Services,
+} from "i18next";
+
+import type { FetchOptions, HttpResourcesBackendOptions } from "./types";
+
+// region async-to-callback bridge
```
zettca: not a big deal, but we're missing an `#` in `#region` to have IDE integration (at least in VSCode)
```suggestion
// #region async-to-callback bridge
```

packages/app-shell-i18next/src/types.ts
```diff
@@ -0,0 +1,73 @@
...
+// instances created by createI18n (with opinionated defaults and deferred init).
+declare const __brand: unique symbol;
+
+/** A branded i18next instance returned by {@link createI18n}. */
+export type I18nInstance = i18n & { readonly [__brand]: true };
+
+// endregion
+
+// region useI18n types
+export interface UseI18nResult {
```
zettca: this should live alongside the function it's declaring the types for

```tsx
import {useI18n, UseI18nResult} from "./useI18n"
```

`/types.ts` should be "reserved" for types that have no "home"

packages/app-shell-ui/src/components/layout/Header/Header.tsx
```diff
@@ -20,7 +20,10 @@ import { BrandLogo } from "../BrandLogo/BrandLogo";
 import HeaderActions from "../HeaderActions/HeaderActions";
 
 export const Header = () => {
-  const { t } = useTranslation(undefined, { keyPrefix: "header.navigation" });
+  const { useTranslation } = useHvAppShellRuntimeContext();
+  const { t } = useTranslation(undefined, {
+    keyPrefix: "header.navigation",
+  });
```
zettca: does this mean that the previous strategy no longer works?

should apps also do this?

packages/app-shell-ui/src/locales/zh-TW/appShell.json
```diff
@@ -0,0 +1,48 @@
+{
+  "errors": {
+    "missing": {
+      "configuration": "設定遺失或不完整。"
```
zettca, line 4: what is `cn`? it's not a [valid ISO language code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes#Table_of_all_possible_two-letter_codes)

packages/app-shell-ui/vite.config.ts
```diff
@@ -1,5 +1,20 @@
...
+    }),
+  ],
+  build: {
+    rollupOptions: {
+      // Exclude locale JSON files from the module graph.
+      // They are copied to dist/locales/ by viteStaticCopy and
+      // only imported statically for the pre-bundled en fallback.
+      external: [/\/locales\/.*\.json$/],
+    },
+  },
```
zettca, line 19: @copilot could this be refactored to use the built-in `lib.entry: ["src/index.ts", src/locales/*.json"]`, or does it not support wildcards or `.json` entrypoints?

package.json
```diff
@@ -80,6 +80,7 @@
     "encoding": "^0.1.13",
     "happy-dom": "^20.0.0",
     "husky": "^8.0.3",
+    "i18next": "^24.2.2",
```
zettca: this should live under the packages that need it, not in the root

package.json
```diff
@@ -89,6 +90,7 @@
     "pretty-quick": "^4.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-i18next": "15.7.1",
```
zettca: this should live under the packages that need it, not in the root

packages/app-shell-shared/src/i18n/index.ts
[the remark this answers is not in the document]
zettca: this is the shared package, so everything's more "sensitive"
I don't see any problems with it though (famous last words)

packages/app-shell-shared/src/types/Config.ts
```diff
@@ -89,6 +89,27 @@ export interface HvAppShellConfig {
...
+   *
+   * Can coexist with `translations`, which provides pre-loaded resources.
+   * When both are provided, `translations` are shown immediately and then
+   * refreshed when the HTTP backend responds.
+   *
+   * Set to `false` to disable HTTP-based translation loading entirely.
+   * Useful when only inline `translations` are used and the extra network
+   * requests would cause 404 errors in legacy setups.
+   */
+  translationsBaseUrl?: string | false;
```
zettca, line 112: do we have a use-case for this? can't we just reuse the case from `baseUrl` for now, and add something ?

if we don't really need it, I suggest leaving it for later, once the need arises, because:
- this is a new addition to `shared`
- we're assuming the `/locales/` in the path, which might not always be the case?

apps/docs/src/content/app-shell/internationalization.md
```diff
@@ -0,0 +1,219 @@
...
+
+```ts
+// app-shell.config.ts
+export default {
+  name: "appName",
+  menu: [{ label: "pageOne", target: "/page1" }],
+} satisfies HvAppShellConfig;
+```
+
+```json
```
zettca:
```suggestion
```jsonc
```

otherwise we'll get those big red invalid lines, as `json` doesn't support comments

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
```diff
@@ -1,10 +1,107 @@
...
+ * The custom App Shell ultimately compiling the code should not use the root I18next provider.
+ * It may not be configured as needed by external code. It can however, have its own I18next instance, which can/should
+ * be configured for the same resource bundle repository (of `./locales/{{lng}}/{{ns}}.json`), rooted at the configured
+ * translations base path via {@link HvAppShellConfig#translationsBaseUrl}, albeit consuming other namespaces.
+ * This repository is required when providing translations outside the App Shell configuration (i.e., not embedded in
+ * the configuration via {@link HvAppShellConfig#translations}).
+ */
 export interface HvAppShellRuntimeContextValue {
   i18n: i18n;
+  useTranslation: UseTranslation;
```
zettca: why do we need this? can't we infer/extract `useTranslation` from the `i18n` instance we already have?

If we want a `useTranslation` for better DX, we can `const {i18n} = useContext(HvAppShellRuntimeContext)` and create it elsewhere (not in the `shared` package), and export it from there. `shared` isn't for utilities

apps/docs/src/content/app-shell/api-reference.md
```diff
@@ -91,6 +91,44 @@ const customEvent = new CustomEvent<HvAppShellEventTheme>()(
...
+
+| Property         | Type                                 | Description                                                                                                     |
+| ---------------- | ------------------------------------ | --------------------------------------------------------------------------------------------------------------- |
+| `language`       | `string`                             | The current language of the App Shell, as a BCP 47 language tag.                                                |
+| `changeLanguage` | `(newLng?: string) => Promise<void>` | Changes the App Shell's current language. Triggers re-rendering of all components using App Shell translations. |
+
+**Example**:
+
+```tsx
+import { useHvAppShellI18n } from "@hitachivantara/app-shell-shared";
```
zettca, line 112: users shouldn't be importing from `shared` directly

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
```diff
@@ -1,10 +1,107 @@
...
+      );
+    }
+
+    return context;
+  };
+
+/**
+ * Gets the App Shell's I18N context.
+ */
+export const useHvAppShellI18n = (): HvUseAppShellI18nContextValue => {
```
zettca: same as above - `shared` isn't for utilities and users shouldn't be using it directly

packages/app-shell-ui/vite.config.ts
zettca: @copilot

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
```diff
@@ -1,70 +1,148 @@
...
+        .then((r) => r.json() as Promise<HvAppShellConfig>)
+        .then(
+          (data) => {
+            req.status = "fulfilled";
+            req.result = data;
+          },
+          (err) => {
+            req.status = "rejected";
+            req.error = err;
+          },
```
zettca: why isn't this a `.catch()` like previously? it would catch more errors, not just `.json()` parsing

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
```diff
@@ -1,70 +1,148 @@
...
+    configProp as HvAppShellConfig | undefined,
+    configUrl,
+  );
 
-  // Load config from URL
-  useEffect(() => {
-    if (configProp || !configUrl) return;
+  // Initialize i18next synchronously before anything below renders.
+  // `initializeI18Next` is idempotent — safe to call on every render.
+  useMemo(() => {
```
zettca: this is a code smell. `useMemo()` is for memoizing values.
`initializeI18Next` is a side-effect, and therefore we should use `useEffect`

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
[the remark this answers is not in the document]
zettca: if `initalizeI18Next`'s replays aren't an issue, we can do something like this

```tsx
if (config) {
  initalizeI18Next(i18n, resolvedBaseUrl)
}
```
in the root of the component

packages/app-shell-ui/src/components/layout/Header/Header.tsx
dcleao: By previous strategy, meaning: `const { t } = useTranslation(undefined, { keyPrefix: "header.navigation" });`.

It might still work. IMO, the important part is that the previous code / the previous way had a bug --- it was vulnerable to a system provider changing the i18next instance that this `useTranslation` would find. The new code always uses the proper i18next instance. Hence why this code is in a `fix` commit.

As for apps, apps do not have access to the same i18next module used by the App Shell "internal" packages. To use i18next, they must bundle it locally and install their own i18next instance.

Only the code of the custom "App Shell" compiled with the vite plugin has access to the same i18next modules and instance. This is why the documentation clearly states that custom app shell code **must** install their own local instance (despite the base code already installing one).

Moreover, the App Shell runtime context always was a private thing for internal sharing amongst "internal" app shell packages.

So, should apps also do this? Normal apps have their own i18next instance and may do it how they please.
As per the custom App Shell code, it already should be installing their own i18next instance, with its own configuration.

packages/app-shell-ui/src/locales/zh-TW/appShell.json
dcleao: There's a note in the corresponding commit justifying this:
> - Note that Pentaho BA for some unknown reason incorrectly uses what appears to be the incorrect language code, "cn", for Traditional Chinese, instead of the correct BCP 47 code, "zh-TW"

packages/app-shell-shared/src/types/Config.ts
dcleao: > The current, immediate use case is Pentaho BA and the default, `baseUrl` is what is needed. Things aligned perfectly.
> 
> Still, in other systems, given this will be "commonly" implemented as a REST endpoint (as it was in Pentaho BA), it may not be possibly to satisfy the request directly under `{baseUrl}/locales`.
> 
> I am not sure this hurts as a feature. But I am ok to remove it for now, if you think it's unwarranted "generality".
> 
> It accumulates, however, the ability to switch off HTTP fetching of resource bundles. Any suggestions for the name of a new config property that would disable this, for "extra" backward compatibility? 
> 
> Something like (the admmitably uninspired):
> 
> `translationsHttpDisabled: [false]` 

Ignore my previous statement.

Pentaho BA is actually using a different URL for the translations (`translationsBaseUrl`) than that used for static resources (`baseUrl`). The reason is that translations are provided by a REST endpoint that computes consolidated translations from the `/locales` repository of each Pentaho BA plugin. In Pentaho BA, the base URLs for REST endpoints are typically different from those of static resources.
- `<base href="/pentaho/api/repos/app-shell/webclient/">`
- `{ "translationsBaseUrl": "/pentaho/plugin/app-shell/" }`  (`/locales` ends up appended here )

The new locales endpoint is introduced in the associated PR, https://github.com/pentaho/pentaho-app-shell/pull/1235.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: We need `useTranslation` for the reactiveness in face of language changes.
There is no way to get this from an `i18n` instance.

There are two features here:
1. use of `HvAppShellRuntimeContextValue#useTranslation` by "internal" App Shell packages; this is currently i18next agnostic (only references i18next types); we could extract the entire runtime context to an "internal" runtime context package so that i18next could be explicitly declared and there would be no need to expose `useTranslation` this way.
2. use of the new hook `useHvAppShellI18n` by any app package; this should be i18next agnostic

apps/docs/src/content/app-shell/api-reference.md
dcleao: I am not sure this is what was concluded from yesterday's discussion.

Are you proposing something like `@hitachivantara/app-shell-i18n` that app packages would depend on to access this?

Note that, even with that approach, app packages would likely still need to declare a dependency on the shared package, if nothing else, for type declarations. There's also some hooks in there such as `useHvAppShellConfig`, `useHvAppShellModel`, for (arguably) advanced app cases. We can agree to see `useHvMenuItems` in here as unwanted and legacy (still looking for some solution, though).

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: Same as https://github.com/pentaho/hv-uikit-react/pull/5154#discussion_r3168556641

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
dcleao: See https://github.com/pentaho/hv-uikit-react/pull/5154/changes#r3153097525 (same issue).

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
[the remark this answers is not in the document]
dcleao: With the proposed approach, the only disadvantage is computing `resolvedBaseUrl` every time. Or, if you prefer, I can move the resolution computation to inside `initializeI18n`, although that ties the code there to some concerns not relevant to it... 

I don't see an issue with using `useMemo` for this. While not its intended, main use case, I believe it is a common idiom do sync ~~computation~~ side-effects guarded by change detection. React is a missing primitive imo: `useEffectSync`.

packages/app-shell-i18next/src/HttpResourcesBackend.ts
dcleao: We either support IntelliJ or VS Code, it seems, and each has a different convention.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: Also, this is not quite a "utility", imo. It gives access to App Shell's own functionality.

apps/docs/src/content/app-shell/api-reference.md
dcleao: Slightly relatedly, would it it be better to name the other new package, `@hitachivantara/app-shell-i18next`, instead something like `@hitachivantara/app-shell-util-i18next`?

packages/app-shell-i18next/src/HttpResourcesBackend.ts
dcleao: Addressed.

packages/app-shell-i18next/src/types.ts
dcleao: Addressed. Moved all types to their "home" module.

packages/app-shell-ui/vite.config.ts
dcleao: Can't say I fully understand this, but here's the response I got:
> `lib.entry` won't work here. Vite processes all entries through the bundler and outputs JS modules — `.json` files would become `.js` files. The locale files must remain as raw `.json` because they're fetched at runtime over HTTP by the i18next backend (`GET /locales/en/appShell.json`). `viteStaticCopy` is the right tool: it copies them as-is without transforming.

package.json
dcleao: Addressed.

package.json
dcleao: Addressed.

apps/docs/src/content/app-shell/internationalization.md
dcleao: Addressed.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
dcleao: Addressed.

apps/docs/src/content/app-shell/api-reference.md
dcleao: Building upon the other comment: https://github.com/pentaho/hv-uikit-react/pull/5154#discussion_r3168546188, there are actually two use cases, an internal and an external one.

Proposed solution:

1. `@hitachivantara/app-shell-shared`
    - `HvAppShellRuntimeContext`
    - `useHvAppShellRuntimeContext`
    - `HvAppShellRuntimeContextValue { i18n: i18n }` // i18next _type_ only
    
2. `@hitachivantara/app-shell-i18n` (external use case)
    - `i18next` agnostic
    - for use by app packages to get and change  the current App Shell language
    - API:
        - `context HvUseAppShellI18nContext` (new)
        - `type HvUseAppShellI18nContextValue { language: string, changeLanguage( lng ) : Promise<void> }`
        - `useHvAppShellI18n() : HvUseAppShellI18nContextValue`
            - re-renders whenever language changes
    - could not depend on `@hitachivantara/app-shell-shared`
    - only contains the context, types. implementation provided by app-shell-ui

3. `@hitachivantara/app-shell-ui`
   - creates and sets the `i18next` instance in the `HvAppShellRuntimeContext`
   - creates and sets the `HvUseAppShellI18nContext` backed by `i18next` (new)

4. `@hitachivantara/{app-shell-navigation, app-shell-ui, app-shell-shared}` (internal use case)
   - use the `i18next` instance in `HvAppShellRuntimeContext`
   - import `useTranslation` from `react-i18next` -  `i18next` modules get bundled in a shared asset

5. `@hitachivantara/app-shell-util-i18next` (optional internal/external use)
   - `i18next` utility for apps, essentially containing the HTTP backend for the conventional repository

packages/app-shell-ui/src/components/layout/Header/Header.tsx
zettca: thanks for the clarification, just wanted to make sure apps wouldn't be affected by this (can keep their existing i18n setups)

packages/app-shell-vite-plugin/src/vite-plugin.ts
```diff
@@ -201,6 +202,14 @@ export async function HvAppShellVitePlugin(
     }
   }
 
+  // Shared dependencies with a bundleSrc are pre-built and statically copied.
+  // Those with bundleSrc: null are built from source in this compilation
+  // (registered as virtual modules, compiled as rollup entries, not externalized in "app" builds).
```
nantunes: I'm confused by this. If not externalized, how will it be a shared dependency? Loaded views, from other packages, that, for example, use app-shell features like `useHvLocation` (from `@hitachivantara/app-shell-navigation`) that read the `@hitachivantara/app-shell-shared`'s shared context to get the available views, how will it get the same instance?

Maybe I'm missing something, but if not needed, I would not introduce this.

apps/docs/src/content/app-shell/internationalization.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+When present, translation backends (such as `HttpResourcesBackend` from `@hitachivantara/app-shell-i18next`) use this
+manifest to skip network requests for unsupported languages, letting i18next follow the `fallbackLng` chain instead.
+If the file is missing or malformed, all languages are allowed.
+
+When building with the **App Shell Vite Plugin**, the plugin automatically generates a `supported-locales.json` manifest.
+It merges the manifest provided by the application (if any) with the list of locales discovered on disk based on the
+locale directories found in `public/locales/`.
+Locales in the application manifest which do not have a corresponding directory on disk are ignored.
```
nantunes: If "locales in the application manifest which do not have a corresponding directory on disk are ignored." then isn't this just "generated from the locale directories found"? Why even try to merge? Is the order in the array important? Isn't the fallback mechanism built-in?

apps/docs/src/content/app-shell/configuration.mdx
```diff
@@ -544,13 +546,39 @@ In the example above, the name of the application, that appears on the browser t
...
+Defaults to `"./"`. When the default is used, apps compiled with the App Shell Vite Plugin will have their locale files served correctly from `locales/` in both production and development.
+
+The URL is resolved relative to:
+
+- `configUrl` (when the config was loaded from a URL), or
+- the configured [`baseUrl`](#baseurl) of the App Shell.
+
+Can coexist with [`translations`](#translations), which provides pre-loaded resources that are shown immediately and then refreshed when the HTTP backend responds.
+
+Set to `false` to disable HTTP-based translation loading entirely. This is useful when only inline [`translations`](#translations) are used and the extra network requests would cause 404 errors in legacy setups.
```
nantunes, line 578: How about the plugin "moving" the config's inline translations to the appropriate file in the locales folder, and always use HTTP (maybe except for `en`)? And only with `translationsBaseUrl: false` return to the legacy behaviour?

Just a thought, as it is it seems ok.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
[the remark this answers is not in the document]
nantunes: This really isn't a good pattern.
If you worry about resolvedBaseUrl, you can memo it, it seems independent of the initalizeI18Next call.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: See the related comment: https://github.com/pentaho/hv-uikit-react/pull/5154#discussion_r3181669441

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
```diff
@@ -1,70 +1,146 @@
...
+ * and the resolved config (or error) is returned synchronously.
+ */
+type ConfigRequest = {
+  promise: Promise<void>;
+  status: "pending" | "fulfilled" | "rejected";
+  result?: HvAppShellConfig;
+  error?: unknown;
+};
+
+const configRequests = new Map<string, ConfigRequest>();
```
nantunes, line 36: This is like a "poor man's" react-query / SWR? It's fine... In the past I copied code from https://github.com/andreiduca/use-async-resource for the same reason.
It just worries me this global in the future might cause problems in tests or something else.

packages/app-shell-ui/src/locales/zh-TW/appShell.json
nantunes: So now we're bringing that BA bug into the codebase of a new lib... Can't we use the proper language code and "patch" the bug in the http loader, with a good disclaimer comment next to it?

packages/app-shell-ui/src/i18n/index.ts
```diff
@@ -19,24 +23,101 @@ export const addResourceBundles = (
...
+
+  const resolvedBaseUrl = resolveTranslationsBaseUrl(translationsBaseUrl);
+
+  instance
+    .init<HttpResourcesBackendOptions>({
+      backend: {
+        baseUrl: resolvedBaseUrl,
+      },
+    })
+    .catch(() => {});
```
nantunes: Why swallow the errors? I imagine that if the init fails the we will be stuck at the loading screen instead of an error page?

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
nantunes: Can't you use https://www.i18next.com/overview/api#onlanguagechanged ?

nantunes (changes requested):
Going in a good direction. 👍

nantunes (changes requested):
> Proposed solution:
>
> 1. `@hitachivantara/app-shell-shared`
>     - `HvAppShellRuntimeContext`
>     - `useHvAppShellRuntimeContext`
>     - `HvAppShellRuntimeContextValue { i18n: i18n }` // i18next _type_ only

👍
 
> 2. `@hitachivantara/app-shell-i18n` (external use case)
>     - `i18next` agnostic
>     - for use by app packages to get and change  the current App Shell language
>     - API:
>         - `context HvUseAppShellI18nContext` (new)
>         - `type HvUseAppShellI18nContextValue { language: string, changeLanguage( lng ) : Promise<void> }`
>         - `useHvAppShellI18n() : HvUseAppShellI18nContextValue`
>             - re-renders whenever language changes
>     - could not depend on `@hitachivantara/app-shell-shared`
>     - only contains the context, types. implementation provided by app-shell-ui

👍

> 3. `@hitachivantara/app-shell-ui`
>    - creates and sets the `i18next` instance in the `HvAppShellRuntimeContext`
>    - creates and sets the `HvUseAppShellI18nContext` backed by `i18next` (new)

Can also subscribe to i18next's `onLanguageChanged` method and set the state of the `language` property.

> 4. `@hitachivantara/{app-shell-navigation, app-shell-ui, app-shell-shared}` (internal use case)
>    - use the `i18next` instance in `HvAppShellRuntimeContext`
>    - import `useTranslation` from `react-i18next` -  `i18next` modules get bundled in a shared asset

Was the mention of `{app-shell-shared}` intentional?

> 5. `@hitachivantara/app-shell-util-i18next` (optional internal/external use)
   - `i18next` utility for apps, essentially containing the HTTP backend for the conventional repository

Not critical, but worth exploring. It would be bundled, not shared, so I guess it could do that `import.meta.resolve?.(`${moduleId}/`)` boilerplate all projects need to copy?

---

Eventually `@hitachivantara/{app-shell-navigation, app-shell-events, app-shell-i18n}` could be just a single package `@hitachivantara/app-shell` external modules use when intentionally depending on App Shell functionality? Or even simply merged in the `@hitachivantara/app-shell-ui` package... It really doesn't matter.

apps/docs/src/content/app-shell/internationalization.md
dcleao: In an initial version, I attempted to preserve the order in the array. Later realized it was not meaningful.
The merge is amongst the locally provided locales and the ones "inherited" from the `app-shell-ui` package.
I will try to improve the description.

apps/docs/src/content/app-shell/api-reference.md
nantunes: > Proposed solution:
>
> 1. `@hitachivantara/app-shell-shared`
>     - `HvAppShellRuntimeContext`
>     - `useHvAppShellRuntimeContext`
>     - `HvAppShellRuntimeContextValue { i18n: i18n }` // i18next _type_ only

👍
 
> 2. `@hitachivantara/app-shell-i18n` (external use case)
>     - `i18next` agnostic
>     - for use by app packages to get and change  the current App Shell language
>     - API:
>         - `context HvUseAppShellI18nContext` (new)
>         - `type HvUseAppShellI18nContextValue { language: string, changeLanguage( lng ) : Promise<void> }`
>         - `useHvAppShellI18n() : HvUseAppShellI18nContextValue`
>             - re-renders whenever language changes
>     - could not depend on `@hitachivantara/app-shell-shared`
>     - only contains the context, types. implementation provided by app-shell-ui

👍

> 3. `@hitachivantara/app-shell-ui`
>    - creates and sets the `i18next` instance in the `HvAppShellRuntimeContext`
>    - creates and sets the `HvUseAppShellI18nContext` backed by `i18next` (new)

Can also subscribe to i18next's `onLanguageChanged` method and set the state of the `language` property.

> 4. `@hitachivantara/{app-shell-navigation, app-shell-ui, app-shell-shared}` (internal use case)
>    - use the `i18next` instance in `HvAppShellRuntimeContext`
>    - import `useTranslation` from `react-i18next` -  `i18next` modules get bundled in a shared asset

Was the mention of `{app-shell-shared}` intentional?

> 5. `@hitachivantara/app-shell-util-i18next` (optional internal/external use)
>    - `i18next` utility for apps, essentially containing the HTTP backend for the conventional repository

Not critical, but worth exploring. It would be bundled, not shared, so I guess it could do that `import.meta.resolve?.(${moduleId})` boilerplate all projects need to copy?

---

Eventually `@hitachivantara/{app-shell-navigation, app-shell-events, app-shell-i18n}` could be just a single package `@hitachivantara/app-shell` external modules use when intentionally depending on App Shell functionality? Or even simply merged in the `@hitachivantara/app-shell-ui` package... It really doesn't matter.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
dcleao: This implements the suspense pattern for getting the configuration, when obtained from `configUrl`.
In this way, it is ensured that i18next is only placed in the component tree (via `I18nextProvider`) once its initialization has been done/started. Not doing so would cause issues downstream, when using `useTranslation`.

Implementing the suspense pattern always requires storing whatever is "initializing" at the module level. React query/ SWR do this, yes. If you'd prefer, I could add react query to app-shell-ui and use it to achieve a similar effect.

packages/app-shell-ui/src/locales/zh-TW/appShell.json
dcleao: Good point. Now that there is a current App Shell language, this can be done by the Pentaho BA code that syncs it with the language of the current Pentaho session. App plugins would always be asked to load the correct language code.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: An in some ways, re-implement `useTranslation`, having to use getFixedT?
But yes, that would be a solution.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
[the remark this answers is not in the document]
dcleao: While I disagree with the sentiment regarding this type of use of `useMemo`, I am fine rewriting this as `useMemo` for `resolvedBaseUrl` plus a sync call to `initializeI18n` guarded by `config` being defined. `initializeI18n` already handles re-entry, so this ends up being equivalent.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: I think it should be possibly to use / piggy-back on `useHvAppShellI18n` to re-render when language changes.

apps/docs/src/content/app-shell/api-reference.md
dcleao: -----
> Can also subscribe to i18next's onLanguageChanged method and set the state of the language property.

Do you mean this regarding the use of `useTranslation` referenced in `#4`, otherwise, not seeing what you'd mean by this regarding `#3`.

Already partially addressed the `onLanguageChanged` issue in a related comment: https://github.com/pentaho/hv-uikit-react/pull/5154#discussion_r3183962006. 

See the next comment for a revised proposal.

----
> Was the mention of {app-shell-shared} intentional?

No, not when taken together with `useTranslation`, as that would imply directly depending on `i18next`, or sharing `useTranslation` as a context prop (like in the current proposal). My bad.

See the next comment for a revised proposal.

----
> 5. @hitachivantara/app-shell-util-i18next (optional internal/external use)
       i18next utility for apps, essentially containing the HTTP backend for the conventional repository
>
> Not critical, but worth exploring. It would be bundled, not shared, so I guess it could do that 
> import.meta.resolve?.(${moduleId}) boilerplate all projects need to copy?

Yes, this package would be bundled, not shared.
This is an optional package.
The main value of this package is the backend implementation, allowing all apps to easily support the locales repository format being recommended.

dcleao: I discussed much of this with @zettca.

The following design looks like a good compromise:

Proposed solution:

1. `@hitachivantara/app-shell-shared`

    For internal use:
    - `HvAppShellRuntimeContext` 
    - `useHvAppShellRuntimeContext`
    - `HvAppShellRuntimeContextValue { i18n: i18n }` // i18next _type_ only

    For external use:
    - `i18next` agnostic
    - for use by app packages to read and change  the current App Shell language
    - `context HvUseAppShellI18nContext`
    - `type HvUseAppShellI18nContextValue { language: string, changeLanguage( lng ) : Promise<void> }`
    - `useHvAppShellI18n() : HvUseAppShellI18nContextValue`
        - re-renders whenever language changes // <- allows not having to use `useTranslation` just to re-render upon language changes
    - preserves `app-shell-shared` in the original spirit: just contain contexts et al.; does add a new public "entity domain", i18n; the context's value "implementation" is fully provided by `app-shell-ui`, unlike in the current code proposal
    
3. `@hitachivantara/app-shell-ui`
   - creates and sets the `i18next` instance in the `HvAppShellRuntimeContext`
   - creates and sets the `HvUseAppShellI18nContext` backed by the same `i18next` instance

4. `@hitachivantara/{app-shell-navigation, app-shell-ui, app-shell-shared}` (internal use case)
   - use the `i18next` instance in `HvAppShellRuntimeContext`
   - import `useTranslation` from `react-i18next` -  `i18next` modules get bundled in a shared asset
      - for `app-shell-shared` might either use `useHvAppShellI18n` as a way to refresh upon language changes, or, instead, implement a local, simplified `useTranslation` (will revert the changes making app-shell-shared being compiled together with the ui)

5. `@hitachivantara/app-shell-i18next` (**optional** internal/external use)
   - obviously, explicitly depends on `i18next`!
   - `i18next` **utility** for apps, essentially containing the HTTP backend for the recommended conventional repository layout
   - the only **new** package

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: This hook will turn into a basic `useContext` wrapper. The actual implementation, and i18next dependency, will move to the context value and be provided by the app-shell-ui.

packages/app-shell-vite-plugin/src/vite-plugin.ts
dcleao: It's still external. This was just changing how it is compiled to become external. Ignoring the original compilation, compiling locally and re-exporting under original name.

In any case, I will be removing this as it should not be needed in the new proposed design: https://github.com/pentaho/hv-uikit-react/pull/5154#discussion_r3184079877.

apps/docs/src/content/app-shell/configuration.mdx
dcleao: The proposed behavior has in part to do with preserving backwards compatibility by the App Shell UI, even if not recompiled with the vite plugin, even if not compiled with it. At some point, we could remove inline translations support by the App Shell UI. Seeing inline translations as "pre-loaded" and external translations as replacing / refreshing the former, instead of, for example being merged with, was the easiest route to implement (as this is what i18next can gives us for free; note while `addResourceBundle` allows achieving "deep-merge", via `deep` and `overwrite` options, this is not as easy to achieve from within the backend's code). Also aligns with the intent to remove.

In any case, the idea to convert inline translations into external translations in the vite plugin looks very nice.

In a way, this would complement what the vite plugin already does: deep-merging external translations from `app-shell-ui` with those provided locally (as overrides). I am not sure you meant to include the merging aspect in your proposal between inline and external translations.

The new proposed `/locales` endpoint is actually already deep-merging inline and external translations (see `PluginLocaleProcessor` in https://github.com/pentaho/pentaho-app-shell/pull/1235/changes#diff-a5192c66e2c6a96fa464cc22c78785f530ed2ed25140de5a6706272b3e1dce07)
(🤦 and one design principle is to have the vite plugin and the Pentaho server behave similarly, whenever possible).

In retrospect, I am not sure that this deep-merging of inline and external translations by the server is a good thing. When merging the locales repositories of the app-shell-ui and of the app being built, I see this as positive, given it allows "overriding" only certain keys (is this really needed/good?). However, in the inline vs external, I see it more as an evolution: the old vs the new way? Or, if we keep both (as currently documented), I see these as "pre-loaded" vs "http-loaded" -- different trade-offs: convenience vs load speed for common languages. But the idea is really for one _or_ the other to be used. So, refresh/replace, and not merge, would be more adequate.

---

I will explore generating external translations for inline translations (and removing these) in the vite plugin. However, this might be harder to provide at this point, given it requires access to the app shell config which I believe is not easily accessible in the vite plugin's current form (`.ts` code is converted to `.json` in some internal vite plugin) (the planned multi-app changes already refactored this part quite a bit to have early access to the app shell config json).

packages/app-shell-ui/src/i18n/index.ts
dcleao: Yeah, you're right, in a way... 😬 

_Config_ loading is being guarded using a suspense boundary, with `HvLoading`. There is no localized text to display anything more user-friendly than this, like `LoadingPage`.

I18next initialization follows when _Config_ becomes available. Still, no better UX is possible at this point.

I18next's `init(...)` call may reject if the configured default namespaces / initial detect language combination fails to load by the backend. Not common, but may happen.

While waiting for loading of the "default bundles" to finish, uses of `useTranslation` by downstream components will cause suspension to the above described suspense boundary. 

----

If something fails at this point, note that the App Shell is barely initialized, and it would be hard to send the user to pages that (likely?) depend on it being initialized, like `ErrorPage`.

----

If you want I can try to explore improving on this.

I could try to use the new `useI18n` hook which already handles properly catching initialization errors (including those not reported by the init promise rejection...). It will throw the error to be caught by an error boundary.

The current error boundary is below i18next initialization. But I can try to add a new one above, without i18n requirements...

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
dcleao: Regarding tests, that's the reasons why I added support for multiple `configUrl` and am using a `Map<string, .>`. Not enough, I know, but may reduce any issues.

@zettca mentioned the new `use` API in react 19, which should be able to replace the need for this explicit "suspense" pattern implementation.

We'd be able to replace this with `use` when we upgrade to that version of react.

https://react.dev/reference/react/use

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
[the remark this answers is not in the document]
dcleao: Addressed.

Added a hook `useI18nInit()` that handles all this plus the error handling issue raised in another comment.

packages/app-shell-ui/src/locales/zh-TW/appShell.json
dcleao: Addressed.

Ended up implementing the exception in Pentaho App Shell's own "BFF" locale endpoint.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: Addressed.

All App Shell internal packages can use `useTranslation` directly, except `app-shell-shared`, which is piggy-backing on `useHvAppShellI18n` to react to language changes and then using `i18n.getFixedT(.)` to get a `t` function.

apps/docs/src/content/app-shell/api-reference.md
dcleao: Addressed.

Implemented as described in previous comment.

packages/app-shell-shared/src/AppShellRuntimeContext.tsx
dcleao: Addressed.

This hook is now "degenerate" - it only returns the context value. The actual context value implementation was moved to `app-shell-ui`.

packages/app-shell-vite-plugin/src/vite-plugin.ts
dcleao: Addressed.

Removed this commit.

apps/docs/src/content/app-shell/internationalization.md
dcleao: Addressed.

Changed the text to make it clearer that the manifest is completely rebuilt from the discovered locales folders.

apps/docs/src/content/app-shell/configuration.mdx
dcleao: Have not addressed this at this point.
Waiting for your feedback, but time is running out on this turn.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
dcleao: Made no changes regarding this.

packages/app-shell-ui/src/i18n/index.ts
dcleao: Addressed.

Added a hook `useI18nInit()` that handles initialization by calling `useI18n`. Added an error boundary with a fallback component based on `GenericError` but for the case where i18n initialization fails, and only has text in English.

zettca (approved):
LGTM 👍

great work 💪

packages/app-shell-ui/src/i18n/index.ts
dcleao: Actually, in the meantime, discovered that it is not reliable to handle initialization errors from i18next, given we cannot easily distinguish between "fatal" and "intermediate" errors, resulting from the normal language fallback process. The previous / originally proposed code from `useI18n` had this issue. I ended up removing the "initialization error" handling capability of the hook... This follows `i18next`'s own design, in that, ultimately, a failed initialization degrades "gracefully" by showing untranslated keys...

I kept the newly added error boundary, given it was needed after all to handle any config loading errors.

apps/docs/src/content/app-shell/configuration.mdx
nantunes: It was just an idea, nothing needs to be changed in the PR.

packages/app-shell-ui/src/components/AppShell/AppShellContainer.tsx
nantunes: Again, just a comment. For a single use, I don't see the need to bring in react-query. I also looked at `use`, but we're still behind.

packages/app-shell-ui/src/i18n/index.ts
nantunes: Just do your best. :-D

nantunes (approved):
You can mark my comments as resolved.

PR #5170, opened 2026-05-13
fix(Badge): support inline position with other content

plagoa:
Fix misplacement of the badge when it doesn't have children allowing for correct positioning when placed inline with other content.

image

### DIFF
--- a/packages/core/src/Badge/Badge.styles.tsx
+++ b/packages/core/src/Badge/Badge.styles.tsx
@@ -31,6 +31,11 @@ export const { staticClasses, useClasses } = createClasses("HvBadge", {
     wordBreak: "keep-all",
     textAlign: "center",
 
+    "&$badgeIcon": {
+      top: "1px",
+      left: "calc(100% - 7px)",
+    },
+
     ":empty": {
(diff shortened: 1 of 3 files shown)


packages/core/src/Badge/Badge.styles.tsx
```diff
@@ -41,4 +41,9 @@ export const { staticClasses, useClasses } = createClasses("HvBadge", {
   },
   badgeIcon: { top: "1px", left: "calc(100% - 7px)" },
   badgeOneDigit: { padding: 0, width: "16px" },
+  badgeInline: {
+    position: "relative",
+    top: "auto",
+    left: "auto",
+  },
```
zettca: these are _basically_ the default `div` styles

instead of "squashing" the existing absolute styles with the default, can we apply the `absolute styles only when not in `badgeInline`?
also, can we make this a `data-badge-inline` attribute, instead of a class?

something like this:
```tsx
":not([data-badge-inline])": {
    position: "absolute",
    top: 0,
    left: "100%",
}
```

zettca (approved):

PR #5172, opened 2026-05-15
feat(AppShell): generate dist artifacts for new packaging layout [PPUC-727]

loonatic:


packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,242 @@
...
+          distPrefix,
+          sourceCondition,
+        );
+      }
+    }
+  }
+
+  result["./package.json"] = "./package.json";
+  result["./app-shell.config.json"] = "./app-shell.config.json";
+
```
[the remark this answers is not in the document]
dcleao: I don't like testing for file existence for this. While it _could_ match the intended behavior, it also would hide/obfuscate the underlying logic.

Let me revisit the intended behavior.

Isn't the goal that the new experimental package layout always bundles both `package.json` and `app-shell.config.json`, regardless of other options? I believe so. In that case, regardless of `inlineConfig` being `true`, the package layout should be preserved, imo. This ensures that compiled "app" packages are also always "bundle" packages, i.e. they are dual use -- as an app shell, or as an app bundle for another app shell.

Not sure about the `generateEmptyShell` case, and whether it qualifies as an exception.
Otherwise, the only option that should prevent generating the `app-shell.config.json` file is that which controls enabling the experimental package layout. And this option already likely controls whether this plugin (the dist `package.json` plugin) is enabled at all (because `package.json` and `app-shell.config.json` should be treated the same in this regard)?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,248 @@
...
+      const stripped = stripPrefix(distPrefix, value);
+      result[key] = stripped;
+      if (!firstJsPath && stripped.endsWith(".js")) firstJsPath = stripped;
+    } else if (typeof value === "object" && value !== null) {
+      result[key] = transformConditions(
+        value as Record<string, unknown>,
+        distPrefix,
+        sourceCondition,
+      );
+    }
```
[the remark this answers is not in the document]
dcleao: I am not sure that we should support the array syntax in `exports`, given they represent fallback paths. While node supports this out of the box, import maps -- the main target environment of these packages -- by themselves, do not.

Would it be best to assume this and pick the first entry upfront?

apps/default-app/vite.config.ts
```diff
@@ -13,7 +13,8 @@ export default defineConfig(({ mode }) => ({
     }),
     HvAppShellVitePlugin({
       mode,
-      type: "app",
+      experimentalNewPackageLayout: true,
+      type: "bundle",
```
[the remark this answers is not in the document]
dcleao: It seems that, now, you have removed `experimentalNewPackageLayout: true,` entirely, and thus opted for not having a sample app to test the new layout. I believe that it would be useful to have it.

Was that intentional?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,256 @@
...
+ */
+const DEFAULT_SOURCE_CONDITION = "@pentaho-apps:source";
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+/**
+ * Strips a directory prefix from a relative path.
+ */
+function stripPrefix(prefix: string, value: string): string {
+  return value.startsWith(prefix) ? `./${value.slice(prefix.length)}` : value;
```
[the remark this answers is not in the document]
dcleao: AFAICT, the node.js exports spec explicitly says that paths must be relative. So, we should _not_ be trying to be lenient and supporting `dist/`. See https://nodejs.org/api/packages.html#path-rules-and-validation-for-export-targets.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,256 @@
...
+      if (typeof value === "string") {
+        const stripped = stripPrefix(distPrefix, value);
+        if (stripped.endsWith(".js")) {
+          result[subpath] = {
+            types: stripped.replace(/\.js$/, ".d.ts"),
+            default: stripped,
+          };
+        } else {
+          result[subpath] = stripped;
+        }
```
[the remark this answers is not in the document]
dcleao: This actually happened to me sometimes. Specifically, not for a "source"/dev-time-only export. The issue is that we never know whether an export is a legitimate source export and not a mistake that we should error out on. However, it seems clear, at least, that any export outside the `./dist` folder should be filtered out.

What are your thoughts on this?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  if (!fs.existsSync(outDir)) return false;
+
+  const distPrefix = `./${buildOutDir}/`;
+  const distPkg: Record<string, unknown> = {};
+
+  for (const field of ["name", "version", "type", "license", "description"]) {
+    if (pkg[field] != null) distPkg[field] = pkg[field];
+  }
+
```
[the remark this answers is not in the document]
dcleao: ~~What was your reasoning for ignoring?~~
~~It seems to me that App Shell built packages already have bundles that should be respected (not destructured), or otherwise runtime semantics might not be kept.~~
~~So, at most, if we wanted to be explicit about it, we'd emit (the default) `sideEffects: true`.~~

Just saw:
```
if (pkg.sideEffects != null) {
    distPkg.sideEffects = Array.isArray(pkg.sideEffects)
      ? pkg.sideEffects.map((entry: string) => stripPrefix(distPrefix, entry))
      : pkg.sideEffects;
  }
```

And some info regarding the property:
> A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export.

Still, not sure if this property makes sense to our type of package. In a way, even for pre-built bundles that should not be dismantled/destructured, a downstream bundler may need to know imports for them should always be kept due to internally performing side-effects...

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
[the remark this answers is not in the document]
dcleao: If you revert, check out the related https://github.com/pentaho/hv-uikit-react/pull/5172#discussion_r3259303529 as well.

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
```diff
@@ -114,9 +114,10 @@ export default function processConfiguration(
 
       finalAppShellConfig.apps = undefined;
 
-      // Replace all @self references using simple string replacement
+      // Replace all $app and @self (deprecated) references using simple string replacement.
       let configString = JSON.stringify(finalAppShellConfig);
 
+      configString = configString.replaceAll(`"$app/`, `"${selfAppName}/`);
```
dcleao, line 139: Let's consider where/how in the source code/documentation we will be marking `@self` as deprecated, and make sure it's easy to remove later on.

packages/app-shell-vite-plugin/src/vite-plugin.ts
```diff
@@ -136,6 +137,60 @@ export interface AppShellVitePluginOptions {
    * @default false
    */
   disableAppsKeyNormalization?: boolean;
+  /**
+   * Enables the experimental new package layout feature set.
+   *
+   * When `true`, the plugin activates behaviours that are part of the
+   * new workspace package layout convention. Currently this gates:
```
dcleao: What `workspace`? If referring to the npm workspace, then that is unrelated, AFAICT.

packages/app-shell-vite-plugin/src/vite-plugin.ts
```diff
@@ -136,6 +137,60 @@ export interface AppShellVitePluginOptions {
...
+   *
+   * - **`dist/package.json` generation** — a cleaned-up `package.json` is
+   *   written to the Vite output directory after each build so that workspace
+   *   symlinks can be resolved correctly by Vite and Node.
+   *
```
dcleao, line 155: > so that workspace symlinks can be resolved correctly by Vite and Node.

I think this is incorrect, in two ways:
1. the export paths in the source `package.json` point to either source (for the `@pentaho-apps:source` condition` or `./dist`. It is the root `package.json` which is used by dependent packages in the (NPM) workspace. The source export is used by the package being compiled/tested or to help the code editor, via ts config. The dist exports are used by dependent packages being compiled/tested.
2. `so that` -- the presented reason is (incorrect, as explained above, and) _narrow_; the dist `package.json` is part of a new app bundle distribution/publishing package layout; certain properties in the source `package.json` either do not apply or need to be adapted.

packages/app-shell-vite-plugin/src/vite-plugin.ts
```diff
@@ -136,6 +137,60 @@ export interface AppShellVitePluginOptions {
...
+   * - `inlineConfig` and `generateEmptyShell` target **App Shell container
+   *   image builds**, where the config is embedded in `index.html` rather
+   *   than written as a separate `app-shell.config.json` file.
+   *
+   * When `inlineConfig` is `true` (or implicitly via `generateEmptyShell`),
+   * `app-shell.config.json` is **not** written to the output directory.
+   * The `dist/package.json` generator accounts for this by only adding the
+   * `"./app-shell.config.json"` export entry when the file is present in the
+   * output directory.
+   *
```
dcleao: The previous list after "Currently, this gates:" misses one item for the `app-shell.config.json` file. It doesn't matter if `inlineConfig` and `generateEmptyShell` slightly modify this behavior (and I have contested if they should in another comment), this item is still of the same relevancy of the `package.json` and should be introduced side-by-side in the list, imo. If needed, any conditions under which that doesn't apply may be complemented in a follow-up paragraph.

dcleao (changes requested):

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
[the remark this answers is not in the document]
loonatic: Refactored. `inlineConfig` being set together with the experimental feature will place it on the index.html while also generating the file. As for `generateEmptyShell` and since it doens't require any `app-shell.config.json`, if set, it will skip writing the file.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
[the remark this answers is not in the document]
loonatic: Went with the first entry.

apps/default-app/vite.config.ts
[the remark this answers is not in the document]
loonatic: Not intentional, the goal is to have default-app with the experimental feature flag `true`.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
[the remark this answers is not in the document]
loonatic: Reverted.

packages/app-shell-vite-plugin/src/vite-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
loonatic: Added deprecation notice in some more code and also updated uikit website docs as well.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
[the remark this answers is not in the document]
loonatic: Filtering exports outside outdir seems reasonable and I'm sure with usage, we can revisit this logic.

apps/docs/src/content/app-shell/index.md
```diff
@@ -14,6 +14,7 @@ At its core, **App Shell** consists of:
 The **App Shell** comprises the following packages:
 
 - [`@hitachivantara/app-shell-events`](https://npm.im/@hitachivantara/app-shell-events)
+- [`@hitachivantara/app-shell-i18next`](https://npm.im/@hitachivantara/app-shell-i18next)
```
dcleao, line 17: 🙏

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
```diff
@@ -118,11 +129,19 @@ export default function processConfiguration(
...
       finalAppShellConfig = JSON.parse(configString);
 
-      if (!inlineConfig) {
+      // Write app-shell.config.json to dist when:
+      // - inlineConfig is false (standard flow), OR
+      // - experimentalNewPackageLayout is true (ensures dual-use: app shell or app bundle)
+      // Exception: generateEmptyShell has no real config to write.
+      if (
+        !inlineConfig ||
+        (experimentalNewPackageLayout && !generateEmptyShell)
```
dcleao: Doesn't `generateEmptyShell` also affect `inlineConfig`? Should the and condition apply at the top?

packages/app-shell-vite-plugin/src/vite-plugin.ts
```diff
@@ -140,57 +140,40 @@ export interface AppShellVitePluginOptions {
...
-   * It is forwarded to the `dist/package.json` generator, which removes the
-   * condition from the transformed `exports` field of the output manifest.
+   * build step. The condition is stripped from `dist/package.json` so it never
+   * leaks to consumers that don't have the TypeScript sources available.
    *
    * @default "@pentaho-apps:source"
-   * @experimental
    */
-  experimentalSourceCondition?: string;
+  sourceCondition?: string;
```
dcleao, line 176: After the new behavior was added, of filtering out exports outside the `./dist` folder, realized that that would also cover any source conditions, as these, supposedly, will not point to the `./dist` directory. Still, because there might be some unforeseen use case where the "source" would be in the same folder as the dist folder, and because with this the semantics is more explicit, and because we (IIRC) will need the `sourceCondition` for other purposes, I think we should keep it and the corresponding behavior of explicitly removing those from exports.

dcleao (changes requested):

packages/app-shell-vite-plugin/src/vite-watch-config-plugin.ts
```diff
@@ -8,6 +8,9 @@ const prepareConfigForDevMode = (
 ) => {
   let configString = JSON.stringify(config);
 
+  // $app is the preferred placeholder as @self is deprecated.
+  configString = configString.replaceAll(`"$app/`, `"${selfAppName}/`);
+  // @deprecated — `@self/` is deprecated in favour of `$app/`. To be removed in a next major.
```
zettca: `@deprecated` is for JSDoc. We've been using `// TODO(<major>)` for similar scenarios. let's keep it that way for consistency/discoverability

also, IMO we just need a single since of this in the code. Once we remove it, we'll remove it everywhere

```suggestion
  // TODO(major): remove in favour of `$app/`
```

package-lock.json
```diff
@@ -773,6 +773,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
```
zettca: this shouldn't be here. ensure you're using the npm version declared under `packageManager`

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
```diff
@@ -14,20 +14,31 @@ import sharedDependencies from "./shared-dependencies.js";
...
-  inlineConfig: boolean,
-  generateEmptyShell: boolean,
-  modules: string[] = [],
+  modules: string[],
+  options: {
+    buildEntryPoint: boolean;
+    inlineConfig: boolean;
+    generateEmptyShell: boolean;
+    experimentalNewPackageLayout: boolean;
+  },
```
zettca: this is a nice move, as so many arguments [is a bad pattern](https://docs.deno.com/runtime/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object).

but since we're changing the API, let's just make it a single object (with a proper `interface`) - there's no "primary" parameter here, and leaving 4 parameters "outside" is still too much

thanks! 😄

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
zettca: IMO it's as easy as Find & Replace `"@self"` (or let an LLM do it)

I don't see a need to add those notices everywhere. I much prefer a smaller diff (now, and once we remove it)

apps/default-app/app-shell.config.ts
```diff
@@ -185,7 +185,7 @@ export default {
           },
           conditions: [
             {
-              bundle: "@self/conditions/useAsyncTrue.js",
+              bundle: "$app/conditions/useAsyncTrue.js",
```
zettca, line 188: can/should we leave some `@self`, so we can "continuously" validate they still work as intended? just a question

zettca (commented):
Great work 👍 

Just some small comments, I'll review the main `dist-package-json` feature _soon_

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+  distPkg.exports = transformExportsForDist(
+    pkg.exports,
+    distPrefix,
+    sourceCondition,
+    outDir,
+  );
+
+  if (pkg.main) distPkg.main = stripPrefix(distPrefix, pkg.main);
+  if (pkg.types) distPkg.types = stripPrefix(distPrefix, pkg.types);
+  if (pkg.typings) distPkg.typings = stripPrefix(distPrefix, pkg.typings);
```
zettca: what's the purpose of this? we already have `exports`
`types`/`typings` are not even _standard_

this probably should be documented

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+ * When a prefix match is found, the returned path starts with `./`.
+ * Non-matching values are returned unchanged.
+ */
+function stripPrefix(prefix: string, value: string): string {
+  if (value.startsWith(prefix)) {
+    return `./${value.slice(prefix.length)}`;
+  }
+
+  return value;
+}
```
zettca, line 74: What's the purpose of this? Why do we need to"hyjack" the user-defined values?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  for (const field of [
+    "dependencies",
+    "peerDependencies",
+    "peerDependenciesMeta",
+    "optionalDependencies",
+  ]) {
+    if (pkg[field]) distPkg[field] = pkg[field];
+  }
```
zettca: why are we cherry-picking the fields to include? can't we just keep what's already there, and simply modify what we need?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+  outDir: string,
+): Record<string, unknown> {
+  const result: Record<string, unknown> = {};
+
+  if (rawExports == null || typeof rawExports !== "object") {
+    const hasIndex = fs.existsSync(path.join(outDir, "index.js"));
+    if (hasIndex) {
+      result["."] = { types: "./index.d.ts", default: "./index.js" };
+    }
+    result["./*"] = "./*";
```
zettca: why are we doing this? isn't having `exports` part of the contract of this feature?

also, we don't need the `types`
```
result["."] = "./index.js";
```

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+  } else {
+    const exports = rawExports as Record<string, unknown>;
+    for (const [subpath, value] of Object.entries(exports)) {
+      if (!subpath.startsWith(".")) continue;
```
zettca:
```suggestion
```

[the spec](https://nodejs.org/api/packages.html#exports) already doesn't allow otherwise

Image

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      if (typeof value === "string") {
+        const stripped = stripPrefix(distPrefix, value);
+        // Skip exports that don't resolve within the output directory.
+        if (stripped === value) continue;
+        if (stripped.endsWith(".js")) {
+          result[subpath] = {
+            types: stripped.replace(/\.js$/, ".d.ts"),
+            default: stripped,
+          };
```
zettca: what's the purpose of this? this is already the default behaviour

```
"./test": "./test.js",
```
will automatically search for a `test.d.ts`

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+        } else {
+          result[subpath] = stripped;
+        }
+      } else if (Array.isArray(value)) {
+        // Fallback arrays: pick the first string entry (import maps don't support fallbacks).
```
zettca: what is this? ~it's not part of the [`exports`](https://nodejs.org/api/packages.html#exports) spec~

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+  outDir: string,
+  buildOutDir: string,
+  sourceCondition: string,
+): boolean {
+  const pkgJsonPath = path.join(root, "package.json");
+  if (!fs.existsSync(pkgJsonPath)) return false;
+
+  const pkg = JSON.parse(fs.readFileSync(pkgJsonPath, "utf-8"));
+
+  if (!fs.existsSync(outDir)) return false;
```
zettca: Why does this `return false`? Let's just wrap this in a `try`/`catch` instead, as it can fail in other places too (eg. `JSON.parse()`). This way we can cleanup the code too:

```suggestion
  const pkg = readJsonFile(path.join(root, "package.json"));
```

(we have a `readJsonFile` util)


```tsx
try {
  generateDistPackageJson()
}
```

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+          value as Record<string, unknown>,
+          distPrefix,
+          sourceCondition,
+        );
+      }
+    }
+  }
+
+  result["./package.json"] = "./package.json";
+  result["./app-shell.config.json"] = "./app-shell.config.json";
```
zettca: if this is always part of `result`, then let's move this up and initialise it with this
will allow for early `return result`, simplifying the logic & nesting

On the other hand... shouldn't this also be the users' responsibility? (in reality, UI Kit's CLI). Why "hide" that this is happening?

zettca (commented):
Looking good overall.

We still need some docs on how to use this `exports` feature.

Also, I think there's unnecessary work being done in a few places.
I think we should follow/implement/enforce only the modern Node APIs, and assume users are following them correctly, and not bother with older/non-standard APIs, or correcting/handling wrong usage

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
loonatic: Yes, my bad, removed.

packages/app-shell-vite-plugin/src/vite-plugin.ts
loonatic: Agreed, kept it as explicit semantic stripping.

packages/app-shell-vite-plugin/src/vite-watch-config-plugin.ts
loonatic: I think you suggested on keeping only 1 of these TODO comments, but ended up normalizing it on all 3 instances as I feel it doesn't bother that much, but let me know.

package-lock.json
loonatic: My bad, already using Node 26 but for some reason my node version was still set as default for 24. Reverted.

packages/app-shell-vite-plugin/src/vite-configuration-processor-plugin.ts
loonatic: Addressed.

apps/default-app/app-shell.config.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Nice catch, node and typescript both resolve through exports first so removed them entirely and added them to an EXCLUDED_FIELDS set.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: The source `package.json` lives at the project root and references files like `./dist/foo.js` and since the generated `package.json` is written inside `dist/`, those paths need to be rewritten relative to the new location. It's just paths normalization in a way.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Flipped to exclusion-based.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: `exports` is part of the contract **but** we also try to be Node compliant (`exports` not mandatory and has a default) and thus, when `exports` is not declared we set a default mapping access and also to any subfolders below since this is best phase (with most information available) where we can properly mount the generated `package.json` with the `index.d.ts` and, also, to simplify consumers code, otherwise they would have to declare default values.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Removed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Removed, seems fallbacks are a Node legacy feature.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Moved to result initialization to be clearly visible.
The goal was to avoid user input as much as needed, and since these are artifacts the plugin itself generates (`dist/package.json` and `dist/app-shell.config.json`) I believe its the plugin responsability to also declare them.
@dcleao what is your opinion?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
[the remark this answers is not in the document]
loonatic: Currently, user declares it and we respect it. I haven't add it to the EXCLUDED_FIELDS set so it gets copied through.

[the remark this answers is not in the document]
loonatic: In offline conversation, we discussed this and got to the conclusion that sideEffects can be dropped from the generated `package.json` as bundles, in App Shell, are considered **external** and thus no downstream bundler processes them.
Added to the EXCLUDED_FIELDS set.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: `types` are useful in multi-app development to better debug in dev mode however I'm wondering if your comment is related to Typescript resolving `.d.ts` from `.js` files "automatically"?

package-lock.json
zettca: they flipped the behaviour again? 😞 

this is related to `npm` (although bumping `node` also bumps `npm`)

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
zettca: All right 👍 

If these files are to be "invisible" to the user, then it shouldn't be their responsibility handling this.

I still don't have all the context regarding how this will be built & distributed

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
zettca: What do you mean "Node compliant"?

[`exports` is the new `main`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#exports), and supersedes it. trying to support both is a waste of time IMO
besides, `main` only allows for a single `index.js` [barrel file, which is a code smell](https://tkdodo.eu/blog/please-stop-using-barrel-files) (especially in runtime/unbundled code), and is not how we recommend exporting modules (ie. the multiple `modules: []` entries)

regarding `.d.ts`, yes. it's automatic

apps/default-app/app-shell.config.ts
```diff
@@ -288,67 +288,68 @@ export default {
...
       // { bundle: "@hv/sample-app/headerActions/ChangeDefaultAppContext.js" },
     ],
   },
 
   mainPanel: {
     views: [
-      // @self app
+      // Own application
+      // TODO(major): remove @self/ support in favour of $app/
       { bundle: "@self/pages/Debug.js", route: "/debug" },
```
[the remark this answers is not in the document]
loonatic: Intended to make sure @self/ keeps working until being removed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,247 @@
...
+      distPrefix,
+      sourceCondition,
+    );
+  } else {
+    distPkg.exports = { "./*": "./*" };
+  }
+
+  distPkg.exports["./package.json"] = "./package.json";
+  distPkg.exports["./app-shell.config.json"] = "./app-shell.config.json";
+
```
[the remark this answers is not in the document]
loonatic: Different concerns, shouldn't be used together and the dev must know what the goal is.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -0,0 +1,247 @@
...
+  const result: Record<string, unknown> = {};
+
+  for (const [key, value] of Object.entries(obj)) {
+    if (key === sourceCondition) continue;
+
+    if (typeof value === "string") {
+      result[key] = stripPrefix(distPrefix, value);
+    } else if (typeof value === "object" && value !== null) {
+      result[key] = transformConditions(
+        value as Record<string, unknown>,
```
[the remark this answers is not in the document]
loonatic: https://github.com/pentaho/hv-uikit-react/pull/5172#discussion_r3273210351

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: > trying to support both is a waste of time IMO

I agree, in a way. But you also waive node compliance. Recommendation does not mean "cannot use", or even being deprecated. We aim to comply with the semantics of the `package.json` properties for the features we support. `main` is surely a part of it, in its relation to `exports`.

---
> single index.js [barrel file, which is a code smell](https://tkdodo.eu/blog/please-stop-using-barrel-files) (especially in runtime/unbundled code)

In which we read:
> How to organize your files “correctly” is a hot, controversial topic among frontend developers.

That tells us there's likely one more way to see things, and to legitimately call it a good or bad practice.

I don't think the App Shell vite plugin should cast a vote in this war. Instead it should try to not prevent legitimate uses. As long as there isn't much more work involved.

---

>  is not how we recommend exporting modules (ie. the multiple modules: [] entries)

Do you know where this is written?

AFAIK, the only requirement is that anything that must be referenced from `app-shell.config.json`, in a `bundle` property, must be a module. All else, seems to be a matter of hot debate.

---

All that said.

I am fine with not supporting `main` (nor `module`, nor `browser`, nor `types`). Mostly to reduce the complexity of the code, and because `exports` is the more modern and capable alternative.

Regarding the `types` condition, it seems you are right that TS does not require it as long as the `*.d.ts` / `*.d.mts` use the corresponding extensions. (and also, that it would only be used by TS; vite does not need these for anything). (a related relevant issue might be sourcemaps, and their effect in exports).

As such, there is no real need to default the `exports` here -- we can live the work of defaulting it for the consumer. We still need to filter exports to non-dist paths and to transform the paths to be relative to the dist folder.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: When the source `app-shell.config.json` file is not present, the App Shell vite plugin already default's it to an empty `{}` specification. The vite plugin already has a tolerant / lenient behavior / spirit.

With these changes, the published `app-shell.config.json` file is always generated and present. By the same lenient spirit, the corresponding `exports` entry is generated.

---

You seem to prefer a non-lenient / strict spirit for the vite plugin. Where the source `app-shell.config.json` file would need to always exist. Where the exports entry would need to be present in the source `package.json`. Where the UI-Kit CLI would create app shell skeletons where this is readily met. Where the vite plugin would error out if any of the above conditions is not met.

---

Regarding the exports entry for the `package.json` file itself, that is now also always transformed and placed on the publication folder. It needs to exist at the source location, due to NPM. It is a case where people would more often forget.

---

I honestly don't see a big issue with this kind of lenient behavior. On other things, I tend to prefer hard errors and bailing out. Let's discuss.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: I think it's not all roses to use exclusion based, and actually prefer the original, inclusion based.

The main reason for using an inclusion list is that it allows us to _focus_ on the properties that are known to be relevant for the "App Shell runtime". Some of those need transformation, others not. The ones that are supported and relevant for the "App Shell runtime" are those which are copied/transformed over and that is clearly stated in the code. We can include more properties whenever support for more features is added.

Using exclusion, we need to exclude a whole lot of `package.json` properties (in fact, the majority). They either contain "private" information, or are dev-only, or apply to types of packages which an app bundle clearly is not (`bin` et al.), or are properties for features the App Shell runtime does not support.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: The spec says that each **key** of an object either starts with a "." or not.
The keys that start with a "." are taken as relative paths.
The keys that do not are taken as conditions.

The variable name `subpath` in misleading, but it is the object key.

So, this was ignoring keys for conditions: ` if (!subpath.startsWith(".")) continue;`.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: AFAICT, this is still part of the node spec; not legacy.

Arrays can be used to specify multiple paths, where node uses the first that exists.

This doesn't make sense in a web use case, especially because we're translating `exports` to equivalent import maps.

Still, in trying to be compliant whenever possible, and not cause hard to understand issues to consumers, we support the array syntax very simply, by picking the first element.

https://nodejs.org/api/packages.html#exports

> `exports` Type: [\<Object>](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object) | [\<string>](https://developer.mozilla.org/docs/Web/JavaScript/Data_structures#string_type) | [\<string[]>](https://developer.mozilla.org/docs/Web/JavaScript/Data_structures#string_type)

The documentation is not great in this regard...
The full picture can be built from the resolution algorithm documentation: 
- https://nodejs.org/api/esm.html#resolution-algorithm-specification
- function `PACKAGE_EXPORTS_RESOLVE`
- function `PACKAGE_TARGET_RESOLVE`

----

This brings the issue that currently the `transformExportsForDist` function is not currently implemeting full recursive support for this (the function top-level should test for string, array or object and then recurse on keys. arrays and strings are also supported top-level.

Again, like already said elsewhere, the goal is to support the spec whenever possible, if not much effort, so as to cause the least surprise to developers, and try to be the least opinionated possible on the "right form" to use `exports`.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
zettca: > But you also waive node compliance.

So what? We're not Node. We don't need to support every feature it supports, especially those we don't want to promote or are there just for backwards-compatibility.

Related:
- App Shell has been [ESM-only](https://antfu.me/posts/move-on-to-esm-only) since day 1.
- UI Kit _goes out of its way_ to [remove `main` from the published `package.json`](https://github.com/pentaho/hv-uikit-react/blob/master/packages/core/package.json#L89-L91).

> How to organize your files “correctly” is a hot, controversial topic

It's debatable mostly because builders correct for the problems (typically with a noticeable `vite build`/`vitest`/etc. impacts). App Shell modules are runtime, so this also affects end-users.

```
/ModuleA.js
/ModuleB.js
/HugeModuleC.js
/index.js
```

We should be doing this
```tsx
import {ModuleB} from "@hv-apps/other-app/ModuleB" // ✅
import {ModuleB} from "@hv-apps/other-app" // ❌
```

otherwise the end-users will be fetching `HugeModuleC` without need (network fetch, bundle size, potential side effects/bugs)


Aren't we the ones handling this new `dist/package.json`?
That being the case, I don't think it's worth it to spend time handling multiple options (what if user has both `main` & `exports`? what if they don't match?)

But I might be missing something, I don't have all of the context on how this'll be used.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -54,68 +54,132 @@ function stripPrefix(prefix: string, value: string): string {
...
+ * - Arrays: resolved to first non-null element (browser import maps have no
+ *   fallback concept — only one URL per specifier). Returns null if all
+ *   elements resolve to null.
+ * - null: returned as-is — callers must filter null results from the output
+ *   since browser import maps have no exclusion concept
  */
-function transformConditions(
-  obj: Record<string, unknown>,
+function transformTarget(
+  target: unknown,
```
dcleao: Couldn't the TS typing be improved?

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -54,68 +54,132 @@ function stripPrefix(prefix: string, value: string): string {
...
+ *   exclusion concept)
+ * - Filters out string exports that don't resolve within the output directory
  */
 function transformExportsForDist(
-  exports: Record<string, unknown>,
+  exports: unknown,
   distPrefix: string,
   sourceCondition: string,
 ): Record<string, unknown> {
+  const subpathMap = normalizeExports(exports);
```
dcleao: What's the goal of this normalization?

Are we assuming that other parts of the system will not support handing string | array | object?

Because, otherwise, we could just return whatever form we're given at input (modulo other simplifications/filtering done).

`transformTarget` already handles the 3/4 (4, if you include `nulll`) types of possible values. A top-level `null` `exports` is not allowed though, and is already being handled separately before calling this function (but should really be moved here, perhaps).

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -163,21 +232,15 @@ function generateDistPackageJson(
...
-    distPkg.exports = transformExportsForDist(
-      pkg.exports as Record<string, unknown>,
-      distPrefix,
-      sourceCondition,
-    );
-  } else {
-    distPkg.exports = { "./*": "./*" };
-  }
+  // Transform exports or initialize with wildcard fallback for import map resolution
+  const exports: Record<string, unknown> =
```
dcleao: Would better handle all cases, including `null`, within `transformExportsForDist`. Within, it already has to handle `null` for recursive target values anyway.

dcleao (changes requested):

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Since generateDistPAckageJson always appends the `package.json` and `app-shell.config.json` as subpath entries after the transformation, it requires the result of said transformation to be a "normalized" object. Looking at `PACKAGE_EXPORTS_RESOLVE` algorithm, I've made `normalizeExports` a precondition to step 2 and step 3, and if not here, we would have to apply the normalization on the result in order to append "our" entries.
All-in-all, seems like a case of placement.

Regarding top-level `null` , agreed 👍

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
loonatic: Addressed.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: Thank you for your clarifications.

---

As I had said, I am fine with not supporting `main`, nor the other even less standard fields (such as `module`).

---
Btw, not that it changes anything:

- `main` can be used for ESM as well, even if `exports` is recommended.
- Regarding the `index` issue, I maintain my position. Not only I believe that there are use cases for which it is legitimate, and can be used without the issues you mention, but also that we should remain agnostic of such issues as much as possible.

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
dcleao: Oh, I see. Thanks

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
zettca: Nevermind, I missed that `string[]` completely, and had never seen it in use before, nor do the examples use it.

That being said, that doesn't mean we need to support the full spec - but it's a welcome addition

zettca (approved):

packages/app-shell-vite-plugin/src/vite-dist-package-json-plugin.ts
```diff
@@ -278,16 +276,17 @@ function generateDistPackageJson(
...
- * - Always adds `./package.json` and `./app-shell.config.json` entries
- *
+ * directory with only the fields needed for deployment:
+ * - Identity fields (`name`, `version`, `type`, `license`, `description`)
+ * - Transformed `exports` field:
+ *   - Normalizes all valid ESM export forms to a subpath map
+ *   - Strips the dev-only `sourceCondition` key from condition objects
+ *   - Removes the Vite `build.outDir` prefix from all paths
+ *   - Flattens array fallbacks to a single resolved path
+ *   - Filters out null targets (excluded subpaths)
```
dcleao: Just realized now. `null` targets (excluded sub-paths) are actually a feature which we should support. It maps nicely to import maps.

dcleao (changes requested):

dcleao (approved):
🚀

PR #5178, opened 2026-05-26
feat(Badge): add new color variants

plagoa:
image

### DIFF
--- a/packages/core/src/themes/pentaho.ts
+++ b/packages/core/src/themes/pentaho.ts
@@ -214,22 +214,17 @@ export const pentaho = mergeTheme(pentahoBase, {
             color: theme.colors.textSubtle,
             backgroundColor: theme.colors.bgPageSecondary,
           },
-          "&[data-color='primary']:not(:empty)": {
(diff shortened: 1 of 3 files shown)


packages/core/src/themes/pentaho.ts
```diff
@@ -214,10 +214,17 @@ export const pentaho = mergeTheme(pentahoBase, {
             color: theme.colors.textSubtle,
             backgroundColor: theme.colors.bgPageSecondary,
           },
-          "&[data-color='primary']:not(:empty)": {
-            color: theme.colors.primary,
-            backgroundColor: theme.colors.primaryDimmed,
-          },
+          ...Object.fromEntries(
+            (["primary", "positive", "warning", "negative"] as const).map(
```
zettca, line 218: can we leverage `semaColors`, aligning this with `HvAvatar`?
we can keep `info` undocumented if you think it's better

```suggestion
            (["primary", ...semaColors]).map(
```

zettca (approved):

francisco-guilherme (approved):

PR #5202, opened 2026-06-29
feat(Header): update to PD 0.12 specs

plagoa:
This will have to be followed up with changes to App Shell and the `default-app`

### DIFF
--- a/packages/styles/src/themes/pentaho.ts
+++ b/packages/styles/src/themes/pentaho.ts
@@ -185,7 +185,7 @@ const pentaho = makeTheme((theme) => ({
     large: "16px",
   },
   header: {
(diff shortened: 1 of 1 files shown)


packages/styles/src/themes/pentaho.ts
```diff
@@ -185,7 +185,7 @@ const pentaho = makeTheme((theme) => ({
     large: "16px",
   },
   header: {
-    height: "64px",
+    height: "48px",
```
zettca, line 188: those (hopefully none) who are hard-coding `64px` instead of `theme.header.height` will be affected

zettca (approved):

PR #5209, opened 2026-07-07
feat(VerticalNavigation): update styles to PD 0.12

plagoa:


packages/core/src/VerticalNavigation/index.ts
```diff
@@ -4,4 +4,5 @@ export * from "./Actions";
 export * from "./Navigation";
 export * from "./NavigationSlider";
 export * from "./TreeView";
+export * from "./NavigationPopup";
```
zettca: this is internal
```suggestion
```

packages/core/src/VerticalNavigation/VerticalNavigation.tsx
```diff
@@ -146,6 +149,8 @@ export const HvVerticalNavigation = forwardRef<
       <div
         id={id}
         ref={ref}
+        data-theme={isPentaho ? activeTheme?.name : undefined}
+        data-color-mode={isPentaho ? "dark" : undefined}
```
zettca: we can put this directly on the `pentaho.ts` theme

```tsx
    HvVerticalNavigation: {
      ["data-theme" as string]: "pentahoPlus",
      ["data-color-mode" as string]: "dark",
```

packages/core/src/VerticalNavigation/VerticalNavigation.styles.tsx
```diff
@@ -10,6 +10,8 @@ export const { staticClasses, useClasses } = createClasses(
       justifyContent: "flex-start",
 
       width: "220px",
+      overflow: "hidden",
+      transition: "width 300ms ease",
```
zettca, line 14: the transition is a bit janky, with the alignment of the icons

also, we should disable this when the user doesn't want animations with [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion)

packages/app-shell-ui/src/components/layout/VerticalNavigation/VerticalNavigation.tsx
```diff
@@ -84,9 +87,20 @@ export const VerticalNavigation = () => {
...
-  });
+  useEffect(() => {
+    if (isCompactMode) {
+      setVerticalNavigationWidth(0);
+      return;
+    }
+
+    setVerticalNavigationWidth(
+      open
+        ? isPentahoTheme
```
zettca: aren't we duplicating the sizing logic?

also, this `useEffect` is problematic...

plagoa:
@zettca updated after feedback.

francisco-guilherme (approved):

PR #5210, opened 2026-07-07
chore: remove compat theme extensions

zettca:


zettca:
no side-effects 👌, but holding this for a later time

PR #5230, opened 2026-08-11
feat: remove NEXT theme

plagoa:


packages/core/src/providers/Provider.tsx
```diff
@@ -64,7 +64,7 @@ export interface HvProviderProps {
   /**
    * The active theme. It must be one of the themes passed to `themes`.
    *
-   * If no value is provided, the first theme from the `themes` list is used. If no `themes` list is provided, the `next` theme will be used.
+   * If no value is provided, the first theme from the `themes` list is used. If no `themes` list is provided, the `pentaho` theme will be used.
```
zettca: this is outdated, there is no longer a `themes` array

```suggestion
   * @default pentaho
```

apps/docs/src/content/docs/theming.mdx
```diff
@@ -21,7 +20,7 @@ Use the `HvProvider` to configure the active theme and color mode:
 </HvProvider>
 ```
 
-- If `themes` is not provided, the default is `next`.
+- If `themes` is not provided, the default is `pentaho`.
```
zettca: this no longer exists. line below too
```suggestion
```

apps/docs/src/components/ThemeSwitcher.tsx
```diff
@@ -5,10 +5,7 @@ import { HvSelect } from "@hitachivantara/uikit-react-core";
 
 import { useDocsTheme } from "../hooks/useDocsTheme";
 
-const themes = [
-  { value: "pentaho", label: "Pentaho" },
-  { value: "next", label: "NEXT" },
-];
+const themes = [{ value: "pentaho", label: "Pentaho" }];
```
zettca: no point in having a theme switcher if we have just one theme - we should remove this

Image

zettca (commented):
looking good, just some small leftovers (from last major 😬)

apps/docs/src/components/ThemeSwitcher.tsx
plagoa: shouldn't we keep it as a way to later add a new theme if we need it? Maybe we don't remove it right now but also don't show it on the docs?

zettca: if we're not using the component at all, then it's dead code
we can easily recover it later, via git

apps/docs/src/content/docs/theming.mdx
```diff
@@ -86,7 +84,7 @@ const MyComponent = () => {
         changeMode(selectedMode === "light" ? "dark" : "light");
       }}
     >
-      <ThemeSwitcher />
+      Toggle mode
```
zettca: this was an icon, which `<HvButton icon>` expects

zettca (approved):

PR #5236, opened 2026-08-13
feat: promote Pentaho overrides to the components 

plagoa:


packages/core/src/Badge/Badge.styles.tsx
```diff
@@ -22,9 +22,10 @@ export const { staticClasses, useClasses } = createClasses("HvBadge", {
       left: "100%",
     },
     ...theme.typography.caption2,
-    color: theme.colors.textDimmed,
+    "--bg-color": theme.colors.textSubtle,
```
zettca: this is already in the fallback below
```suggestion
```

packages/core/src/BaseDropdown/BaseDropdown.styles.tsx
```diff
@@ -17,9 +17,9 @@ export const { useClasses, staticClasses } = createClasses("HvBaseDropdown", {
     cursor: "pointer",
     userSelect: "none",
     position: "relative",
-    backgroundColor: theme.colors.bgContainer,
+    backgroundColor: `light-dark(${theme.colors.dimmer}, ${slate[950]})`,
```
zettca: They use `Global Colors/Pure/pure-white`, which for us is just `"white"`

```suggestion
    backgroundColor: `light-dark(white, ${slate[950]})`,
```

packages/core/src/BaseInput/BaseInput.styles.tsx
```diff
@@ -35,9 +35,9 @@ export const { staticClasses, useClasses } = createClasses("HvBaseInput", {
     borderRadius: theme.radii.round,
     height: "32px",
     borderWidth: 1,
-    borderColor: theme.colors.text,
+    borderColor: theme.colors.textDimmed,
     boxSizing: "border-box",
-    backgroundColor: theme.colors.bgContainer,
+    backgroundColor: `light-dark(${theme.colors.dimmer}, ${slate[950]})`,
```
zettca:
```suggestion
    backgroundColor: `light-dark(white, ${slate[950]})`,
```

packages/core/src/BulkActions/BulkActions.tsx
```diff
@@ -69,7 +69,7 @@ export const HvBulkActions = forwardRef<
     numSelected = 0,
     selectAllConjunctionLabel = "/",
     showSelectAllPages,
-    semantic = true,
+    semantic = false,
```
zettca: this no longer exists, we should remove the prop

packages/core/src/Button/Button.tsx
```diff
@@ -123,7 +123,7 @@ export const HvButton = fixedForwardRef(function HvButton<
     startIcon,
     endIcon,
     size,
-    radius,
+    radius = "full",
```
zettca: we should keep these props clean/empty when possible. we can put this on the `borderRadius` in `Button.styles.ts`, instead of overriding here the `.base` value

same for other components

packages/core/src/Card/Card.styles.tsx
```diff
@@ -7,26 +7,47 @@ export const { staticClasses, useClasses } = createClasses("HvCard", {
...
+    boxShadow: theme.colors.shadow,
     "&:focus-visible": {
       ...outlineStyles,
     },
     "&:focus": {
       outline: "none",
     },
+
+    // default non-semantic card
+    "&[data-color=sema0]": {
```
zettca, line 24: we should  probably review this `sema0` (later)

packages/core/src/Header/Header.styles.tsx
```diff
@@ -12,7 +12,8 @@ export const { staticClasses, useClasses } = createClasses("HvHeader", {
     flexShrink: 0,
     zIndex: theme.zIndices.banner,
     borderTop: "none",
-    boxShadow: theme.colors.shadow,
+    borderBottom: `1px solid ${theme.colors.borderSubtle}`,
+    boxShadow: "none",
```
zettca:
```suggestion
```

packages/core/src/Select/Select.styles.ts
```diff
@@ -1,9 +1,14 @@
...
-import { theme } from "@pentaho/uikit-styles";
+import { slate, theme } from "@pentaho/uikit-styles";
 
 export const { staticClasses, useClasses } = createClasses("HvSelect", {
   root: {
     position: "relative",
+    borderRadius: theme.radii.round,
+    "& .HvButton-subtle[data-color=text]": {
+      borderColor: theme.colors.textDimmed,
+      backgroundColor: `light-dark(${theme.colors.dimmer}, ${slate[950]})`,
```
zettca:
```suggestion
      backgroundColor: `light-dark(white, ${slate[950]})`,
```

packages/core/src/Avatar/Avatar.tsx
```diff
@@ -78,8 +78,8 @@ export const HvAvatar = forwardRef<
     children: childrenProp,
     component = "div",
     size: sizeProp,
-    backgroundColor,
-    color: colorProp,
+    backgroundColor = "primaryDimmed",
+    color: colorProp = "primaryStrong",
```
zettca: we should avoid default values for overrides - let's set them below

zettca (approved):
Looking good 👍 

We just need to move some prop overrides to the base style values

PR #5241, opened 2026-08-18
refactor(core,widgets): migrate @mui/base to @base-ui/react

plagoa:


zettca (approved):
Looking good 👍 

There's just an outline that we shouldn't show on the popup container

Image

PR #5254, opened 2026-08-21
docs: update migration guide

plagoa:


apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
 
-WIP
+### 1) Package scope renamed
+
+All public packages moved from `@hitachivantara/*` to `@pentaho/*`.
+
+Update every import and dependency entry.
+
+```diff
+diff --git a/src/App.tsx b/src/App.tsx
```
zettca:
```suggestion
```
these diff lines don't help the user

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
+
+```diff
+diff --git a/src/dashboard.tsx b/src/dashboard.tsx
+-import { HvDashboard } from "@hitachivantara/uikit-react-pentaho";
++import { HvDashboard } from "@pentaho/uikit-react-widgets";
+```
+
+### 2.1) Standalone icons package is no longer publicly available
+
+The standalone icons package is still present in the repository, but in v7 it is no longer publicly available as a consumer-installable package.
```
zettca: The user (or the AI) doesn't need to know `icons` still exists. Let's keep this simple, migration/breaking-change oriented

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
+The NEXT theme is no longer exported.
+
+- Removed export: `next`
+- Removed bundle entry: `themes.next`
+
+Also, the Pentaho theme identifier was renamed:
+
+- `pentahoPlus` -> `pentaho`
+
+If your app checks `theme.name`, update any strict comparisons.
```
zettca: apps don't do this

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
+For most consumers this is transparent, but it can affect:
+
+- custom style overrides that target internal DOM/class structure
+- test selectors coupled to internal markup
+
+Revalidate any deep customization around `Select`, tabs/canvas panels, and dropdown-like controls.
+
+### 6) Grid implementation update
+
+`HvGrid` migrated from legacy MUI Grid APIs to current MUI Grid APIs.
```
zettca: we should have a migration diff example for this

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
+
+- `Carousel`
+- `Controls`
+- `Login`
+- `ScrollToHorizontal`
+- `ScrollToVertical`
+- `SimpleGrid`
+- `Stack`
+
+Remove those imports and replace with alternatives in your application code.
```
zettca: which alternatives? we should have a minimal alternative code snippet or link for users to have an alternative

I suggest:

- Carousel: suggest a minimal [CSS carousel](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Overflow/Carousels#setting_up_scroll_snapping_on_the_list) (`overflow-x-scroll snap-x snap-mandatory`), or using [embla](https://www.embla-carousel.com/)
- Controls: is a widget/"template" - copy source code from `v6.x`
- Login: was a minimal template, point to our [Login examples](https://pentaho.github.io/uikit-docs/master/examples/login)
- ScrollTo: `HvListContainer`+`HvListItem` with `component="a"` and `href="#hash` and [CSS smooth scroll](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scroll-behavior)
- SimpleGrid: `grid grid-cols-2 md:grid-cols-4`, etc.
- Stack: `flex gap-sm`, etc.

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
+
+Revalidate any deep customization around `Select`, tabs/canvas panels, and dropdown-like controls.
+
+### 6) Grid implementation update
+
+`HvGrid` migrated from legacy MUI Grid APIs to current MUI Grid APIs.
+
+If you rely on grid internals or legacy behavior, verify layout parity (especially spacing and breakpoint behavior).
+
+## Suggested migration sequence
```
zettca: isn't this kind of an index? shouldn't it be higher up?

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
...
+
+1. Update all package dependencies from `@hitachivantara/*` to `@pentaho/*`.
+2. Rename `uikit-react-pentaho` usage to `uikit-react-widgets`.
+3. Remove any direct usage of `uikit-react-lab`.
+4. Remove `next` theme usage and replace `pentahoPlus` checks.
+5. Replace removed core components (`Carousel`, `Stack`, etc.).
```
zettca: if we're using the ticks ` to represent actual code, we should use the actual component names!

```suggestion
5. Replace removed core components (`HvCarousel`, `HvStack`, etc.).
```

apps/docs/src/content/docs/migration.md
```diff
@@ -1,10 +1,153 @@
 # Migration to v7
 
-UI Kit v7 is a **major release** focused on tighter integration with the Pentaho UI, and AI tools.
+UI Kit v7 is a major release with breaking changes in package names, themes, and component availability.
```
zettca: I think we should leave this as a "thematic" introduction. "major release with breaking changes" doesn't add much

We should keep all the non-breaking change "context"/rationale for the release here

below `## Breaking changes`, we should focus on the actual breaking changes & diff's. show HOW, the WHY is irrelevant

